PR #31, opened 2024-05-19
Fix typos in docstrings

pitmonticone:


jstoobysmith (approved):
Many thanks for these. This is very helpful.

PR #57, opened 2024-06-14
feat(Feynman diagrams): symmetry factor and connectedness

jstoobysmith:
Add the symmetry factors and connectedness properties of Feynman diagrams in Phi^4 theory.


jstoobysmith:
(Still to do - speed up the calculation of symmetry factors, as factorially slow currently)

PR #134, opened 2024-08-30
refactor: golf a few proofs

pitmonticone:


jstoobysmith:
Many thanks for this!

PR #203, opened 2024-10-21
feat: TensorSpecies

jstoobysmith:


jstoobysmith:
The aim of this PR is to introduce the general notion of a `TensorSpecies`. The underlying structure will replace the structure of `TensorStruct`, and we will add new propositional conditions.

PR #211, opened 2024-10-25
feat: Bispinors

jstoobysmith:


jstoobysmith:
More lemma's about Bispinors will come in a later pull request. I stop for now as need properties related to the interaction of products and contractions.

PR #225, opened 2024-11-05
feat(gamma-relations): Update matrix definition of gamma-matrices

fenghourun:


jstoobysmith (approved):

jstoobysmith:
Many thanks! You should be able to merge this when you want - let me know if you can't as it means I've done the rules wrong for pushing to main.

Issue #295, opened 2025-01-24
Improve meta-programming and display around informal results


jstoobysmith:
There are two parts to this issue:

- [x]  The way `informal_lemma` and `informal_definition` currently work in HepLean is that they are turned through elaboration into elements of  the type `InformalLemma` or `InformalDefinition`. This does not seem like the most optimal approach, and an approach similar to `TODO` elements will likely work better. 
- [ ]  The output of `informal_lemma` and `informal_definition` on the [HepLean website](https://heplean.com/InformalGraph.html) needs improving. Either with an updated form of graph or just a list of them.

jstoobysmith:
Let me explain in a bit more detail what I mean regarding the second point: 

Both Lean blueprint  and HepLean output a dependency graph. In the case of Lean blueprints this dependency graph is generated via Latex. In the case of HepLean it is generated via `informal_lemma` and `informal_definition`. This I want to remain the same. 

However, the outputted display of the dependency graph for HepLean (see [here](https://heplean.com/InformalGraph.html)) is IMO worse then that for Lean blueprint (see [here](https://imperialcollegelondon.github.io/FLT/blueprint/dep_graph_chapter_7.html)). 
In particular, the two main features of the Lean blueprint display that I would like for the HepLean display are: 

1. Better fitting to the page (the HepLean graph goes off to the right on my screen). 
2. Pop-up information when a node is clicked. (This already exists for the HepLean graph but it is at the bottom and I think goes unnoticed). 

I think this will ultimately be a task in HTML or manipulating dot file formats.

jstoobysmith:
@pitmonticone I added a bit more detail here c.f. the dependency graph. Not sure if this is something you would know how to do easily? No worries if not.

PR #300, opened 2025-01-27
feat: make informal_definition and informal_lemma commands

kuotsanhsu:
After the change, the dependency graph of informal definitions/lemmas in SVG format looks the same as before. @jstoobysmith mentioned in #295 that the display can be improved; what do you mean exactly? Do you wish to use something other than graphviz?


jstoobysmith:
@kuotsanhsu This looks amazing! Thank you. 

Concerning the display: The display of a similar graph for the FLT project is:
https://imperialcollegelondon.github.io/FLT/blueprint/dep_graph_chapter_8.html
To me this looks a lot better than what we have for HepLean, and if we could replicate this it would be great. 
Maybe @pitmonticone  has an idea of how this can be done? 

On a side note - I really need to get better with putting more details in 'TODO's and 'Issues' etc. So sorry things are unclear.

kuotsanhsu:
The docBlame linter called by runLinter checks for missing docstrings, and linters called by runLinter can be disabled using the "nolint" attribute per declaration. Thus, I removed `scripts/MetaPrograms/no_docs.lean`.

I also added the attribute `nolint docBlame` to `informal_{definition,lemma}` by default in their command definitions so that are not required to have docstrings.

There is also no separate command definitions for `informal_{definition,lemma}_note`. They can be created like this: `@[note_attr_informal] informal_definition ...`.

jstoobysmith:
One of the reasons I made `scripts/MetaPrograms/no_docs.lean` was to check for missing doc-strings on lemmas (I had the idea at one point that I would go through all lemmas and give them doc-strings - but this would be a huge task now). 
So maybe it's worth keeping the script there but removing it from `lint_all.lean`. 

I think it would be nice to automatically give `informal_{definition, lemma}` their `math` argument as a doc-string. I tried this with the old implementation of `informal_{definition, lemma}` before but could not get it to work. Do you think it would be possible with this implementation?

kuotsanhsu:
> Concerning the display: The display of a similar graph for the FLT project is: https://imperialcollegelondon.github.io/FLT/blueprint/dep_graph_chapter_8.html To me this looks a lot better than what we have for HepLean, and if we could replicate this it would be great. Maybe @pitmonticone has an idea of how this can be done?

I have never used Blueprint before. At first glance, Blueprint seems to require TeX files. If we were to head in this direction, we should definitely consult @pitmonticone on this. If you just want to have (1) the URL link to the code and (2) the docstring display on hover, I can hack something in the dot file generator in a separate PR.

As a side note, it's just awesome to witness efforts like Mathlib, HepLean, the FLT project, and the like; glad I'm alive.

jstoobysmith:
> I have never used Blueprint before. At first glance, Blueprint seems to require TeX files. If we were to head in this direction, we should definitely consult @pitmonticone on this. If you just want to have (1) the URL link to the code and (2) the docstring display on hover, I can hack something in the dot file generator in a separate PR. 

Yep these two things are what I would want, and a slightly better embedding of the dot-file into the html webpage - I think we should avoid TeX.

jstoobysmith (approved):

jstoobysmith:
(Feel free to merge this PR when you are ready! Many thanks :))

kuotsanhsu:
> I think it would be nice to automatically give `informal_{definition, lemma}` their `math` argument as a doc-string. I tried this with the old implementation of `informal_{definition, lemma}` before but could not get it to work. Do you think it would be possible with this implementation?

What about dropping the `math` field from the structures, and just write it as docstring. I can modify the dotfile generator to display the docstrings.

jstoobysmith:
@kuotsanhsu Yes that would work I think.

pitmonticone:
Hi everyone, I’ll be quite busy over the next two weeks. 

Could you please provide a clear explanation of what you’d like to have? 

I’ll make sure to take a look at it as soon as I find the time.

jstoobysmith:
@pitmonticone I will update this #295 latter today with an update of what I envision, and will let you know :).

PR #345, opened 2025-02-19
feat: 1d quantum Harmonic oscillator 

jstoobysmith:
This PR introduces the 1d quantum Harmonic oscillator.


jstoobysmith:
Mathlib has Hermite polynomials:

https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Polynomial/Hermite/Gaussian.html

So need to replace those appearing here

PR #347, opened 2025-02-25
feat: Completeness of eigenvectors for harmonic oscillator

jstoobysmith:
Reduce the completeness of eigenvectors for harmonic oscillator to a single theorem in analysis: Plancherel's theorem, which is not yet in Mathlib.


jstoobysmith:
Lots of refactoring to be done here.

PR #405, opened 2025-03-19
feat: Twin paradox

jstoobysmith:


PhysLean/Relativity/Special/TwinParadox/Basic.lean
```diff
@@ -0,0 +1,121 @@
...
+namespace SpecialRelativity
+
+open Manifold
+open Matrix
+open Real
+open ComplexConjugate
+open Lorentz
+open Vector
+
+/-- The twin paradox assuming instantaneous accelration -/
```
or4nge19: typo : accelration

jstoobysmith: Fixed. Thanks

PR #426, opened 2025-03-24
General 1D Potential

Cobord:


PhysLean/QuantumMechanics/OneDimension/GeneralPotential/Basic.lean
```diff
@@ -0,0 +1,209 @@
+/-
+Copyright (c) 2025 Ammar Husain. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ammar Husain
+-/
```
jstoobysmith, line 5: Please can you confirm you are ok with this copyright statement.

PhysLean/QuantumMechanics/OneDimension/GeneralPotential/Basic.lean
```diff
@@ -0,0 +1,209 @@
...
+  rw [ht1,ht2]
+  exact hψ1_x
+  exact hψ2_x
+  exact hψ1_xx
+  exact hψ2_xx
+
+/-- The proposition on `Q` corresponding to the condition that
+  `Q.V` is bounded from below. -/
+def Bounded : Prop :=
+  (∃ E, ∃ R, ∀ z < -R, E < Q.V z) ∧ ∃ E, ∃ R, ∀ z > R, E < Q.V z
```
jstoobysmith: I modified your bounded definition. Please can you check this is right. 
You had \forall E rather then \exists E. I think \exists is correct here.

PhysLean/QuantumMechanics/OneDimension/GeneralPotential/Basic.lean
Cobord: yes

PhysLean/QuantumMechanics/OneDimension/GeneralPotential/Basic.lean
Cobord: I wanted it to keep growing  (for the propogating regioin to be bounded) to be able to avoid non-normalizable concerns when using L2 theorems. But without the precise theorem statement for which some condition on the potential growth is an assumption, I'd rather just remove it until have something more specific to prove.

jstoobysmith: Ok, thanks for your explanation. This makes sense. I've removed it for now. Once this PR has passed the CI workflow (which it should do automatically now), I will merge it. 

Many thanks for this :) - and sorry for the delay, I really thought I had submitted this review earlier.

jstoobysmith (approved):

jstoobysmith:
I've merged this PR. Many thanks for this Ammar. Would love to see this all taken further :).

PR #427, opened 2025-03-25
Update README links to new documentation location

oeb25:
This PR updates the https://heplean.github.io/HepLean/docs/HepLean/... links to https://physlean.com/docs/PhysLean/... where some of the files were moved around.

To prevent this in the future (I mean the things moving around) one might consider using [this feature of doc-gen4](https://github.com/leanprover/doc-gen4?tab=readme-ov-file#how-does-docsnatadd-from-the-lean-zulip-work) if we know that definition or theorem names are more stable.

I also spotted some links to https://heplean.github.io/HepLean/ in [`/scripts/stats.lean`](https://github.com/HEPLean/PhysLean/blob/16554ebe68c9d58a963fcbf80c7bc5a05cad7268/scripts/stats.lean#L81-L99) and [`/scripts/MetaPrograms/informal.lean`](https://github.com/HEPLean/PhysLean/blob/08e9226e22155d89e6a7c503ab02994d4ef51f23/scripts/MetaPrograms/informal.lean#L115), but did not update them now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,23 +27,23 @@ into Lean 4.
 ## Some parts of PhysLean
 PhysLean _currently_ includes, but is not limited to, the following parts:
 
-__Lorentz [🗂️](https://heplean.github.io/HepLean/docs/HepLean/Lorentz/Group/Basic.html):__  The Lorentz group, Lorentz algebra, Weyl fermions, Real Lorentz vectors, complex Lorentz vectors, complex Lorentz tensors, bispinors, Pauli matrices, etc.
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):

jstoobysmith:
Many thanks for this! Yeah I suspect theorems and def names are more stable, although hopefully these particular files are fairly stably located now.

Once the CI checks (which this PR will pass because it doesn't change the Lean code) have finished I will merge.

PR #439, opened 2025-04-01
update Causality + typos

or4nge19:
part 1 of lemmas needed for causal order and invariance


jstoobysmith:
Hey, thanks for this. The linter is saying that some of the lines are too long. In particular:

```
PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean:131 : ERR_LIN : Line has more than 100 characters
PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean:163 : ERR_LIN : Line has more than 100 characters
PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean:255 : ERR_LIN : Line has more than 100 characters
PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean:263 : ERR_LIN : Line has more than 100 characters
```

Would you mind fixing this?

PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean
```diff
@@ -151,6 +158,172 @@ def causallyUnrelated {d : ℕ} (p q : Vector d) : Prop :=
...
+  simp at hv
+  rw [h] at hv
+  simp at hv
+  have h_spatial_nonneg : 0 ≤ ∑ i, v (Sum.inr i) * v (Sum.inr i) :=
+    Finset.sum_nonneg (fun i _ => mul_self_nonneg (v (Sum.inr i)))
+  exact lt_irrefl 0 (h_spatial_nonneg.trans_lt hv)
+
+/-- Time component squared is positive for timelike vectors -/
+@[simp]
+lemma time_squared_pos_of_timelike {d : ℕ} {v : Vector d}
```
jstoobysmith: This lemma does not match its name or description.

PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean
```diff
@@ -151,6 +158,172 @@ def causallyUnrelated {d : ℕ} (p q : Vector d) : Prop :=
...
+    (hv : causalCharacter v = .timeLike) :
+    v (Sum.inl 0) ≠ 0 :=
+  time_component_ne_zero_of_timelike hv
+
+/-- If a vector has zero norm, then all its components are zero -/
+lemma vector_zero_of_sum_sq_zero {d : ℕ} {v : Fin d → ℝ}
```
jstoobysmith: I think it might be easier here if you replace `Fin d → ℝ` with `EuclideanSpace ℝ (Fin d)`. We will get for free an inner product which corresponds to your ` ∑ i, v i * v i`. I suspect Cauchy etc. have already been proved for this inner product.

jstoobysmith: Happy for you to leave this one as is for now, but maybe we should add a TODO item regarding this.

PhysLean/Relativity/Lorentz/RealTensor/Vector/Causality.lean
```diff
@@ -151,6 +158,172 @@ def causallyUnrelated {d : ℕ} (p q : Vector d) : Prop :=
...
+
+/-- Causally preceding is reflexive -/
+@[simp]
+lemma causallyPrecedes_refl {d : ℕ} (p : Vector d) : causallyPrecedes p p := by
+  right; simp [pastLightConeBoundary]
+
+/-- For timelike vectors in Minkowski space, the squared time component exceeds
+    the sum of squared spatial components -/
+@[simp]
+lemma timelike_time_dominates_space {d : ℕ} {v : Vector d}
```
jstoobysmith: Maybe we should also write TODO items for the spacelike and lightlike versions of this lemma.

or4nge19:
still a few lint errors, fixed now. I'm rebuilding for a last check before pulling again

jstoobysmith:
Thanks again for this.
I have done the following: 

1. Split Properties.lean up into TimeLike and LightLike
2. Removed a lot of duplicated results (lots of things where proved multiple times)
3. For now removed the results related to proportionality for timelike vectors (I currently cannot see why these are needed - but maybe they are needed for something which is to come later?) 

Another quick comment: A lot of the names contained timeLike but did not assume the argument was timelike.

or4nge19:
no problem, thanks!. all these are needed for results that come later but it's ok to remove what does not look relevant at this tage, I will introduce it along with the results where it is needed. 
some are duplications with a different layout as one never knows what is best way to formulate it until uses it. If removing them does not break the rest of the file, then it will not break the results to come so no problem.

jstoobysmith:
Great thanks for understanding! 

I think the duplicate lemma thing comes a bit down to sticking with a conversion. E.g. always using `timeComponent v` instead of `v (Sum.inl 0)`, always using `a < b` and not `b > a` etc. . Also you can play tricks like `abbrev timeComponent` instead of `def timeComponent` which allows Lean to treat  `timeComponent v` exactly as `v (Sum.inl 0)`.

jstoobysmith:
And yes introducing these results when its clearer what they are needed for will be helpful :).

jstoobysmith (approved):

PR #466, opened 2025-04-10
feat : basis and coord definitions for space

zhikaip:
Not sure what sort of API and descriptions are needed around this, please check if the definitions are fine.


jstoobysmith:
Many thanks for this. I think it would be nice if `basis` had the type `Basis (Fin d) (Space d)` (or something along those lines). This should already be defined for Euclidean space like `EuclideanSpace.single` was.

I like your definition of `coord` - and I think we should keep this. Once `basis` is defined as an actual basis, we should be able to show that 

```
coord i p = basis.repr p i
```

I made a quick PR which moved `noncomputable` to the same line as `def`.

jstoobysmith:
Once `deriv` is defined some API regarding the derivative of coordinates etc. would be nice.

jstoobysmith:
[PiLp.basisFun](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Lp/PiLp.html#PiLp.basisFun) might work for the basis

zhikaip:
Do we want [EuclideanSpace.basisFun](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/InnerProductSpace/PiL2.html#EuclideanSpace.basisFun) or [PiLp.basisFun](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Lp/PiLp.html#PiLp.basisFun)?

jstoobysmith:
The former i.e. [EuclideanSpace.basisFun](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/InnerProductSpace/PiL2.html#EuclideanSpace.basisFun). Good spot :)

jstoobysmith:
Made some small changes. Some of the `Space` I changed back to `EuclideanSpace ℝ (Fin 3)` since in general I don't think they correspond to Space positions.

zhikaip:
How would you suggest to fix the macro? I"m not sure what's the difference between scoped notation and macro, do I just have to add an explanation for its usage? The notation in [SpaceTime](https://github.com/HEPLean/PhysLean/blob/master/PhysLean/Relativity/SpaceTime/Basic.lean) is defined a bit differently, not sure which one is better

jstoobysmith:
From what I understand, `notation` is a specific example of a `macro` (easier to use but less general). The `scoped` bit just means that the notation will only work if the corresponding namespace is open. 

For macros and notation I think it's best to keep the documentation of the thing it corresponds to - since this is what you see on hovering your curser above it. 

Does `@[inherit_doc deriv]` work here - if so this seems like a good solution?

jstoobysmith:
I fixed the doc string here. Turns out `@[inherit_doc deriv]` does work for macros :).

jstoobysmith (approved):
I've reviewed this PR, so once the workflows have finished and pasted, feel free to submit.

PR #478, opened 2025-04-12
refactor: remove `erw`s from LinearMaps.lean

jnlindsay:
This PR is my first attempt to make progress on https://github.com/HEPLean/PhysLean/issues/385. A few questions / comments:

1. Surely, the performance hit by `erw` varies depending on context, and we'd like to focus on where it's particularly bad. @jstoobysmith, do you have a preference of which `erw`s we should focus on? Nevertheless, being new to Lean, LinearMaps.lean seemed like a nice way for me to cut my teeth. 
1. My proofs for `BiLinearSymm` and `TriLinearSymm` are identical. Do we need more abstraction?
1. There's a note `TODO "6VZLZ" "Replace the definitions in this file with Mathlib definitions."` at the top of the file. Haven't looked into this yet.
1. It's possible I've completely missed the point about `erw`s haha, especially about simp-normal form.

### DIFF
--- a/PhysLean/Mathematics/LinearMaps.lean
+++ b/PhysLean/Mathematics/LinearMaps.lean
@@ -85,7 +85,7 @@ def mk₂ (f : V × V → ℚ) (map_smul : ∀ a S T, f (a • S, T) = a * f (S,
 
 lemma map_smul₁ (f : BiLinearSymm V) (a : ℚ) (S T : V) : f (a • S) T = a * f S T := by
   have h : f (a • S) = a • (f S) := by
-    exact f.toLinearMap.map_smul a S
+    exact f.map_smul a S
   simp [h]
 
 lemma swap (f : BiLinearSymm V) (S T : V) : f S T = f T S := f.swap' S T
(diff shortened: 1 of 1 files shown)


jstoobysmith:
Hey many thanks for this, this is very helpful.  To answer your questions: 

1. Removing any `erw`'s are helpful. I think the once in the `Tensor` file are particular bad, but these are also the more complicated to remove - so maybe not a good starting point. 
2. It's fine that the proofs are identical :). 
3. _
4. I've seen it said that if there is an `erw` and `rw` doesn't work it is an indication that there is a simp lemma missing. If it's possible to work out what that simp lemma is then adding it should simplify further results also (and make the whole API easier to work with).

If we could work out exactly what the `erw`'s are doing (i.e. what it is unfolding) then I think working out what the simp lemmas we need to add are would be easier. There might be a way of doing this using profiling, but I'm not sure.

jstoobysmith (approved):
I've approved this, so feel free to merge :).

or4nge19:
Hi :) I'm happy to help if/when you work on 6VZLZ" "Replace the definitions in this file with Mathlib definitions." 
It's something I've started thinking about and meantime got more familiar with the relevant API (it was changed recently) so can give some guidance

jnlindsay:
> Hi :) I'm happy to help if/when you work on 6VZLZ" "Replace the definitions in this file with Mathlib definitions." It's something I've started thinking about and meantime got more familiar with the relevant API (it was changed recently) so can give some guidance

That would be great, thanks :)

PR #479, opened 2025-04-12
Feat: space and time derivatives

zhikaip:
1. Slight change to Space.deriv macro for easier use
(slight changes to the definition of curl for better clarity and matching with div but the definition itself should be the exact same)

2. Added Time.deriv


jstoobysmith (approved):
Many thanks! I've approved this so you should be able to merge - sorry for the delay :).

PR #481, opened 2025-04-14
feat: some vector calculus

zhikaip:
Attempts at doing some vector calculus. I'm not sure how to structure this, do we want this in space or in the mathematics folder?

I'm not sure what actually belongs here, The main thing I tried to do here is to  prove div_of_curl_eq_zero and ended up with a lot of helper lemmas. Please feel free to give me suggestions to clean it up or remove anything you don't see fit, it's just practice for me. Many thanks!

If this is generally ok I'll keep going.

### DIFF
--- a/PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
+++ b/PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
@@ -1,0 +1,200 @@
+import PhysLean.ClassicalMechanics.Space.Basic
+import Mathlib.Analysis.InnerProductSpace.Calculus
+
+/-!
+
+# VectorIdentities
+
+In this file we define common vector calculus identities based on `Space`.
+
+-/
+
+/-!
+
+## General auxiliary lemmas
+
+-/
+
+section
+
+variable
+  {𝕜} [RCLike 𝕜]
+  {X} [NormedAddCommGroup X] [NormedSpace 𝕜 X]
+  {Y} [NormedAddCommGroup Y] [NormedSpace 𝕜 Y]
+  {Z} [NormedAddCommGroup Z] [NormedSpace 𝕜 Z]
+  {ι : Type*} [Fintype ι] {Y' : ι → Type*} [∀ i, NormedAddCommGroup (Y' i)]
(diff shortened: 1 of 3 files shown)


jstoobysmith:
This in the Space folder seems good to me :). Having lots of helper lemmas is a good thing - this is what people mean when they talk about an API, you never know when those helper lemmas will come in use. 

We could maybe put the `General auxiliary lemmas` in the Mathematics folder, but there is only a few of them, so this is not vital. 

But this is a very nice PR :). 

I'll add one or two comments two small comments to the code.

PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
```diff
@@ -0,0 +1,209 @@
...
+  apply h
+end
+
+/-!
+
+## Space.deriv lemmas
+
+-/
+
+open Space
```
jstoobysmith: It would be better if this was `namespace Space`.

PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
```diff
@@ -0,0 +1,209 @@
...
+
+/-- Derivatives on space distiribute over addition. -/
+lemma spacederiv_add [NormedAddCommGroup M] [NormedSpace ℝ M]
```
jstoobysmith: Since the function is called `deriv` not `spacederiv`, it would be better if this was called `deriv_add`. 
The namespace will make this `Space.deriv_add`

jstoobysmith: (Same with some of the lemmas below)

zhikaip: actually this is the same issue, it wasn't letting me name `deriv_add/deriv_commute` becuase of the namespace clash, `namespace Space` works!

zhikaip:
renamed the relevant theorems and corrected namespace

jstoobysmith:
Great! To fix the linter you need to import the new file you made into the PhysLean.lean file (in alphabetical order).

Otherwise this all looks good to me :).

jstoobysmith (approved):
I've approved this - so feel free to merge when ready :)

PR #486, opened 2025-04-15
feat: smul and linearity of div and curl (+cleanup)

zhikaip:
show div and curl are `IsLinearMap`.


jstoobysmith (approved):
Approved - thanks :)

PR #488, opened 2025-04-16
feat: Definition of a second.

jstoobysmith:
Defining the definition of a second. 

See the PhysLean.ClassicalMechanics.Time.Second module for an explanation of this definition.


jstoobysmith:
This PR is going to be broken down into a number of smaller PRs.

jstoobysmith:
Closing this as I am not confident that this approach is the best.

PR #497, opened 2025-04-18
refactor: remove `erw`s from PauliMatrices folder

jnlindsay:
Toward https://github.com/HEPLean/PhysLean/issues/385

### DIFF
--- a/PhysLean/Relativity/PauliMatrices/Relations.lean
+++ b/PhysLean/Relativity/PauliMatrices/Relations.lean
@@ -46,7 +46,8 @@ lemma pauliCo_contr_pauliContr :
     simp only [Nat.reduceAdd, Nat.succ_eq_add_one, Fin.isValue, Fin.succAbove_zero,
       Function.comp_apply, cons_val_zero, cons_val_one, head_cons, ofRat_basis_repr_apply]
     rw [← PhysLean.RatComplexNum.toComplexNum.map_mul]
-    erw [PhysLean.RatComplexNum.ofNat_mul_toComplexNum 2]
+    change (2 : ℕ) * _
+    rw [PhysLean.RatComplexNum.ofNat_mul_toComplexNum 2]
   rw [contrT_basis_repr_apply]
   conv_lhs =>
     enter [2, x]
--- a/PhysLean/Relativity/PauliMatrices/Relations.lean
+++ b/PhysLean/Relativity/PauliMatrices/Relations.lean
@@ -6,7 +6,7 @@ Authors: Joseph Tooby-Smith
 import PhysLean.Relativity.PauliMatrices.Basic
 /-!
 
-## Contractiong of indices of Pauli matrix.
+## Contracting of indices of Pauli matrix.
 
 The main result of this file is `pauliMatrix_contract_pauliMatrix` which states that
 `η_{μν} σ^{μ α dot β} σ^{ν α' dot β'} = 2 ε^{αα'} ε^{dot β dot β'}`.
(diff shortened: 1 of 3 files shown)


jstoobysmith:
(I'll wait until you ping me to review this :))

jnlindsay:
https://github.com/HEPLean/PhysLean/blob/3ec1551374bfe1370ea56d143f54fa8e4887677d/PhysLean/Relativity/PauliMatrices/Basic.lean#L95

I have failed to replace these `erw`s — any help is appreciated! I'm not sure how to get Lean to match `complexContrBasisFin4 0` with `S.basis c b0`, for example. Time to return to the Lean books.

I played around with `@[simp]`ing `fromTripleT_apply_basis` but haven't quite figured it out yet.

jstoobysmith:
I think the problem with this is that in this case `S.basis c b0` is defined using  `complexContrBasisFin4 0`. 
So I think we should write some lemmas which relate the two, and then we can use rw to turn them into a form `fromTripleT_apply_basis` can handle.

jstoobysmith:
I'll have a go at something so you can see what I mean / so I can test if it works.

jstoobysmith:
See my commit :). 

You can tell Lean to look for these things by itself - in fact this is exactly what `erw` is doing. The problem is that it trys to  find everything like this, which is very time consuming.

jstoobysmith:
(Given we don't have the `erw`s we can actually now shorten the proof as well).

Edit: Never mind I forgot I put `set_option maxHeartbeats 0 in` in there - which I've now removed, but couldn't make the proof shorter.

PhysLean/Relativity/PauliMatrices/Relations.lean
```diff
@@ -46,7 +46,8 @@ lemma pauliCo_contr_pauliContr :
     simp only [Nat.reduceAdd, Nat.succ_eq_add_one, Fin.isValue, Fin.succAbove_zero,
       Function.comp_apply, cons_val_zero, cons_val_one, head_cons, ofRat_basis_repr_apply]
     rw [← PhysLean.RatComplexNum.toComplexNum.map_mul]
-    erw [PhysLean.RatComplexNum.ofNat_mul_toComplexNum 2]
+    change (2 : ℕ) * _
+    rw [PhysLean.RatComplexNum.ofNat_mul_toComplexNum 2]
```
jnlindsay, line 50: Here `change` works, but I'm unsure why rewriting the lemma into
```Lean
lemma ofNat_mul_toComplexNum (n : ℕ) (a : RatComplexNum) :
    (n : ℂ) * toComplexNum a = toComplexNum (n * a) := by
  simp only [map_mul, map_natCast]
```
with the explicit `(n : ℂ)`, or even `↑n`, doesn't work. The tactic `erw?` says
```
❌ at reducible transparency,
  2
and
  ↑2
are not defeq, but they are at default transparency.
```

jstoobysmith: It's to do with the instances I think. The exact expression appearing in `pauliCo_contr_pauliContr` for the `2` is

```
@OfNat.ofNat ℂ 2 instOfNatAtLeastTwo
```

so this lemma would work 

```Lean
lemma ofNat_mul_toComplexNum' (n : ℕ) (a : RatComplexNum) :
    (@OfNat.ofNat ℂ n.succ.succ instOfNatAtLeastTwo) * toComplexNum a =
    toComplexNum (n.succ.succ * a) := by
  simp only [Nat.succ_eq_add_one,  map_mul, map_add, map_natCast,
    map_one, mul_eq_mul_right_iff]
  left
  exact rfl
```

But I think using `change` in this case is likely the better solution.

jstoobysmith (approved):
Approved. Many thanks for this :).

PR #506, opened 2025-04-21
chore: clean up proofs

zhikaip:
cleaned up some proofs in PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
1. remove unused variables
2. naming of fderiv_pi'
3. include `coord` and `basis` in `unfold` line rather than simp

changed Time.deriv definition to `t` for clarity


jstoobysmith (approved):
Thanks. These all look good to me - approved :).

PR #508, opened 2025-04-22
feat(jnlindsay): restricted Lorentz group

jnlindsay:
- [x] Definition of the restricted Lorentz group as a set
- [x] Show the restricted Lorentz group is a subgroup of the Lorentz group
  - [ ]  Maybe also: connected, normal subgroup

### DIFF
--- a/PhysLean/Relativity/Lorentz/Group/Basic.lean
+++ b/PhysLean/Relativity/Lorentz/Group/Basic.lean
@@ -107,7 +107,8 @@ instance lorentzGroupIsGroup : Group (LorentzGroup d) where
   inv_mul_cancel A := Subtype.eq (LorentzGroup.mem_iff_dual_mul_self.mp A.2)
 
 lemma inv_eq_dual (Λ : LorentzGroup d) :
-  (Λ⁻¹ : LorentzGroup d) = minkowskiMatrix.dual Λ.1 := by rfl
+    (Λ⁻¹ : LorentzGroup d) = minkowskiMatrix.dual Λ.1 := by
+  rfl
 
 /-- `LorentzGroup` has the subtype topology. -/
 instance : TopologicalSpace (LorentzGroup d) := instTopologicalSpaceSubtype
--- a/PhysLean/Relativity/Lorentz/Group/Basic.lean
+++ b/PhysLean/Relativity/Lorentz/Group/Basic.lean
@@ -107,7 +107,7 @@ instance lorentzGroupIsGroup : Group (LorentzGroup d) where
   inv_mul_cancel A := Subtype.eq (LorentzGroup.mem_iff_dual_mul_self.mp A.2)
 
 lemma inv_eq_dual (Λ : LorentzGroup d) :
-  (Λ⁻¹ : LorentzGroup d) = minkowskiMatrix.dual Λ := by rfl
+  (Λ⁻¹ : LorentzGroup d) = minkowskiMatrix.dual Λ.1 := by rfl
 
 /-- `LorentzGroup` has the subtype topology. -/
 instance : TopologicalSpace (LorentzGroup d) := instTopologicalSpaceSubtype
--- a/PhysLean/Relativity/Lorentz/Group/Basic.lean
+++ b/PhysLean/Relativity/Lorentz/Group/Basic.lean
@@ -106,6 +106,9 @@ instance lorentzGroupIsGroup : Group (LorentzGroup d) where
   inv A := ⟨minkowskiMatrix.dual A.1, LorentzGroup.dual_mem A.2⟩
   inv_mul_cancel A := Subtype.eq (LorentzGroup.mem_iff_dual_mul_self.mp A.2)
 
+lemma inv_eq_dual (Λ : LorentzGroup d) :
+  (Λ⁻¹ : LorentzGroup d) = minkowskiMatrix.dual Λ := by rfl
+
 /-- `LorentzGroup` has the subtype topology. -/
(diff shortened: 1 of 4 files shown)


jnlindsay:
TODO
- [x] Figure out why the linter doesn't like `instance`

PhysLean/Relativity/Lorentz/Group/Basic.lean
```diff
@@ -106,6 +106,10 @@ instance lorentzGroupIsGroup : Group (LorentzGroup d) where
   inv A := ⟨minkowskiMatrix.dual A.1, LorentzGroup.dual_mem A.2⟩
   inv_mul_cancel A := Subtype.eq (LorentzGroup.mem_iff_dual_mul_self.mp A.2)
 
+lemma inv_eq_dual (Λ : LorentzGroup d) :
```
jnlindsay, line 109: Is this lemma redundant?

jstoobysmith: Did you need it for anything? I think Lean might already reduce the `Λ⁻¹` to `minkowskiMatrix.dual Λ.1`.

PhysLean/Relativity/Lorentz/Group/Proper.lean
```diff
@@ -176,6 +176,11 @@ instance : DecidablePred (@IsProper d) := by
   intro Λ
   apply Real.decidableEq
 
+/-- The product of two proper Lorentz transformations is proper. -/
+lemma mul_proper_of_proper_proper {Λ Λ' : LorentzGroup d} (h : IsProper Λ) (h' : IsProper Λ') :
```
jstoobysmith: maybe mul_isProper_of_isProper_isProper

PhysLean/Relativity/Lorentz/Group/Restricted.lean
```diff
@@ -3,25 +3,59 @@ Copyright (c) 2024 Joseph Tooby-Smith. All rights reserved.
...
+/-- The restricted Lorentz group comprises the proper and orthochronous elements of the
+Lorentz group. -/
+def Restricted (d : ℕ) : Set (LorentzGroup d) :=
+  { Λ : LorentzGroup d | IsProper Λ ∧ IsOrthochronous Λ }
+
+open Matrix
+open minkowskiMatrix
+
+/-- The restricted Lorentz group is a subgroup of the Lorentz group. -/
+def RestrictedIsSubgroup {d : ℕ} : Subgroup (LorentzGroup d) where
```
jstoobysmith: I think it would be better to call this `restricted`, and delete the definition `Restricted` above.

PhysLean/Relativity/Lorentz/Group/Restricted.lean
```diff
@@ -3,25 +3,59 @@ Copyright (c) 2024 Joseph Tooby-Smith. All rights reserved.
...
+    have h_η₀₀ : @minkowskiMatrix d (Sum.inl 0) (Sum.inl 0) = 1 := by rfl
+    have h_dual : (dual Λ.1) (Sum.inl 0) (Sum.inl 0) = Λ.1 (Sum.inl 0) (Sum.inl 0) := by
+      rw [dual_apply, h_η₀₀, one_mul, mul_one]
+
+    exact ⟨
+      by rw [IsProper, inv_eq_dual, det_dual, Λ_proper],
+      by rw [IsOrthochronous, inv_eq_dual, h_dual]; exact Λ_ortho⟩
+
+/-- The restricted Lorentz group is a group. -/
+instance restrictedIsGroup {d : ℕ} : Group (Restricted d) :=
```
jstoobysmith: If you follow my previous comment I don't think this will be needed.

jnlindsay: I also got rid of the `Coe` because it seems Lean does it automatically?

PhysLean/Relativity/Lorentz/Group/Basic.lean
jnlindsay: I do use it, and `det_dual : ... (dual Λ).det = Λ.det` seems to only pattern-match `dual ↑Λ` but not `↑Λ⁻¹`.

PhysLean/Relativity/Lorentz/Group/Restricted.lean
jnlindsay: I've used `Restricted` (uppercase R), according to the informal definition :) let me know if I should change it

jstoobysmith: The general convention is that all definitions are lower case, except types and props which are upper case. 

I actually don't know about `Restricted` because it gets cast as a type, but strictly isn't. So am happy to keep it as is or change to lower case.

PhysLean/Relativity/Lorentz/Group/Restricted.lean
jstoobysmith: Yeah I think it will do this automatically:)

PhysLean/Relativity/Lorentz/Group/Basic.lean
jstoobysmith: Ah, ok. Yeah I don't think there is much harm in keeping this then.

jstoobysmith (approved):
I've approved these changes so feel free to merge when you are ready.

jstoobysmith:
Ok, looking at convention in Mathlib for e.g. Matrix.specialUnitaryGroup, they seem to use lower case for `restricted`, so maybe it would be wise if we followed - we can change this in a future PR

jstoobysmith:
Many thanks for this :)

PR #516, opened 2025-04-25
Feat: time deriv and curl commute

zhikaip:
This is still very messy, I'll slowly clean this up. 

A few questions:
1. should this be its own file or should this go into VectorIdentities? my concern is that VectorIdentities is space only, maybe it (and/or this new file) should be moved to somewhere else but still in `ClassicalMechanics`.
2. Related to above, `fderiv_uncurry` was a private lemma and I think this means it can't be used outside of the file? I removed the `private` for now, but maybe it should also sit somewhere else (`TimeSlice` is currently in `Relativity.SpaceTime`)


jstoobysmith:
Hey, to answer your questions: 

> should this be its own file or should this go into VectorIdentities? my concern is that VectorIdentities is space only, maybe it (and/or this new file) should be moved to somewhere else but still in ClassicalMechanics.

If we can think of a good folder/file name then I would be up for having it somewhere else.  Something that e.g. includes the notion of space and time but isn't `Spacetime` (for obvious reasons) would be nice. One thought I had was having a directory called `Worldvolume` but this terminology may be a bit niche. 
But I also think it's ok where it is for now.

> Related to above, fderiv_uncurry was a private lemma and I think this means it can't be used outside of the file? I removed the private for now, but maybe it should also sit somewhere else (TimeSlice is currently in Relativity.SpaceTime)

Yes this is what private means. 
We could create a file for these little generic `fderiv` lemmas in the Mathematics directory.

PR #517, opened 2025-04-25
feat : create file for fderiv_curry functions

zhikaip:
Moved all the fderiv functions to this new file in `Mathematics`.

Maybe I'll put `fderiv_swap` here as well, but I just want to check this setup is fine.

Also is there a particular reason why you used `N W M`? I changed everything to `X Y Z` for now, but let me know if you want to change this back


PhysLean/Mathematics/FDerivCurry.lean
```diff
@@ -0,0 +1,97 @@
+/-
+Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joseph Tooby-Smith
```
jstoobysmith: If you are happy with this, I think it would be good if you added your name to the author list as well.

PhysLean/Mathematics/FDerivCurry.lean
```diff
@@ -0,0 +1,97 @@
...
+  exact h
+
+lemma fderiv_uncurry_sum (f : X → Y → Z) (xy dxy : X × Y) :
+    fderiv 𝕜 (↿f) xy dxy
+    =
+    fderiv 𝕜 (↿f) xy (dxy.1, 0) + fderiv 𝕜 (↿f) xy (0, dxy.2) := by
+  rw [← ContinuousLinearMap.map_add]
+  simp
+
+theorem fderiv_uncurry'' (f : X → Y → Z) (hf : Differentiable 𝕜 (↿f)) :
```
jstoobysmith: Better to have these as `lemma` rather than `theorem` I think. 
Also would be nice if we could think of another name for them rather than relying on primes.

jstoobysmith:
I probably choose `N W M` because I was thinking of manifolds. But `X Y Z` seems better to me.

PhysLean/Mathematics/FDerivCurry.lean
zhikaip: should I add Tomas as well?

PhysLean/Mathematics/FDerivCurry.lean
zhikaip: maybe `fderiv_curry_fst` and `fderiv_curry_snd` for the ones originally you wrote in `TimeSlice`? then something like `fderiv_uncurry_clm_comp` and so on

jstoobysmith: Yeah, I think this would work.

PhysLean/Mathematics/FDerivCurry.lean
jstoobysmith: @lecopivo Hi Tomas,  given your help on the zulip, can we add you as an author to this file?

lecopivo: Sure, I have no problem with that

PhysLean/Mathematics/FDerivCurry.lean
lecopivo: Why do you have them as two theorems anyway? They can be merged into one theorem by adding them together.

jstoobysmith: Do you mean the following. Instead of `fderiv_uncurry` and `fderiv_uncurry'` have something like: 

```lean
 fderiv 𝕜 (↿f) xy (dx, dy) = ...
 ```
 
 and prove it using the existing proofs and e.g. 
 
 ```lean
 fderiv 𝕜 (↿f) xy (dx, dy) = fderiv 𝕜 (↿f) xy (dx, 0) +  fderiv 𝕜 (↿f) xy (0, dy) 
 ```

Comment (lecopivo) [reply PhysLean/Mathematics/FDerivCurry.lean]: Yes exactly, with the only difference that lhs should be
```
fderiv 𝕜 (↿f) xy dxy = ...
```
Having `dxy` instead of `(dx, dy)` is more flexible/general

Comment (zhikaip) [reply PhysLean/Mathematics/FDerivCurry.lean]: I'll try to refactor this

Comment (lecopivo) [reply PhysLean/Mathematics/FDerivCurry.lean]: In my opinion, the current theorem `fderiv_uncurry_sum` should be called `fderiv_uncurry` and it is already in the form I'm was advocating.

Comment (jstoobysmith) [reply PhysLean/Mathematics/FDerivCurry.lean]: `fderiv_uncurry_sum` is just a statement about linearity of fderiv. Not quite sure it is what you where advocating for.

Comment (lecopivo) [reply PhysLean/Mathematics/FDerivCurry.lean]: Ohh sorry I misread the theorem statement, ignore what I said :)

**Comment by zhikaip** [comment]
updated and added general form of `fderiv_swap`, tried my best to tidy up the helper lemmas which should be better once Tomas' mathlib PR goes through. added a TODO item for this.

File: PhysLean/Mathematics/FDerivCurry.lean [hunk 3]
```diff
@@ -0,0 +1,249 @@
+/-
+Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joseph Tooby-Smith, Zhi Kai Pong, Tomáš Skřivan
```
jstoobysmith: Would you mind putting the authors in alphabetical order by last name? (you first, then Tomas, then me)
This tends to be the convention we follow.

PhysLean/Mathematics/FDerivCurry.lean
```diff
@@ -0,0 +1,249 @@
...
+lemma ContDiff2.fderiv_differentiable (f : X → Y → Z) (hf : ContDiff 𝕜 2 ↿f) :
+    Differentiable 𝕜 (fderiv 𝕜 (↿f)) := by
+  change ContDiff 𝕜 (1 + 1) ↿f at hf
+  rw [contDiff_succ_iff_fderiv] at hf
+  have hd2 := hf.2.2
+  apply ContDiff.differentiable hd2
+  rfl
+
+def inclX (y' : Y) : X → X × Y := fun x' => (x', y')
+def inclY (x' : X) : Y → X × Y := fun y' => (x', y')
```
jstoobysmith: I know I defined these, but I think it is better to use predefined Mathlib functions instead of these. 
I think [Prod.mk](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Prod.mk) for the second, and `flip (Prod.mk)` for the first.

PhysLean/Mathematics/FDerivCurry.lean
```diff
@@ -0,0 +1,249 @@
...
+  · apply Differentiable.comp
+    · apply ContDiff2.fderiv_differentiable
+      exact hf
+    · fun_prop [inclY]
+  · conv_rhs =>
+      enter [y]
+      rw [fderiv_inr_fst_clm]
+    fun_prop
+
+lemma fderiv_uncurry_differentiable_x_inclX_y (f : X → Y → Z) (x : X) (hf : ContDiff 𝕜 2 ↿f) :
```
jstoobysmith: The statement of this lemma doesn't depend on inclX, so maybe it is best not to have it in the name.
Likewise with `x` and `y` which are really free variables.

PhysLean/Mathematics/FDerivCurry.lean
```diff
@@ -0,0 +1,249 @@
...
+    fderiv 𝕜 f x dx y
+    =
+    fderiv 𝕜 (f · y) x dx := by
+  rw [fderiv_clm_apply] <;> first | simp | fun_prop
+
+TODO "AZ2QN" "Replace following helper lemmas with fun_prop after
+    #24056 in Mathlib has gone through."
+
+/-- Helper lemmas showing differentiability from ContDiff 𝕜 2 ↿f. -/
+lemma ContDiff2.differentiable (f : X → Y → Z) (hf : ContDiff 𝕜 2 ↿f) :
```
jstoobysmith: I think `ContDiff.two_differentiable` is better here, and similar in the next one. Then you can use the field (dot) notation with these lemmas.

jstoobysmith:
Sorry a couple more minor things regarding names. I really appreciate all the work and time you have spent on this and learning about derivatives etc, so thank you! I hope it will become very fruitful as PhysLean grows :).

PhysLean/Mathematics/FDerivCurry.lean
zhikaip: Not totally sure about this one, please check if the new names are better

jstoobysmith (approved):
Many thanks - I've approved this for merger :)

PhysLean/Mathematics/FDerivCurry.lean
jstoobysmith: Looks good to me for now - we can always change it if we come up with something better in the future.

PR #520, opened 2025-04-28
feat: setting up homogeneous media for EM

zhikaip:
attempt at setting this up.  thought process is as follows:

1. redefine Maxwell's with respect to bundled `ChargedMedium` structure as `GaussLawElectricHomogeneous`, `GaussLawMagneticHomogeneous` etc.
2. show that this is equivalent to original formulation with the variables from bundled structure
3. we can then "embed" (not sure what's the right word) optical medium into charged medium and obtain charge and current free Maxwell's equations

Please check if this makes sense and feel free to make suggestions :)

### DIFF
--- a/PhysLean/Electromagnetism/Basic.lean
+++ b/PhysLean/Electromagnetism/Basic.lean
@@ -28,4 +28,35 @@ open realLorentzTensor
 /-- The vector potential of an electromagnetic field-/
 abbrev VectorPotential (d : ℕ := 3) := SpaceTime d → ℝT[d, .up]
 
+/-- The electric permittivity and the magnetic permeability of free space. -/
+structure EMSystem where
+  /-- The permittivity of free space. -/
+  ε₀ : ℝ
+  /-- The permeability of free space. -/
+  μ₀ : ℝ
+
+TODO "6V2UZ" "Charge density and current density should be generalized to signed measures,
+  in such a way
+  that they are still easy to work with and can be integrated with with tensor notation.
+  See here:
+  https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Maxwell's.20Equations"
+
+/-- The charge density. -/
+abbrev ChargeDensity := Time → Space → ℝ
+
+/-- Current density. -/
+abbrev CurrentDensity := Time → Space → EuclideanSpace ℝ (Fin 3)
+
+namespace EMSystem
+variable (𝓔 : EMSystem)
+open SpaceTime
+
+/-- The speed of light. -/
+noncomputable def c : ℝ := 1/(√(𝓔.μ₀ * 𝓔.ε₀))
+
+/-- Coulomb's constant. -/
+noncomputable def coulombConstant : ℝ := 1/(4 * Real.pi * 𝓔.ε₀)
+
+end EMSystem
+
 end Electromagnetism
--- a/PhysLean/Electromagnetism/Homogeneous.lean
+++ b/PhysLean/Electromagnetism/Homogeneous.lean
@@ -63,39 +63,35 @@ variable (OM : OpticalMedium)
 local notation "ε" => OM.ε
 local notation "μ" => OM.μ
 
-/-- An optical medium which is charge and current free. -/
-def IsFree (OM : OpticalMedium)
+/-- The Maxwell equations for charge and current free medium. -/
+def OpticalMedium.FreeMaxwellEquations (OM : OpticalMedium)
     (E : ElectricField) (B : MagneticField) : Prop :=
-    OM.free.ρ = 0 ∧ OM.free.J = 0 ∧
-    MaxwellEquations OM OM.free.ρ OM.free.J E B
+  MaxwellEquations OM OM.free.ρ OM.free.J E B
 
-/-- Maxwell's equations for charge and current free medium. -/
 theorem OpticalMedium.gaussLawElectric_of_free (E : ElectricField) (B : MagneticField)
-    (h : IsFree OM E B) :
+    (h : OM.FreeMaxwellEquations E B) :
     ε * (∇ ⬝ E t) x = 0 := by
-  have h' := h.2.2.1
-  unfold GaussLawElectric at h'
-  rw [h.1] at h'
-  simp_all
+  have h' := h.1
+  rw [h']
+  rfl
 
-theorem OpticalMedium.gaussLawMagnetic_of_isfree (E : ElectricField) (B : MagneticField)
-    (h : IsFree OM E B) :
+theorem OpticalMedium.gaussLawMagnetic_of_free (E : ElectricField) (B : MagneticField)
+    (h : OM.FreeMaxwellEquations E B) :
     (∇ ⬝ B t) x = 0 := by
-  have h' := h.2.2.2.1
-  unfold GaussLawMagnetic at h'
+  have h' := h.2.1
   rw [h']
 
-theorem OpticalMedium.ampereLaw_of_isfree (E : ElectricField) (B : MagneticField)
-    (h : IsFree OM E B) :
+theorem OpticalMedium.ampereLaw_of_free (E : ElectricField) (B : MagneticField)
+    (h : OM.FreeMaxwellEquations E B) :
     (∇ × B t) x = μ • ε • ∂ₜ (fun t => E t x) t := by
-  have h' := h.2.2.2.2.1
-  unfold AmpereLaw at h'
-  rw [h.2.1] at h'
+  have h' := h.2.2.1
+  rw [h']
   simp_all
+  right
+  rfl
 
-theorem OpticalMedium.faradayLaw_of_isfree (E : ElectricField) (B : MagneticField)
-    (h : IsFree OM E B) :
+theorem OpticalMedium.faradayLaw_of_free (E : ElectricField) (B : MagneticField)
+    (h : OM.FreeMaxwellEquations E B) :
     (∇ × E t) x = - ∂ₜ (fun t => B t x) t := by
-  have h' := h.2.2.2.2.2
-  unfold FaradayLaw at h'
+  have h' := h.2.2.2
   rw [h']
--- a/PhysLean/Electromagnetism/Homogeneous.lean
+++ b/PhysLean/Electromagnetism/Homogeneous.lean
@@ -17,12 +17,6 @@ PhysLean.Electromagnetism.MaxwellEquations.
 
 namespace Electromagnetism
 
-namespace Homogeneous
-/-- Optical Medium has scalar electric permittivity as a function of space. -/
-structure OpticalMedium extends EMSystem where
-  /-- The electric permittivity of the material. -/
-  ε : Space → ℝ
-
 /-- Charged Medium is defined as Optical Medium with charge and current. -/
 structure ChargedMedium extends OpticalMedium where
   /-- The charge density. -/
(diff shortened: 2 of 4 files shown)


PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -0,0 +1,118 @@
...
+
+-/
+
+namespace Electromagnetism
+
+namespace Homogeneous
+/-- Optical Medium has scalar electric permittivity as a function of space. -/
+structure OpticalMedium extends EMSystem where
+  /-- The electric permittivity of the material. -/
+  ε : Space → ℝ
```
jstoobysmith: This definition of `OpticalMedium` has two electric permittivities, one from `EMSystem` and one from `ε`. I think it might be best to rename `EMSystem` as `OpticalMedium`.

PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -0,0 +1,118 @@
...
+  /-- The charge density. -/
+  ρ : ChargeDensity
+  /-- The current density. -/
+  J : CurrentDensity
+
+open Space
+open Time
+
+/-- Gauss's law for the Electric field in homogeneous medium. -/
+def GaussLawElectricHomogeneous (CM : ChargedMedium) (E : ElectricField) : Prop :=
```
jstoobysmith: I would redo these definitions as e.g. 

```
abbrev ChargedMedium.GaussLawElectric (E : ElectricField) : Prop := GaussLawElectric CM.toOpticalMedium CM.ρ
```

The use of `abbrev` allows Lean to 'see through' the definition and treat this just as `GaussLawElectric` - which will mean we don't have to prove two sets of lemmas.

PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -0,0 +1,118 @@
...
+  ρ := fun _ _ => 0
+  J := fun _ _ => 0
+
+variable (OM : OpticalMedium)
+
+local notation "ε₀" => OM.free.ε₀
+local notation "μ₀" => OM.free.μ₀
+
+/-- Maxwell's equations for free space. -/
+theorem GaussLawElectricFree (E : ElectricField)
```
jstoobysmith: maybe a better name is `ChargedMedium.gaussLawElectric_of_free`.

PhysLean/Electromagnetism/Homogeneous.lean
zhikaip: do we want to change the  `ε₀` in the original Maxwell's to `ε` then?

jstoobysmith: I think this would make sense. Same with mu

zhikaip: should `OpticalMedium` also sit in a `Homogeneous` namespace or do you think we don't actually need `Homogeneous`?

jstoobysmith: IMO I don't think we need the `Homogeneous` namespace, but maybe making it clear that this is the Homogeneous case in a doc-string or something would be good.

PhysLean/Electromagnetism/MaxwellEquations.lean
```diff
@@ -8,60 +8,39 @@ import PhysLean.Electromagnetism.Basic
...
+  ∀ t : Time, ∀ x : Space, ε x * (∇ ⬝ E t) x = ρ t x
 
 /-- Gauss's law for the Magnetic field. -/
 def GaussLawMagnetic (B : MagneticField) : Prop :=
   ∀ t : Time, ∀ x : Space, (∇ ⬝ B t) x = 0
 
 /-- Ampère's law. -/
 def AmpereLaw (E : ElectricField) (B : MagneticField) : Prop :=
-  ∀ t : Time, ∀ x : Space, (∇ × B t) x = μ₀ • (J t x + ε₀ • ∂ₜ (fun t => E t x) t)
+  ∀ t : Time, ∀ x : Space, (∇ × B t) x = μ x • (J t x + ε x • ∂ₜ (fun t => E t x) t)
```
jstoobysmith, line 45: I think with Maxwell's equations in an inhomogeneous medium, the epsilons and mu's need to be inside the derivatives. See e.g.:

https://physics.stackexchange.com/questions/193543/

PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -0,0 +1,93 @@
...
+  rfl
+
+theorem ChargedMedium.faradayLaw_of_free (E : ElectricField) (B : MagneticField)
+    (h : OM.free.FaradayLaw E B) :
+    (∇ × E t) x = - ∂ₜ (fun t => B t x) t := by
+  rw [h]
+
+def IsFree (OM : OpticalMedium) (E : ElectricField) (B : MagneticField) : Prop :=
+    OM.free.GaussLawElectric E ∧ OM.free.GaussLawMagnetic B ∧
+    OM.free.AmpereLaw E B ∧ OM.free.FaradayLaw E B
```
zhikaip: I want to be able to write `h : IsFree OM E B` to indicate when an `OpticalMedium` is charge and current free, and then use the `_of_free` theorems without having to worry about the individual assumptions. Is this the right way to set this up? I saw that there is `class` but I'm not sure what's the difference

PhysLean/Electromagnetism/MaxwellEquations.lean
zhikaip: I'm concerned about discontinuities at the boundary, I understand this is something we want to proof anyway but is this really the correct way to set up optical medium (i.e. having a step function for `ε`)?

PhysLean/Electromagnetism/Homogeneous.lean
jstoobysmith: Yes this is the right way of doing this.  The thing `class` is used for things like the structure of a vector space etc. which Lean infers automatically. 

Your description of `IsFree` here though does not match the definition you have made.

PhysLean/Electromagnetism/MaxwellEquations.lean
jstoobysmith: I'm not entirely sure. I think either step functions for `ε` say, or restriction of the domain of `x` to some sub manifold of Space.

jstoobysmith: I think to do this properly we will have to take care of surface current and charge densities etc. 

See e.g. 
https://en.wikipedia.org/wiki/Interface_conditions_for_electromagnetic_fields.

I guess what we currently have is the 'bulk' Maxwell's equations.

jstoobysmith: I think if this is the direction you want to go in, formalizing the notion of an `interface` in space may be a good starting point. I'm not sure how easy this would be though.

jstoobysmith: Depending on your plan after this PR, it might be easiest at this point to remove the space dependency  on epsilon and mu.

PhysLean/Electromagnetism/Homogeneous.lean
zhikaip: If I'm understanding this correctly `IsFree` should just be 
`
def IsFree (OM : OpticalMedium) : Prop :=
    OM.free.ρ = 0 ∧ OM.free.J = 0
`
right?

what do I need to further define to use `IsFree OM` to get `OM.free.AmpereLaw E B`?

or should I use `IsFree OM` as an assumption above instead?

jstoobysmith: If I understand correctly what you are wanting, then I think you want:


```
def IsFree (CM : ChargedMedium) : Prop := CM.free.ρ = 0 ∧ CM.free.J = 0
```

Then you can prove: 

```
lemma free_isFree (OM : OrderedMedium) : IsFree OM.free := by sorry 
```

Then, use `IsFree CM` as an assumption instead, in e.g. `gaussLawMagnetic_of_free` (though rename as 
`gaussLawMagnetic_of_isFree` (which I would also actually make an iff statement). 

I'm not sure if this gives you exactly what you are wanting, if not could you elaborate a bit more?

zhikaip: I've made an attempt at this which I hope is correct, please have a look at the update when you have time, thanks!

PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -0,0 +1,101 @@
...
+variable (OM : OpticalMedium)
+
+local notation "ε" => OM.ε
+local notation "μ" => OM.μ
+
+/-- An optical medium which is charge and current free. -/
+def IsFree (OM : OpticalMedium)
+    (E : ElectricField) (B : MagneticField) : Prop :=
+    OM.free.ρ = 0 ∧ OM.free.J = 0 ∧
+    MaxwellEquations OM OM.free.ρ OM.free.J E B
```
jstoobysmith: The problem with this definition (IMO) is that `OM.free.ρ = 0` and `OM.free.J = 0` hold by definition. 

Your doc-string says:
```
 An optical medium which is charge and current free.
 ```
 But an optical medium (as you have defined it) does not have attached to it a current or a charge. 
 
 I now think I understand what you are wanting to do though. 
 
 How about the following:
 
 ```
  /-- The Maxwell equations in free space. --/
  def OpticalMedium.FreeMaxwellEquations (OM : OpticalMedium) (E : ElectricField) (B : MagneticField) : Prop :=
    MaxwellEquations OM OM.free.ρ OM.free.J E B
   ```

File: PhysLean/Electromagnetism/MaxwellEquations.lean (hunk 4 above)
Comment (jstoobysmith) [reply PhysLean/Electromagnetism/MaxwellEquations.lean:45]: (Resolving this conversation to clean this PR up a bit).

File: PhysLean/Electromagnetism/Homogeneous.lean (hunk 6 above)
Comment (zhikaip) [reply PhysLean/Electromagnetism/Homogeneous.lean]: thanks, I think this is what I want! I've updated the branch :)

File: PhysLean/Electromagnetism/Basic.lean [hunk 7]
```diff
@@ -28,4 +28,35 @@ open realLorentzTensor
 /-- The vector potential of an electromagnetic field-/
 abbrev VectorPotential (d : ℕ := 3) := SpaceTime d → ℝT[d, .up]
 
+/-- The electric permittivity and the magnetic permeability of free space. -/
+structure EMSystem where
```
jstoobysmith, line 32: We can probably remove EMSystem totally - I don't think it is used for anything now?
Otherwise, this looks good to me! Thanks.

zhikaip: It isn't being used, but I think it's still a good idea to have those constants defined, and I think `Electromagnetism.Basic` is a good place for them?

zhikaip: or is the rationale to define them each time in a structure when required instead of having "global" constants?

jstoobysmith: You are viewing the difference between `EMSystem` and `OpticalMedium` as `EMSystem`  corresponding specifically to free space, whilst `OpticalMedium` corresponding to some medium, which is not necessarily free space?

If so this is ok, but I think we should make it clear in the doc-string for `OpticalMedium`.

zhikaip: yes that's my idea. is the new doc-string ok?

jstoobysmith (approved):
All looks good - I've approved so feel free to merge.

PR #524, opened 2025-04-29
feat: restricted Lorentz group equal to identity component

jnlindsay:


PhysLean/Relativity/Lorentz/Group/Orthochronous.lean
```diff
@@ -219,6 +223,28 @@ lemma IsOrthochronous.iff_inv_isOrthochronous {Λ : LorentzGroup d} :
...
     MonoidHom.mem_ker, MonoidHom.mem_ker, orthchroRep_inv_eq_self]
 
+/-- Two Lorentz transformations are both orthochronous or both not orthochronous if they are mapped
+to the same element via the homomorphism from `LorentzGroup` to `ℤ₂`. -/
+lemma what_to_call_this? {Λ Λ' : LorentzGroup d} (h : orthchroMap Λ = orthchroMap Λ') :
+    IsOrthochronous Λ ↔ IsOrthochronous Λ' := by
+  rw [IsOrthochronous.iff_in_orthchroRep_ker, IsOrthochronous.iff_in_orthchroRep_ker]
+  rw [MonoidHom.mem_ker, MonoidHom.mem_ker]
+  change orthchroMap Λ = 1 ↔ orthchroMap Λ' = 1
+  rw [h]
```
jnlindsay: @jstoobysmith Any suggestions on what to name this lemma?

jstoobysmith: Maybe we could call it something like `isOrthochronous_iff_of_orthchroMap_eq`

jnlindsay:
TODO:
- [ ] Show restricted Lorentz group is a normal subgroup with the first isomorphism theorem.

jnlindsay:
(rebase)

jnlindsay:
https://github.com/HEPLean/PhysLean/blob/b9776ffbc7ddcc69b5a3d65459190989ecbd5ea8/PhysLean/Relativity/Lorentz/Group/Boosts/Generalized.lean#L13-L14

@jstoobysmith Is it accurate to call these "four velocity" vectors?

jstoobysmith:
Do you not like the `Four` or the `Velocity`? if the former then I think `Lorentz Velocity` may be better. I think they can be interpreted as velocity, because there norm is restricted to be `1`. Really I think one could develop the theory around Lorentz Velocities as a separate API and then connect it to this. But that might be a bit of work.

jnlindsay:
> Do you not like the `Four` or the `Velocity`? if the former then I think `Lorentz Velocity` may be better. I think they can be interpreted as velocity, because there norm is restricted to be `1`. Really I think one could develop the theory around Lorentz Velocities as a separate API and then connect it to this. But that might be a bit of work.

Ahh sorry I meant the "four" bit, seeing as we're in $d + 1$ (or $d$? forget) dimensions. According to [this Wiki page](https://en.wikipedia.org/wiki/Four-vector) it would have four components, right?

jstoobysmith:
Yeah! so I agree in `d + 1` spacetime-dimensions (which is what this is) then `Lorentz Velocity` would be better then `Four velocity` as it doesn't have 4-components in general, only in `3 + 1`.

jstoobysmith (approved):

PR #533, opened 2025-05-01
feat: api and fixes

zhikaip:
adding some more API for what I need

I added `ε > 0 ∧ μ > 0` to `MaxwellEquations` for now, should those sit inside `OpticalMedium` instead? not sure which way is better

### DIFF
--- a/PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
+++ b/PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
@@ -221,17 +221,20 @@ lemma curl_of_curl (f : Space → EuclideanSpace ℝ (Fin 3)) (hf : ContDiff ℝ
       try apply contDiff_euclidean.mp
       exact hf
 
+@[simp]
 lemma grad_zero : ∇ (0 : Space d → ℝ) = 0 := by
   unfold grad Space.deriv
   simp only [fderiv_zero, Pi.zero_apply, ContinuousLinearMap.zero_apply]
   rfl
 
+@[simp]
 lemma div_zero : ∇ ⬝ (0 : Space d → EuclideanSpace ℝ (Fin d)) = 0 := by
   unfold div Space.deriv Finset.sum
   simp only [Pi.zero_apply, fderiv_const, ContinuousLinearMap.zero_apply, Multiset.map_const',
     Finset.card_val, Finset.card_univ, Fintype.card_fin, Multiset.sum_replicate, smul_zero]
   rfl
 
+@[simp]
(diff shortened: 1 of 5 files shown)


jstoobysmith:
Would probably be better in `OpticalMedium` I think.

PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
```diff
@@ -221,6 +221,24 @@ lemma curl_of_curl (f : Space → EuclideanSpace ℝ (Fin 3)) (hf : ContDiff ℝ
       try apply contDiff_euclidean.mp
       exact hf
 
+lemma grad_zero : ∇ (0 : Space d → ℝ) = 0 := by
+  unfold grad Space.deriv
```
jstoobysmith, line 226: If you can replace the `unfold`'s with `rw [grad, Space.deriv]` and similar with the below, this is probably better, but no worries if not. 

The tactic `unfold` can be a bit annoying when one has to refactor as it depends on the exact implementation of the quantities.

PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
```diff
@@ -221,6 +221,24 @@ lemma curl_of_curl (f : Space → EuclideanSpace ℝ (Fin 3)) (hf : ContDiff ℝ
...
+  simp only [fderiv_zero, Pi.zero_apply, ContinuousLinearMap.zero_apply]
+  rfl
+
+lemma div_zero : ∇ ⬝ (0 : Space d → EuclideanSpace ℝ (Fin d)) = 0 := by
```
jstoobysmith, line 231: We can probably add all these `_zero` lemmas to the simp database. 
That is, in the line above each of them write `@[simp]`, then when you use `simp` in the future it will check if these results can be applied.

PhysLean/ClassicalMechanics/Space/VectorIdentities.lean
zhikaip: hmm I tried but doesn't really work, either the `Space.deriv` and dependent definitions can't be rewritten, or some of them are nested inside another function that `rw` can't see through

jstoobysmith: Ok, no worries, there are probably appropriate lemmas we can include (e.g. a trivial `Space.deriv_def`), but maybe its not worth worrying to much about this.

jstoobysmith (approved):
Approved - feel free to merge.

PR #544, opened 2025-05-06
feat: plane wave

zhikaip:
I hope I'm stating the right thing here.

defined the travelling plane wave and prove it satisfies the wave equation.

### DIFF
--- a/PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
+++ b/PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
@@ -80,7 +80,8 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner s s = 1)
     not_false_eq_true, pow_eq_zero_iff]
   right
   ext i
-  have hd_inner (x : EuclideanSpace ℝ (Fin d)): DifferentiableAt ℝ (fun x => inner x s - c * t) x := by
+  have hd_inner (x : EuclideanSpace ℝ (Fin d)):
+      DifferentiableAt ℝ (fun x => inner x s - c * t) x := by
     apply DifferentiableAt.sub
     apply DifferentiableAt.inner
     repeat fun_prop
--- a/PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
+++ b/PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
@@ -79,7 +79,7 @@ lemma wave_dx {u v : Fin d} {f₀' : ℝ → ℝ →L[ℝ] EuclideanSpace ℝ (F
     (fun x' => fderiv ℝ (fun x => (inner (f₀ (inner x s - c * t))
     (EuclideanSpace.single u 1))) x' (EuclideanSpace.single v 1))
     =
-    fun x' => (inner (f₀' (inner x' s - c * t) (s v)) (EuclideanSpace.single u 1):ℝ) := by
+    fun x' => (inner (f₀' (inner x' s - c * t) (s v)) (EuclideanSpace.single u 1) : ℝ) := by
   ext x'
   rw [fderiv_inner_apply]
   simp only [fderiv_const, Pi.zero_apply, ContinuousLinearMap.zero_apply, inner_zero_right,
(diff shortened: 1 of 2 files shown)


PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -17,7 +17,90 @@ namespace ClassicalMechanics
...
-/-- The general form of the wave equation where c is the propagation speed. -/
+/-- The general form of the wave equation where `c` is the propagation speed. -/
 def WaveEquation (f : Time → Space d → EuclideanSpace ℝ (Fin d))
-    (t : Time) (x : Space d) (c : ℝ) : Prop :=
-    c^2 • Δ (f t) x - ∂ₜ (fun t => (∂ₜ (fun t => f t x) t)) t = 0
+  (t : Time) (x : Space d) (c : ℝ) : Prop :=
+  c^2 • Δ (f t) x - ∂ₜ (fun t => (∂ₜ (fun t => f t x) t)) t = 0
+
+/-- A vector-valued plane wave travelling in the direction of `s` with propagation speed `c`. -/
+noncomputable def PlaneWave (f₀ : ℝ → EuclideanSpace ℝ (Fin d))
```
jstoobysmith: I think we should do lower case, i.e. `planeWave`, as not a prop or a type.

jstoobysmith:
This looks good to me, although I don't understand the following: 

Did you put a normalization condition on `s`? It looks like not? But then I don't understand how your plane wave satisfies the wave equation.

jstoobysmith:
I think your assumptions combine to imply that `f₀` has constant derivative. 
I think what you want is:

```lean
/-- The plane wave satisfies the wave equation. -/
theorem planeWave_isWave (c : ℝ) (s : Space d)
    (f₀ : ℝ → EuclideanSpace ℝ (Fin d)) (f₀' : ℝ → ℝ →L[ℝ] EuclideanSpace ℝ (Fin d))
    (h' : ∀ x, HasFDerivAt f₀ (f₀' x) x) :
    WaveEquation (PlaneWave f₀ c s) t x c := by
  unfold WaveEquation PlaneWave
  have hdt : ∂ₜ (fun t => f₀ (inner x s - c * t)) = fun t => -c • f₀' (inner x s - c * t) 1 := by
    unfold Time.deriv
    ext t i
    change fderiv ℝ (f₀ ∘ fun t => (inner x s - c * t)) t 1 i = _
    rw [fderiv_comp, fderiv_const_sub, fderiv_const_mul]
    simp only [fderiv_id', ContinuousLinearMap.comp_neg, ContinuousLinearMap.comp_smulₛₗ,
      RingHom.id_apply, ContinuousLinearMap.comp_id, ContinuousLinearMap.neg_apply,
      ContinuousLinearMap.coe_smul', Pi.smul_apply, PiLp.neg_apply, PiLp.smul_apply, smul_eq_mul,
      neg_mul, neg_inj, mul_eq_mul_left_iff]
    left
    rw [HasFDerivAt.fderiv (h' (inner x s - c * t))]
    repeat fun_prop
  conv_lhs =>
    enter [2, 1, t]
    rw [hdt]
  sorry
 ```

**Comment by zhikaip** [comment]
I think for the general from the normalization isn't necessary, any vector will do as long as `inner x s` is constant. I'll try to fix the derivative

**Comment by jstoobysmith** [comment]
I believe the normalization will come up somewhere (but happy to be proven wrong). 
Namely because if `s = r * n` for `r` a scalar and `n` a normalized vector then: 
```lean
(inner x s - c * t) = r * (inner x n - (c /r) *t)
```
Then a plane wave `f(inner x s - c * t)` is really a function of `(inner x n - (c /r) *t)` and you have consequently 'lost' your notion of speed.

But it will either show up or it won't :).

zhikaip:
good point there, I think you're right I did not think this through properly :) how would I state that then? if I have `(hs : inner s s = 1)` lean complains about `unused variable hs`, or can I just ignore the warning?

jstoobysmith:
Once you change your proof to fix the derivatives, I think this warning will go away.

zhikaip:
one more question: is this `(h'' : ∀ x, HasFDerivAt (f₀' x) (f₀'' x) x)` the correct way of stating the second derivative?

jstoobysmith:
I think you probably need something like ```(h'' : ∀ x, HasFDerivAt (fun x => f₀' x 1) (f₀'' x) x)```

zhikaip:
I have more confidence that this should be correct and tried my best to simplify, let me know if you have any further suggestions

jstoobysmith:
This looks great, and seems correct to me. 
I think it might be worth making e.g. `hdt` and `hdt'` separate lemmas, to break things up a bit.

zhikaip:
The integration check is still complaining about "unusedArguments", what's the best way around this? is there a cleverer way to integrate `(hs : inner s s = (1:ℝ))` into the definition?

jstoobysmith:
For the definition you can prevent it by 

```lean
@[nolint unusedArguments]
def .... (_ : inner s s = (1:ℝ)) : ... := ...
```

jstoobysmith (approved):
Approved - feel free to merge when you are ready. Thanks :).

PR #550, opened 2025-05-11
feat : properties of EM wave

zhikaip:
TODO:
- [x] Prove the two theorems at the end by integrating with respect to t
 ~(need to learn integration, probably need to prove cross-product commutes with integration w.r.t. time)~
don't actually need integration
- [ ] E B s form an orthogonal triad
 (Straighforward from cross-product but not sure how to state)

Maybe in separate PR:
- [ ] Define MulAction SO(d) on Space d
- [ ]  wave equation is invariant under MulAction SO(3)

The ultimate goal is to show it is justified to assume propagation direction in z (`s = (EuclideanSpace.single 2 1)`) and the `ElectricField` lies in the xy-plane. I probably want at least the first two tasks completed before merging, but let me know if there are other suggestions :)


zhikaip:
I thought it might be a good idea if you could kindly comment on this first and maybe merge this part first before it gets too messy.

This PR proves 4 main results in PhysLean.Electromagnetism.Wave:
1. ∀ t, ∃ constE, ∀ x, inner ℝ (E t x - constE) s = 0
2. ∀ t, ∃ constB, ∀ x, inner ℝ (B t x - constB) s = 0
3. ∀ x, ∃ constE, ∀ t, (√(μ • ε)) • (E t x) = - s ×₃ (B t x) + constE
4 .∀ x, ∃ constB, ∀ t, (√(μ • ε))⁻¹ • (B t x) = s ×₃ (E t x) + constB 

which are direct consequences of Maxwell's equations. I think I can't do better for 3 and 4, the constant field can be space-varying, whereas showing the time-varying component must be transverse might be doable from combining 1 and 3 (correspondingly 2 and 4).

Feel free to give suggestions for cleaning up code and naming, I'm still getting used to this.

jstoobysmith:
I'll go through this tomorrow :). 

Some possible future things:

With e.g. 3 if you _also_ assume that E is a plane wave, I think you should relatively easily be able to show that constE must be a function of either both time and space or neither, and since it is independent of time it must be independent of space. Maybe we could eventually have this as an additional lemma.

Similarly in 1 I think one can show you can take `constE` to be a vector in the direction of `s` e.g. `c * s`, and then one gets something like `∀ t, ∃ c, ∀ x, inner ℝ (E t x) = c`, but since this also tells you that `c` must be a function of `x. s - c t` and since it is independent of `t` it must be independent of `x`.

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -127,10 +122,243 @@ theorem waveEquation_magneticField_of_freeMaxwellEquations
...
+      exact function_differentiableAt_fst (hf := hE.two_differentiable) ..
+    · exact differentiable_crossProduct_if_planewave h' hBwave
+
+/-- An electric planewave induces an magnetic field equal to `s ×₃ E` plus a constant field. -/
+lemma B_eq_s_cross_electricPlaneWave {E₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
+    {s : Space} {hs : inner ℝ s s = (1:ℝ)} {E : ElectricField} {B : MagneticField}
+    (hc : c = (√(μ • ε))⁻¹) (hEwave : E = electricPlaneWave E₀ c s hs) (h' : Differentiable ℝ E₀)
+    (hm : OM.FreeMaxwellEquations E B) (hB : ContDiff ℝ 2 ↿B) : ∀ x, ∃ constB, ∀ t,
+    (√(μ • ε))⁻¹ • (B t x) = (WithLp.equiv 2 (Fin 3 → ℝ)).symm
+    (WithLp.equiv _ _ s ×₃ WithLp.equiv _ _ (E t x)) + constB := by
```
jstoobysmith: Quick comment for now. Would be good if we could write this without `WithLp.equiv`, maybe need to define our own notation?

jstoobysmith: (I've seen [this](https://leanprover.zulipchat.com/#narrow/channel/113489-new-members/topic/Using.20crossProduct.20in.20EuclideanSpace.20.28Fin.203.29/with/520119542)). Maybe a temporary notation until things get sorted out on the mathlib side?

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -216,3 +209,95 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner ℝ s s = 1)
     rw [hs]
   rw [hsj]
   simp
+
+/-!
+## Helper lemmas for further derivation
+-/
+
+lemma wave_fderiv_inner_coord_sub {f₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
```
jstoobysmith: Could we put a doc-string for this lemma - it's hard to see what it is stating.

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -216,3 +209,95 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner ℝ s s = 1)
...
+  rw [← mul_one (s u), ← smul_eq_mul (s u), ContinuousLinearMap.map_smul]
+  rw [← mul_one (s v), ← smul_eq_mul (s v), ContinuousLinearMap.map_smul]
+  simp only [PiLp.smul_apply, smul_eq_mul, mul_one]
+  ring
+  repeat
+    intro x
+    apply DifferentiableAt.hasFDerivAt
+    fun_prop
+
+lemma differentiable_t_if_planewave {s : Space} {hs : inner ℝ s s = 1}
```
jstoobysmith: maybe `time_differentiable_of_eq_planewave`

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -216,3 +209,95 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner ℝ s s = 1)
...
+    {f₀ : ℝ → EuclideanSpace ℝ (Fin 3)} {f : Time → Space → EuclideanSpace ℝ (Fin 3)}
+    (h' : Differentiable ℝ f₀) (hf : f = planeWave f₀ c s hs) :
+    Differentiable ℝ fun t => f t x := by
+  rw [hf]
+  unfold planeWave
+  fun_prop
+
+open Matrix
+
+lemma differentiable_crossProduct_if_planewave {s : Space} {hs : inner ℝ s s = 1}
```
jstoobysmith: maybe `crossPoduct_differentiable_of_right_eq_planewave`

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -216,3 +209,95 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner ℝ s s = 1)
...
+  unfold planeWave
+  apply differentiable_pi''
+  intro i
+  fin_cases i <;>
+  · simp only [Nat.succ_eq_add_one, Nat.reduceAdd, Fin.isValue, PiLp.inner_apply,
+      RCLike.inner_apply, conj_trivial, LinearMap.mk₂_apply, WithLp.equiv_pi_apply,
+      Fin.reduceFinMk, WithLp.equiv_symm_pi_apply, Matrix.cons_val]
+    fun_prop
+
+lemma wave_deriv_inner_eq_inner_deriv_proj {f₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
```
jstoobysmith: would replace `deriv` here with `fderiv` - will make it easier to search for

PhysLean/Electromagnetism/Homogeneous.lean
```diff
@@ -71,7 +71,8 @@ def OpticalMedium.FreeMaxwellEquations (OM : OpticalMedium)
 
 theorem OpticalMedium.gaussLawElectric_of_free (E : ElectricField) (B : MagneticField)
     (h : OM.FreeMaxwellEquations E B) :
-    (∇ ⬝ E t) x = 0 := by
+    ∀ t x, (∇ ⬝ E t) x = 0 := by
```
jstoobysmith: Would be usual to put these `forall`'s as arguments e.g. 

```lean
theorem OpticalMedium.gaussLawElectric_of_free (E : ElectricField) (B : MagneticField)
    (h : OM.FreeMaxwellEquations E B) (t : Time) (x : Space) :
    (∇ ⬝ E t) x = 0 := by
```

is exactly the same

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -127,10 +122,243 @@ theorem waveEquation_magneticField_of_freeMaxwellEquations
...
+noncomputable def magneticPlaneWave (B₀ : ℝ → EuclideanSpace ℝ (Fin 3))
+    (c : ℝ) (s : Space) (hs : inner ℝ s s = (1:ℝ)) : MagneticField :=
+    planeWave B₀ c s hs
+
+/-- An electric plane wave minus a constant field is transverse for all x. -/
+lemma electricPlaneWave_transverse_forall_x {E₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
+    {s : Space} {hs : inner ℝ s s = (1:ℝ)} {E : ElectricField}
+    {B : MagneticField} (hEwave : E = electricPlaneWave E₀ c s hs)
+    (h' : Differentiable ℝ E₀) (hm : OM.FreeMaxwellEquations E B) :
+    ∀ t, ∃ constE, ∀ x, inner ℝ (E t x - constE) s = 0 := by
```
jstoobysmith: I think it would be nicer to make the goal 
`∃ (c : Time → EuclideanSpace ℝ (Fin 3)), ∀ x t, inner ℝ (E t x - c t) s = 0`

Then maybe `transverse_upto_time_fun_of_eq_electricPlaneWave`

Similar for the magnetic field case below.

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -127,10 +122,243 @@ theorem waveEquation_magneticField_of_freeMaxwellEquations
...
+  apply Differentiable.comp
+  fun_prop
+  exact fun x => wave_differentiable
+  fun_prop
+
+open Matrix
+
+/-- The time derivative of a magnetic planewave induces an electric field with
+time derivative equal to `-s ×₃ B'`. -/
+lemma dt_E_eq_s_cross_neg_dt_magneticPlaneWave {B₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
```
jstoobysmith: maybe `time_deriv_electricPlaneWave_eq_cross_time_deriv_magneticPlaneWave`. Similar for the below.

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -127,10 +122,243 @@ theorem waveEquation_magneticField_of_freeMaxwellEquations
...
+      ← OM.faradayLaw_of_free E B, hEwave, electricPlaneWave, ← hc]
+  unfold planeWave curl coord basis Space.deriv
+  ext i
+  fin_cases i <;>
+  · simp [-PiLp.inner_apply]
+    rw [wave_fderiv_inner_coord_sub h']
+  exact hm
+
+/-- A magnetic planewave induces an electric field equal to `-s ×₃ B` plus a constant field. -/
+lemma E_eq_s_cross_neg_magneticPlaneWave {B₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
```
jstoobysmith: I think we it would be good to change the goal here to 

```
∃ (c : Space → EuclideanSpace ℝ (Fin 3)), ∀ x t,  (√(μ • ε)) • (E t x) = - (WithLp.equiv 2 (Fin 3 → ℝ)).symm
    (WithLp.equiv _ _ s ×₃ WithLp.equiv _ _ (B t x)) + c x
```

Then maybe `electricPlaneWave_eq_cross_magneticPlaneWave_upto_space_fun`. Similar with the below.

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -127,10 +122,243 @@ theorem waveEquation_magneticField_of_freeMaxwellEquations
...
+    {s : Space} {hs : inner ℝ s s = (1:ℝ)} {E : ElectricField} {B : MagneticField}
+    (hc : c = (√(μ • ε))⁻¹) (hBwave : B = magneticPlaneWave B₀ c s hs) (h' : Differentiable ℝ B₀)
```
jstoobysmith: Just to check, what is the reason you have behind including e.g. `(hBwave : B = magneticPlaneWave B₀ c s hs)` as an assumption rather than using `magneticPlaneWave B₀ c s hs` directly in the goal? Is it because it is cleaner?

PhysLean/Electromagnetism/Wave.lean
zhikaip: do we want to wrap this around the mathlib `crossProduct` (e.g. the following) or just define a new one?

```
/-- Cross product in `EuclideanSpace ℝ (Fin 3)`. Uses `⨯` using `\X` or 
`\vectorproduct` or `\crossproduct`. -/
def crossProduct : EuclideanSpace ℝ (Fin 3) → EuclideanSpace ℝ (Fin 3) →
    EuclideanSpace ℝ (Fin 3) := fun a b => (WithLp.equiv 2 (Fin 3 → ℝ)).symm
    (WithLp.equiv _ _ a ×₃ WithLp.equiv _ _ b)

@[inherit_doc crossProduct]
infixl:74 " ⨯ " => crossProduct
```

PhysLean/Electromagnetism/Wave.lean
zhikaip: I think I need it in the form of `E t x` or `B t x` when rewriting using Maxwell's equations?

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
zhikaip:
```
/-- If `f₀` is a function of `(inner ℝ x s - c * t)`, the difference of `fderiv`
with respect to cross spatial coordinates is equal to the difference between
the corresponding cross components of the propagation direction `s` times the time derivative. -/
```

I feel like this isn't clear enough, but not sure how to phrase it better.

PhysLean/Electromagnetism/Wave.lean
jstoobysmith: I think wrapping it around mathlib's is better, and then writing a lemma `crossProduct_def` that can be used to rewrite the euclidean one in terms of the mathlib one so we  don't have to rewrite an api for it. I would also replace `def` with `abbrev` here.

Maybe if we could do it purely with notation, but I think that might be more difficult.

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
jstoobysmith: I think it might be better to replace this lemma with a lemma that relates the space derivative to the time derivative directly, rather than including the difference.

e.g.  

```
c * ((fun x' => (fderiv ℝ (fun x => inner ℝ (f₀ (inner ℝ x s - c * t))
    (EuclideanSpace.single u 1)) x') (EuclideanSpace.single v 1)) x) =  s u * ∂ₜ (fun t => f₀ (inner ℝ x s - c * t)) t v`
```

or something similar, from what this lemma should follow directly.

PhysLean/Electromagnetism/Wave.lean
jstoobysmith: You should be able to rewrite with Maxwell's equations even if you have `magneticPlaneWave B₀ c s hs` there instead of `B t x`. Lean will be able to work out that `magneticPlaneWave B₀ c s hs` is a specific type of `B t x`. Though I understand this would make things a lot more messy so happy if things are kept as you have them.

PhysLean/Electromagnetism/Wave.lean
jstoobysmith: Maybe something like:

```lean
set_option quotPrecheck false in
infixl:70 "⨯ₑ₃"  => fun a b => (WithLp.equiv 2 (Fin 3 → ℝ)).symm
    (WithLp.equiv 2 (Fin 3 → ℝ) a ×₃ WithLp.equiv 2 (Fin 3 → ℝ) b)
```

zhikaip: I've put this under `ClassicalMechanics.VectorFields` for now, let me know if you think it's better suited somewhere else.

zhikaip:
I've addressed the comments and made changes accordingly, hopefully haven't missed anything.

PhysLean/ClassicalMechanics/WaveEquation/Basic.lean
```diff
@@ -216,3 +207,91 @@ theorem planeWave_isWave (c : ℝ) (s : Space d) (hs : inner ℝ s s = 1)
...
   simp
+
+/-!
+## Helper lemmas for further derivation
+-/
+
+/-- If `f₀` is a function of `(inner ℝ x s - c * t)`, the fderiv of its components
+with respect to spatial coordinates is equal to the corresponding component of
+the propagation direction `s` times time derivative. -/
+lemma wave_fderiv_inner_coord_sub {f₀ : ℝ → EuclideanSpace ℝ (Fin 3)}
```
jstoobysmith: Can we change the name of this result to remove the `sub`, maybe

`space_fderiv_of_inner_product_wave_eq_space_fderiv`

jstoobysmith:
Left one more comment, and the linter pulled up an error. After these are fixed I think this looks good to me, and I can approve it for merge.

PhysLean/Electromagnetism/Wave.lean
jstoobysmith: This seems like a fine place for now.

jstoobysmith (approved):
Approved - Many thanks for this! I'm excited to see where this all goes :).

PR #554, opened 2025-05-14
feat(StringTheory): Add SU(5)+U(1) potential properties

jstoobysmith:
Add properties of the potential for the SU(5)+U(1) F-theory potential.

The key reference for this material is: arXiv:0912.0853


jstoobysmith:
Concerning: 

- "I have ensured that no lemmas or definitions are duplications of other results." 

There are some duplicate lemmas in this PR, however they will be overwritten in future PRs.

PR #555, opened 2025-05-14
split #501 

or4nge19:

PR #501, opened 2025-04-18
Create PseudoRiemannian

or4nge19:
This PR is divided in 2 parts. Next is Riemannian Metric (with implementation)

PR #564, opened 2025-05-22
Pseudo-Riemannian Metric - continue #501, #555

or4nge19:


jstoobysmith:
@or4nge19 Hi Matteo, would I be ok to close this PR and #501?

PR #571, opened 2025-05-25
chore(Mathematics/Fin/Involutions): golf a bit

pitmonticone:


jstoobysmith (approved):
Many thanks! Approved :)

PR #573, opened 2025-05-26
Use `ext` to golf proofs

eric-wieser:
See [Chaining Extensionality Lemmas in Lean’s Mathlib](https://link.springer.com/chapter/10.1007/978-3-031-66997-2_4) for a bit more insight into what `ext` is doing under the hood here.

You could consider dropping these lemmas entirely now and using `ext; dsimp` in their place, but that's a style decision I'll leave up to you.

### DIFF
--- a/PhysLean/Mathematics/PiTensorProduct.lean
+++ b/PhysLean/Mathematics/PiTensorProduct.lean
@@ -32,25 +32,13 @@ open TensorProduct
 ## induction principals for pi tensor products
 
 -/
+attribute [local ext] TensorProduct.ext
+
 lemma induction_tmul {f g : ((⨂[R] i : ι1, s1 i) ⊗[R] ⨂[R] i : ι2, s2 i) →ₗ[R] M}
     (h : ∀ p q, f (PiTensorProduct.tprod R p ⊗ₜ[R] PiTensorProduct.tprod R q)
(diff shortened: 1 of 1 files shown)


eric-wieser:
> This checklist is experimental and can be ignored if you wish.

Unfortunately I am forced to ignore it, only users with triage rights can check the boxes

jstoobysmith (approved):
Hi Eric, many thanks for this (and your other PR)! Ultimately I think it would be nice to replace these lemmas with `ext; dsimp`, I had a go replacing a few of them - but it didn't seem to work verbatim. But in the meantime I've approved this PR so feel free to merge when you (and the linters) are ready.

More generally the `Mathematics` folder of PhysLean contains mathematical results that aren't yet in/don't quite fit in mathlib but ultimately one day should be replaced with Mathlib results, so knowing these results can be simplified is very helpful :).

Don't worry about the checklist - but it is good to know you need triage rights to use it - so thanks for pointing this out. Though more generally if you find it easier to use branches over forks I can add you as an external collaborator on the project - let me know.

eric-wieser:
> But in the meantime I've approved this PR so feel free to merge when you (and the linters) are ready.

Thanks! Though it needs to be you that clicks the merge button, it isn't available to me.

jstoobysmith:
Opps, thanks for the heads up! I've merged both your PRs and sent you an invitation to join as an `outside collaborator' , which will hopefully give you more permissions to do things.

PR #574, opened 2025-05-26
feat(Relativity/SpaceTime/CliffordAlgebra): half of the isomorphism

eric-wieser:
The reverse direction of this isomorphism is an `Algebra.adjoin` problem so I'll leave it for someone else; but `CliffordAlgebra` is a place I can help with!


jstoobysmith (approved):
Many thanks - I've approved so feel free to merge :).

PR #577, opened 2025-05-27
chore: add some simp lemmas for the Pauli matrices

eric-wieser:
I'd advocate dropping `σ0` entirely and using `1`, but that could happen in another PR.

### DIFF
--- a/PhysLean/Relativity/PauliMatrices/Matrix.lean
+++ b/PhysLean/Relativity/PauliMatrices/Matrix.lean
@@ -57,6 +57,20 @@ lemma σ3_selfAdjoint : σ3ᴴ = σ3 := by
   rw [eta_fin_two σ3ᴴ]
   simp [σ3]
 
+/-! ### Products
+
+These lemmas try to put the terms in numerical order.
+We skip `σ0` since it's just `1` anyway.
+-/
+
+@[simp] theorem σ1_mul_σ1 : σ1 * σ1 = 1 := by simp [σ1, one_fin_two]
+@[simp] theorem σ2_mul_σ2 : σ2 * σ2 = 1 := by simp [σ2, one_fin_two]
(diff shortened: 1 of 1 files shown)


PhysLean/Relativity/PauliMatrices/Matrix.lean
```diff
@@ -57,6 +57,20 @@ lemma σ3_selfAdjoint : σ3ᴴ = σ3 := by
   rw [eta_fin_two σ3ᴴ]
   simp [σ3]
 
+/-! ### Products
+
+These lemmas try to put the terms in numerical order.
+We skip `σ0` since it's just `1` anyway.
+-/
+
+@[simp] theorem σ1_mul_σ1 : σ1 * σ1 = 1 := by simp [σ1, one_fin_two]
```
jstoobysmith: Can we replace `theorem` here with `lemma`. I know in Mathlib this would be a `theorem`, but we are trying to reserve `theorem` here for key results and use `lemma` for everything else.

jstoobysmith:
c.f. replacing `σ0` with `1`. The motivation of using `σ0` is that the `σ`'s form a 4-vector, and this is typically how physicists would write it. Maybe we could get away with defining notation for it though instead of a definition, so one gets all of the benefits of `1`.

PhysLean/Relativity/PauliMatrices/Matrix.lean
eric-wieser: Done, this was just muscle memory.

eric-wieser:
> c.f. replacing `σ0` with `1`. The motivation of using `σ0` is that the `σ`'s form a 4-vector, and this is typically how physicists would write it. Maybe we could get away with defining notation for it though instead of a definition, so one gets all of the benefits of `1`.

I'll try an `abbrev` in a follow-up PR.

jstoobysmith (approved):
Approved, many thanks. Hopefully you should be able to merge this time.

PR #579, opened 2025-05-27
feat : E B s orthogonal for EMwave

zhikaip:
Waiting for 
#25193
#25205
in mathlib but otherwise I think this is what I wanted to show


PhysLean/Electromagnetism/Wave.lean
```diff
@@ -283,78 +288,286 @@ lemma electricPlaneWave_eq_cross_magneticPlaneWave_upto_space_fun
...
+  use Ecu 0
+  intro t x
+  rw [hcxE]
+  have hu : inner ℝ x s - c * (c⁻¹ * inner ℝ x s) = 0 := by
+    rw [← mul_assoc]
+    simp [hc_non_zero]
+  rw [← hu, ← hcuE (c⁻¹ * inner ℝ x s) x, hcxE]
+
+/-- `B - s ⨯ₑ₃ E` is constant for an EMwave. -/
+lemma magneticField_sub_s_cross_electricField_eq_const_of_EMwave
```
jstoobysmith: I would probably rename this:

`magneticField_sub_cross_electricField_eq_const_of_planeWave`

and similar with what follows. I've removed the `s`, as the choice in `s` is a choice in what you call a variable, also replaced `EMwave` with `planeWave` , as this is what these waves are called in e.g. `hEwave`.

PhysLean/Electromagnetism/Wave.lean
```diff
@@ -283,78 +288,286 @@ lemma electricPlaneWave_eq_cross_magneticPlaneWave_upto_space_fun
...
+  use Bcu 0
+  intro t x
+  rw [hcxB]
+  have hu : inner ℝ x s - c * (c⁻¹ * inner ℝ x s) = 0 := by
+    rw [← mul_assoc]
+    simp [hc_non_zero]
+  rw [← hu, ← hcuB (c⁻¹ * inner ℝ x s) x, hcxB]
+
+/-- Pending #25205. -/
+lemma Fin.forall_fin_three {p : Fin 3 → Prop} : (∀ i, p i) ↔ p 0 ∧ p 1 ∧ p 2 :=
```
jstoobysmith: If this is a specific lemma in `#25205`, one could replace it's name with 

```
_root_. Fin.forall_fin_three
```
then after we update to a mathlib containing #25205 this will fail to build - telling us that we should remove it :).

Same with other results in this PR from mathlib PRs. The `_root_.` resets any namespaces.

jstoobysmith (approved):
Approved: Many thanks :). This looks really great!

PR #582, opened 2025-05-28
chore(Relativity/PauliMatrices/SelfAdjoint): golf proofs

eric-wieser:


jstoobysmith (approved):
Approved - thanks :).

PR #598, opened 2025-06-06
feat : setting up harmonic wave

zhikaip:
I'm convinced that the complex representation is crucial for a modern treatment of optics, and it is used in the context of time-harmonic plane waves (which can represent any signal/disturbance via Fourier theory). I think I'll try to define some general properties here (e.g. time-average of real representation vs complex representation) under the `wave` folder before actually moving on to optics.

(should showing equivalent representation between definitions be a theorem or lemma? feels like it's a important "definition" that allows us to go between different representations, but I'll imagine this will come up a lot and much of physics is giving names to mathematical objects depending on context.)


jstoobysmith:
I personally think it should be a lemma  - bit of a matter of taste, and not to say it's not important etc.

You mean complex representations of electric and magnetic field etc? If so I'm happy for this, but think they would be better as different types (i.e. I don't replace the real electric and magnetic field, but connect the two with an API).

zhikaip:
Not the field itself, but just using `e^iwt` for `cos(wt)`. Maybe I'll ask in zulip to clarify what I want and discuss what's the best way to implement this.

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
```diff
@@ -0,0 +1,70 @@
...
+    fun t r => a r * Real.cos (ω k * t - g r)
+
+TODO "EGQUA" "Show that the wave equation is invariant under rotations and any direction `s`
+    can be rotated to `EuclideanSpace.single 2 1` if only one wave is concerened."
+
+set_option linter.unusedVariables false in
+/-- Transverse monochromatic time-harmonic plane wave where the direction of propagation
+  is taken to be `EuclideanSpace.single 2 1`. -/
+@[nolint unusedArguments]
+noncomputable def TransverseHarmonicPlaneWave {c : ℝ} {k : Space} {E₀x E₀y ω δx δy : ℝ}
```
jstoobysmith: Lower case for definitions which are not Prop's or Type's so `transverseHarmonicPlaneWave`.

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
```diff
@@ -0,0 +1,70 @@
...
+    {Ex Ey : Time → Space → ℝ} (hk : k = EuclideanSpace.single 2 (ω/c))
+    (hx : Ex = harmonicWave (fun _ => E₀x) (fun _ => ω) (fun r => inner ℝ k r - δx) k)
```
jstoobysmith: Instead of these equality hypothesis, I think it would be better to have e.g. 

`let Ex := harmonicWave (fun _ => E₀x) (fun _ => ω) (fun r => inner ℝ k r - δx) k`

after the `Time → Space → EuclideanSpace ℝ (Fin 3) :=`

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
```diff
@@ -0,0 +1,70 @@
...
+
+Note TODO `EGQUA` may require considerable effort to be made rigorous and may heavily depend on
+the status of Fourier theory in Mathlib.
+
+-/
+
+namespace ClassicalMechanics
+
+/-- General form of time-harmonic wave in terms of angular frequency `ω` and wave vector `k`. -/
+noncomputable def harmonicWave (a ω g: Space → ℝ) (k : Space) :
```
jstoobysmith: Should `k` really be a point in `Space` here? Maybe `ℝ` is better.

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
```diff
@@ -0,0 +1,70 @@
...
+noncomputable def TransverseHarmonicPlaneWave {c : ℝ} {k : Space} {E₀x E₀y ω δx δy : ℝ}
+    {Ex Ey : Time → Space → ℝ} (hk : k = EuclideanSpace.single 2 (ω/c))
+    (hx : Ex = harmonicWave (fun _ => E₀x) (fun _ => ω) (fun r => inner ℝ k r - δx) k)
+    (hy : Ey = harmonicWave (fun _ => E₀y) (fun _ => ω) (fun r => inner ℝ k r - δy) k) :
+    Time → Space → EuclideanSpace ℝ (Fin 3) :=
+    fun t x => Ex t x • EuclideanSpace.single 0 1 + Ey t x • EuclideanSpace.single 1 1
+
+/-- The transverse harmonic planewave representation is equivalent to the general planewave
+  expression with `c = ω/k`. -/
+theorem TransverseHarmonicPlaneWaveisPlaneWave {c : ℝ} {k : Space} {E₀x E₀y ω δx δy : ℝ}
```
jstoobysmith: something like `transverseHarmonicPlaneWave_eq_planeWave`?

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
zhikaip: `k` is really `‖k‖ • s` where `s` is the direction. The wave vector does see use very often, I take your point that `k` isn't really a point, should I change all the `s` in previous wave-related definitions to `EuclideanSpace ℝ (Fin 3)`?

jstoobysmith: Maybe we could define an abbreviation?

```
abrev WaveVector (d : Nat := 3) := EuclideanSpace ℝ (Fin d)
```

jstoobysmith: Then `ω` should be a function from `WaveVector → ℝ`

zhikaip: what about `s` then? it is defined as `Space` in the other definitions, but calling it `WaveVector` also seems weird, should it just be `EuclideanSpace ℝ (Fin 3)`?

Or maybe `direction` should be a `structure` with `direction.s : EuclideanSpace ℝ (Fin 3)` and `direction.norm := norm s = 1`?

jstoobysmith: Yeah, `Direction` as a structure seems like a good suggestion. Maybe this could go in the `Space` directory in it's own file with an API around it? Then have functions `Space.toDirection` and `WaveVector.toDirection` etc. I think this makes sense, since `s` is dimensionless, right?

jstoobysmith:
> Not the field itself, but just using `e^iwt` for `cos(wt)`. Maybe I'll ask in zulip to clarify what I want and discuss what's the best way to implement this.

Please also post on Zuilp, but this is related to these `Phasor`s you mentioned right? Seems like a nice implementation would then be to define in a new file `.../Phasor.lean` the type `Phasor` and API relating it to waves etc.

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
zhikaip: I defined `Direction` in `Space.Basic` and had a go at refactoring the wave code I previously did, please check if those make sense.

zhikaip:
I removed the complex representation in this file for now, and I'll ask about the implementation in zulip tomorrow.

PhysLean/ClassicalMechanics/WaveEquation/HarmonicWave.lean
```diff
@@ -0,0 +1,81 @@
...
+theorem _root_.WithLp.equiv_eq_zero_iff {p V} [AddCommGroup V] {v : WithLp p V} :
+    WithLp.equiv p V v = 0 ↔ v = 0 := Iff.rfl
+
+@[simp]
+theorem _root_.EuclideanSpace.single_eq_zero_iff {ι 𝕜} [RCLike 𝕜] [DecidableEq ι] {i : ι} {x : 𝕜} :
+    EuclideanSpace.single i x = 0 ↔ x = 0 := Pi.single_eq_zero_iff
+
+/-- The transverse harmonic planewave representation is equivalent to the general planewave
+  expression with `c = ω/‖k‖`. -/
+lemma transverseHarmonicPlaneWave_eq_PlaneWave {c : ℝ} {k : WaveVector} {E₀x E₀y ω δx δy : ℝ}
```
jstoobysmith: Just lower-case `p` for the last `PlaneWave` here, and then everything else (except what the linter pulls up) looks fine to me :). Many thanks for this, and for adding the Direction type.

(Also, you may have noticed that the review process has slightly changed - this is just to make it more in line with other projects).

jstoobysmith (approved):
Approved - thanks.

PR #600, opened 2025-06-08
feat: polarization ellipse

zhikaip:
Defining the polarization ellipse with the real presentation which gives geometric intuition (and the nomenclature), and the more general definition that will come later will be shown to coincide with the definitions here.

### DIFF
--- a/PhysLean/Optics/Polarization/Basic.lean
+++ b/PhysLean/Optics/Polarization/Basic.lean
@@ -1,0 +1,114 @@
+/-
+Copyright (c) 2025 Zhi Kai Pong. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Zhi Kai Pong
+-/
+import PhysLean.ClassicalMechanics.WaveEquation.HarmonicWave
+import PhysLean.Electromagnetism.Wave
+/-!
+
+# Polarization
+
+In this file the real representation is used to develop representations of polarizations
+such as the polarization ellipse, Stokes parameters and the Poincaré sphere for monochromatic
+time-harmonic electromagnetic plane waves.
+
+More general definitions that can be applied to a wider range of situations will be shown to
+be equivalent to the definitions in this file where appropriate.
+
+-/
+
+/-!
+
+## Monochromatic wave
+
+-/
+
+namespace Optics
+open ClassicalMechanics
+open Electromagnetism
+open Real
+
+/-- x-component of monochromatic time-harmonic wave. -/
+noncomputable def Ex (k : WaveVector) (E₀x ω δx : ℝ) :=
+    harmonicWave (fun _ _ => E₀x) (fun _ r => inner ℝ k r - δx) (fun _ => ω) k
+
+/-- y-component of monochromatic time-harmonic wave. -/
+noncomputable def Ey (k : WaveVector) (E₀y ω δy : ℝ) :=
+    harmonicWave (fun _ _ => E₀y) (fun _ r => inner ℝ k r - δy) (fun _ => ω) k
+
+set_option linter.unusedVariables false in
+/-- General form of monochromatic time-harmonic electromagnetic plane wave where
+  the direction of propagation is taken to be `EuclideanSpace.single 2 1`. -/
+@[nolint unusedArguments]
+noncomputable def harmonicElectromagneticPlaneWave (k : WaveVector) (E₀x E₀y ω δx δy : ℝ)
+    (hk : k = EuclideanSpace.single 2 (ω/c)) :
+    ElectricField :=
+    fun t r => Ex k E₀x ω δx t r • EuclideanSpace.single 0 1 +
+    Ey k E₀y ω δy t r • EuclideanSpace.single 1 1
+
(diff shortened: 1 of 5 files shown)


zhikaip:
I realized there is nothing stopping me from defining an `electricPlaneWave` that is not transverse. It will probably lead to contradictions somewhere, but when things starts getting complicated we might end up with proving a lot of results about unphysical situations before realizing that they're impossible in the first place. Is this something I should be worried about?

an example: for the particular instance `transverseHarmonicPlaneWave `, I just had `fun t r => fx t r • EuclideanSpace.single 0 1 + fy t r • EuclideanSpace.single 1 1` with an arbitrary `k` previously which is obviously not necessarily transverse, and I realized that i had to add the condition `(hk : k = EuclideanSpace.single 2 (ω/c))`. 

Essentially I think I'm defining physical notions then proving properties about them, but there should be something in the other direction that "an object corresponds to this physical notion only if it have those properties". If you think this warrants more discussion, I can also post in zulip.

jstoobysmith:
Yes, I agree here, and I think this approach is actually the nicer thing to do. Namely, define things with as minimal assumptions as possible initially, and then add conditions when needed, or when shown needed etc. 
It also makes things easier to work with, because you don't have to carry around assumptions which are not needed.

PhysLean/Optics/Polarization/Basic.lean
```diff
@@ -0,0 +1,114 @@
...
+
+/-- y-component of monochromatic time-harmonic wave. -/
+noncomputable def Ey (k : WaveVector) (E₀y ω δy : ℝ) :=
+    harmonicWave (fun _ _ => E₀y) (fun _ r => inner ℝ k r - δy) (fun _ => ω) k
+
+set_option linter.unusedVariables false in
+/-- General form of monochromatic time-harmonic electromagnetic plane wave where
+  the direction of propagation is taken to be `EuclideanSpace.single 2 1`. -/
+@[nolint unusedArguments]
+noncomputable def harmonicElectromagneticPlaneWave (k : WaveVector) (E₀x E₀y ω δx δy : ℝ)
```
jstoobysmith, line 46: In the doc-string of this definition, and maybe also the two above, could we put the meaning of `E₀x E₀y ω δx δy`. I think this would help with the readability of the code.

PhysLean/Optics/Polarization/Basic.lean
```diff
@@ -0,0 +1,114 @@
...
+
+-/
+
+namespace Optics
+open ClassicalMechanics
+open Electromagnetism
+open Real
+
+/-- x-component of monochromatic time-harmonic wave. -/
+noncomputable def Ex (k : WaveVector) (E₀x ω δx : ℝ) :=
```
jstoobysmith: I would be a proponent for slightly more descriptive names here. Not sure what exactly to suggest though. Maybe, `monochromX`?

PhysLean/Optics/Polarization/Basic.lean
```diff
@@ -0,0 +1,114 @@
...
+/-!
+
+## Polarization ellipse
+
+-/
+
+namespace Polarization
+
+/-- Alternative representation for `Ex` to simplify calculations. -/
+noncomputable def Ex (E₀x τ δx: ℝ) := E₀x * cos (τ + δx)
```
jstoobysmith: I would make this as a lemma rather than a definition, relating it to `Ex` above.

zhikaip: is the current form clear enough? otherwise I'm thinking maybe I can use local notation `Ex` and `Ey`

jstoobysmith: I think in `polarizationEllipse` you would want `monochromX` and `monochromY`, and then the first thing you do in your proof is use `rw [ <- eq_monochromX, <- eq_monochromY]`.

jstoobysmith: But yeah introducing local notation for e.g. `monochromX` and `monochromY` wouldn't be a bad thing.

zhikaip: please check if you're happy with this, I think this is the form in standard literature and is also clearer to me.

jstoobysmith: This now all looks good to me! Though I think the linter might complain about a lack of doc-string for notation.

jstoobysmith: BTW, in case you're not aware you can run the linters locally using 

```
lake exe lint_all 
```
and 
```
./scripts/lint-style.sh
```
which is usually a lot quicker then doing them on github, since they don't have to re-build everything locally.

zhikaip: I ran this already and I don't think notation requires doc-string, do you want me to add one?

jstoobysmith: No, not if the linter doesn't pull it up. I thought that notation did need doc-strings, maybe it is not the case `local notation...`.

jstoobysmith (approved):
Ok, I think you are right regarding the notation, so I've approved this so it can be merged once the linters have finished. Many thanks :).

PR #604, opened 2025-06-09
Todo #6YB2U: the force corresponding to the harmonic oscillator potential

LodeVermeulen:
This PR derives the force that is the consequence of the harmonic oscillator potential, alongside some useful API lemmas.

Suggestions on improvements are appreciated, especially on some of the `have` statements that derive really basic stuff.


PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
```diff
@@ -93,61 +96,71 @@ lemma inverse_ω_sq : (S.ω ^ 2)⁻¹ = S.m/S.k := by
...
   left
-  trans (deriv (Neg.neg ∘ x) t)^2
-  · rfl
-  rw [deriv_comp]
-  · simp
-  · fun_prop
-  · exact hx t
+  have hx : deriv (- xₜ) t = - deriv xₜ t := by
+    exact deriv.neg
+  rewrite [hx]
```
jstoobysmith: does `rw [deriv.neg]` not work here?

jstoobysmith: If not then `rw [show deriv (- xₜ) t = - deriv xₜ t from deriv.neg]`?

PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
```diff
@@ -93,61 +96,71 @@ lemma inverse_ω_sq : (S.ω ^ 2)⁻¹ = S.m/S.k := by
...
+lemma force_is_linear (x : Space 1) : force S x = - S.k • x := by
+  unfold force potentialEnergy
+  change -∇ ((1 / (2 : ℝ)) • S.k • (fun (x : Space 1) => ⟪x, x⟫_ℝ)) x = -S.k • x
+  rw [grad_smul, grad_smul]
+  · rw [grad_inner]
+    simp only [one_div, Pi.smul_apply, neg_smul, neg_inj, smul_smul, smul_algebra_smul_comm]
+    simp only [mul_smul]
+    have h4 : (2 : ℕ) • (2⁻¹ : ℝ) • S.k • x = (2 : ℝ) • (2⁻¹ : ℝ) • S.k • x := by
+      norm_cast
+    rw [h4]
```
jstoobysmith: you could do `rw [show (2 : ℕ) • (2⁻¹ : ℝ) • S.k • x = (2 : ℝ) • (2⁻¹ : ℝ) • S.k • x by norm_cast]` here instead of the `have`, but yeah I don't think there is a simpler  way.

PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
```diff
@@ -93,61 +96,71 @@ lemma inverse_ω_sq : (S.ω ^ 2)⁻¹ = S.m/S.k := by
...
-
-/- This variable should be removed once the above `semiformal_result` is implemented. -/
-variable (force : (S : HarmonicOscillator) → (x : Time → ℝ) → Time → ℝ)
+noncomputable def force (S : HarmonicOscillator) (x : Space 1)
+  : EuclideanSpace ℝ (Fin 1) := - ∇ (potentialEnergy S) x
 
 /-- The force on the classical harmonic oscillator is `- k x`. -/
-semiformal_result "6YB2U" force_is_linear (x : Time → ℝ) :
-  force S x = - S.k • x
+lemma force_is_linear (x : Space 1) : force S x = - S.k • x := by
```
jstoobysmith: Can we actually rename this to `force_eq_linear`? I know I named it `force_is_linear`, so this is my bad :).

jstoobysmith:
Concerning the linters, let me point out that they can be run locally using `lake exe lint_all` and `./scripts/lint-style.sh`. Running them locally is quicker than waiting for github to pull up the errors.
(you may already know this, but just in case, I like to point it out as I think it's useful to know. :)).

LodeVermeulen:
> Concerning the linters, let me point out that they can be run locally using `lake exe lint_all` and `./scripts/lint-style.sh`. Running them locally is quicker than waiting for github to pull up the errors. (you may already know this, but just in case, I like to point it out as I think it's useful to know. :)).

Thanks, I was actually looking for a way to run them locally but couldn't find it. I will do this shortly.

PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
LodeVermeulen: The latter works!

PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
zhikaip: shouldn't the `grad_inner` have `(2:ℝ)` in the definition which avoids this?

jstoobysmith: Yeah, I think that would be an option.

jstoobysmith: Would probably make things easier in the long run aswell

PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
```diff
@@ -94,67 +97,70 @@ lemma inverse_ω_sq : (S.ω ^ 2)⁻¹ = S.m/S.k := by
...
 @[sorryful]
-def EquationOfMotion (x : Time → ℝ) : Prop := sorry
+def EquationOfMotion (x : Time → Space 1) : Prop := sorry
 
 /-- The equations of motion are satisfied if and only if Newton's second law holds. -/
-@[sorryful]
-lemma equationOfMotion_iff_newtons_second_law (x : Time → ℝ) :
-    EquationOfMotion x ↔ ∀ t, force S x t = S.m * deriv (fun t' => deriv x t') t := by sorry
+semiformal_result "6YBEI" equationOfMotion_iff_newtons_second_law (x : Time → Space 1) :
+    EquationOfMotion x ↔ ∀ t, force S (x t) = S.m • deriv (fun t' => deriv x t') t
```
LodeVermeulen: I had to revert to this way of writing it down, as the current definition of `force` is not a function of time, so Lean would throw an error when attempting to write down its time derivative.

jstoobysmith: Could you make it the following?

```lean4
@[sorryful]
lemma equationOfMotion_iff_newtons_second_law (x : Time → Space 1) :
    EquationOfMotion x ↔ ∀ t, force S (x t) = S.m • deriv (fun t' => deriv x t') t
```

The aim is to get rid of eventually get rid of the `semiformal_result`s.

LodeVermeulen: Ah that would be the better way to do this, yes. Changed it.

LodeVermeulen:
This took me a lot of time and effort but everything should be sorry free. I highly suspect a lot of the results can be golfed still, but I will mark this PR as ready for review now.

jstoobysmith (approved):
Great! This looks really great to me! Thank you for putting in the effort to do this! I don't see any needed changes so have approved this PR - feel free to merge (let me know if you can't). 

Once again thanks for this!

PR #612, opened 2025-06-16
feat: define variational derivative

lecopivo:
This PR defines variational gradient which allows us to state many physical laws in the form `δS = 0` and then show it is equivalent to the corresponding differential equation.

In particular there is 
```
inductive HasVarGradientAt (S' : (X → U) → (X → ℝ)) (grad : X → U) (u : X → U)  (μ : Measure X) : Prop := ...
```
which is saying that `grad` is variational gradient of `∫ x, S' u x ∂μ`. 

One potentially surprising thing with `HasVarGradientAt` is that you have to provide the unintegrated action `S'` rather than the full action `S u = ∫ x, S' u x ∂μ`. 

---

For example the statement that the variation of 1d pacticle's action gives Newton's law i.e. for
```
S(x) = ∫ 1/2*m*‖ẋ(t)‖² - V(x(t)) dt
```
we have
```
δS/δx(t) = - m*ẍ(t) - V'(x(t))`
```
is stated as
```
example (m : ℝ) (u V : ℝ → ℝ) (hu : ContDiff ℝ ∞ u) (hV : ContDiff ℝ ∞ V) : 
    HasVarGradientAt
      (fun (u : ℝ → ℝ) (t : ℝ) => 1/2 * m * deriv u t ^ 2 - V (u t))
      (fun t => - m * deriv (deriv u) t - deriv V (u t))
      u := ...
```

### DIFF
--- a/PhysLean/Mathematics/VariationalCalculus/HasVarAdjoint.lean
+++ b/PhysLean/Mathematics/VariationalCalculus/HasVarAdjoint.lean
@@ -6,6 +6,30 @@ Authors: Tomas Skrivan, Joseph Tooby-Smith
 
 import PhysLean.Mathematics.VariationalCalculus.IsTestFunction
 
+/-! Variational adjoint
+
+Definition of adjoint of linear function between function spaces. It is inspired by the definition
+of distributional adjoint of linear maps between test functions as described here:
+https://en.wikipedia.org/wiki/Distribution_(mathematics) under 'Preliminaries: Transpose of a linear
+operator' but we require that the adjoint is function between test functions too.
+
+The key results are:
+  - variational adjoint of identity is identity, `HasVarAdjoint.id`
+  - variational adjoint of composition is composition of adjoint in reverse order,
+    `HasVarAdjoint.comp`
+  - variational adjoint of deriv is `- deriv`, `HasVarAdjoint.deriv`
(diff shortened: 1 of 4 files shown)


PhysLean/Mathematics/VariationalCalculus/HasVarAdjoint.lean
```diff
@@ -0,0 +1,258 @@
...
+namespace HasVarAdjoint
+
+variable {μ : Measure X}
+
+theorem id : HasVarAdjoint (fun φ : X → U => φ) (fun φ => φ) μ where
+  test_fun_preserving  _ hφ := hφ
+  test_fun_preserving' _ hφ := hφ
+  adjoint _ _ _ _ := rfl
+
+theorem comp {F : (X → V) → (X → W)} {G : (X → U) → (X → V)} {F' G'}
```
jstoobysmith: Just a quick comment: we are following the convention to use `lemma` for everything but the most important results (this is unlike most other Lean projects).

lecopivo: Ok, I replaced all `theorem` with `lemma`

jstoobysmith (approved):
I fixed the linter errors, and have 'approved' this PR. 
Feel free to merge if/when you are ready (if it lets you, otherwise let me know). Thanks :).

PR #613, opened 2025-06-19
feat(Strings): Update charges value

jstoobysmith:
Update the allowed values charges can take from integers to a general type which is an `AddCommGroup`. This allows the charges directory to also be used with multiple U(1)'s or finite groups.


jstoobysmith:
Replaced with #647

PR #617, opened 2025-06-20
feat: extensionality property of variational adjoint

lecopivo:
Adding extensionality property to variational adjoint saying that for `F φ x = F φ' x` we require `φ x = φ' x` to be equal on some larger set. This guarantees that `F φ x` can depend only on values `φ y` for `y` not too far away from `x`.

This property allows us to then show that `F f` has some meaningful value for `f` smooth functions.

### DIFF
--- a/PhysLean/Mathematics/VariationalCalculus/Basic.lean
+++ b/PhysLean/Mathematics/VariationalCalculus/Basic.lean
@@ -48,7 +48,7 @@ lemma fundamental_theorem_of_variational_calculus {f : X → V}
 weakened to `Continuous f`.
 
 The proof is by contradiction, assume that there is `x₀` such that `f x₀` then you can easily
- construct `g` test function with support on the neighborhood of `x₀` such that `⟪f x, g x⟫ ≥ 0`. -/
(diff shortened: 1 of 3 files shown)


jstoobysmith:
Last sorry can be filled in with:

```lean4
have h : ∀ x ∈ K, φ =ᶠ[nhds x] φ' := by 
        intro x hx 
        apply Filter.eventuallyEq_of_mem (s := Metric.thickening 1 K)
        refine mem_interior_iff_mem_nhds.mp ?_
        rw [@mem_interior]
        use Metric.thickening 1 K 
        simp 
        apply And.intro 
        · exact Metric.isOpen_thickening
        · rw [@Metric.mem_thickening_iff_exists_edist_lt]
          use x 
          simpa using hx 
        · intro x hx 
          have hx' : x ∈ Metric.cthickening 1 K := Metric.thickening_subset_cthickening 1 K hx
          exact hφ x hx'
```

lecopivo:
Excellent! I also generalized `HasVarAdjoint.unique` to smooth functions by adapting your proof of `HasVarGradientAt.unique`. 

Thus `F' (fun _ => 1)` is unique and has a proper meaning!

jstoobysmith (approved):

PR #618, opened 2025-06-20
feat: variational adjoint for `adjoint`, `fderiv`, `gradient` and `deriv`

lecopivo:
Variational adjoint of `adjoint`, `fderiv`, `gradient`, `deriv`, `prod.mk`, `prod.fst`, `prod.snd` and corresponding theorems for `HasVarAdjDerivAt`


lecopivo:
I was just playing around with refactoring API around `IsTestFunction` as some proofs are unnecessarily complicated.

The theorem
```
lemma IsTestFunction.comp_left {f : X → V} (hf : IsTestFunction f)
    {g : V → U} (hg1 : g 0 = 0) (hg : ContDiff ℝ ∞ g) :
    IsTestFunction (fun x => g (f x))
```
Is exactly the right composition theorem :)

jstoobysmith:
I filled in the sorry to: `divergence_prodMk` but added the instances `[FiniteDimensional 𝕜 E] [FiniteDimensional 𝕜 F]`, since otherwise you can't use `divergence_eq_sum_fderiv'`

lecopivo:
> I filled in the sorry to: `divergence_prodMk` but added the instances `[FiniteDimensional 𝕜 E] [FiniteDimensional 𝕜 F]`, since otherwise you can't use `divergence_eq_sum_fderiv'`

Thanks! That assumption is totally fine and it will be the case in every application.

PhysLean/Mathematics/VariationalCalculus/HasVarAdjDeriv.lean
```diff
@@ -143,49 +143,74 @@ lemma unique {X : Type*} [NormedAddCommGroup X] [InnerProductSpace ℝ X]
...
 protected lemma deriv (F : (ℝ → U) → (ℝ → ℝ)) (F') (u) (hF : HasVarAdjDerivAt F F' u) :
     HasVarAdjDerivAt (fun φ : ℝ → U => deriv (F φ))
     (fun ψ x => F' (fun x' => - deriv ψ x') x) u :=
   comp (F:=deriv) (G:=F) (hF := deriv' (F u) hF.apply_smooth_self) (hG := hF)
 
-lemma fmap [CompleteSpace U] [CompleteSpace V]
+lemma fmap
+    {U} [NormedAddCommGroup U] [InnerProductSpace ℝ U]
+    {V} [NormedAddCommGroup V] [InnerProductSpace ℝ V]
+    [CompleteSpace U] [CompleteSpace V]
```
lecopivo, line 447: I think we really want this to be formulated for `InnerProductSpace'`

PhysLean/Mathematics/VariationalCalculus/IsLocalizedfunctionTransform.lean
```diff
@@ -0,0 +1,317 @@
...
+lemma id : IsLocalizedFunctionTransform (id : (Y → V) → (Y → V)) := by
+  intro K cK
+  use K
+  constructor
+  · exact cK
+  · intro φ φ' hφ
+    intro x hx
+    simp
+    rw [hφ x hx]
+
```
lecopivo, line 75: At this point you could formulate 
``` 
lemma fmap (f : Y -> V -> W) (F : (X -> U) -> (Y -> V)) (hF : IsLocalizedFunctionTransform F): IsLocalizedFunctionTransform (fun g y => f y (F g y)) := ...
```
Then proof of `add` would be something like
```
apply fmap (f:= fun _ y => y.1 + y.2) (F:=fun g x => (F g x, G g x))
apply prod hF hG
```

lecopivo: But you can leave it as it is

lecopivo:
Excelent work! Refactor exactly as I had in mind and great that `HasVarAdjoint.adjFDeriv_apply` is done!

PhysLean/Mathematics/VariationalCalculus/HasVarAdjoint.lean
```diff
@@ -444,7 +381,483 @@ lemma smul_right {F : (X → U) → (X → V)} {ψ : X → ℝ} {F' : (X → V)
...
+            · rfl
+            rw [fderiv_comp]
+            simp
+            simp [f',f ]
+            · exact ContinuousLinearMap.differentiableAt _
+            · exact hψ.differentiable y
+            · exact real_inner_comm' (φ y) dy
+  ext := IsLocalizedFunctionTransform.adjFDeriv
+
+protected lemma gradient {d} :
```
lecopivo, line 658: Because `gradient f x = adjFDeriv R f 1` this theorem should be direct consequence of `adjFDeriv_apply` and it is pointless to have this long proof.

You will have to rewrite `divergence` to `Space.div`, but it is a question what should appear in the statement. I would personally deprecate `Space.div`

jstoobysmith: Agreed that we should deprecate`Space.div`, and also `Space.grad` at some point. 
I've simplified this proof.

PhysLean/Mathematics/VariationalCalculus/HasVarAdjDeriv.lean
jstoobysmith: Fixed.

jstoobysmith:
Ok, so I think that is all the sorries filled in except `divergence_smul`, where I have put `sorryful` attribute. Nothing depends on this lemma anyway. I think it is then a case of tidying and then at somepoint putting your proof of the Euler-lagrange equations.

jstoobysmith:
I believe I've now fixed all the linters etc. I also added your proof of the Euler-lagrange equation for a particle (needed to modify it a bit). 

I suspect there are lots of proofs that could be golfed and/or generalized here, but I think it might be best to merge this PR now, and then we can make these changes in future ones. E.g. I haven't made the change with the `Norm₂` yet.

jstoobysmith (approved):

PR #620, opened 2025-06-20
feat: Space.grad equal gradient

jstoobysmith:
Proving the result:

```lean
lemma grad_eq_gradiant {d} (f : Space d → ℝ) :
    ∇ f = gradient f := by ...
```

The latter has general API around it which may be useful.


zhikaip:
do I need to do anything? I did look at `gradient` at the time, but I'm not that famliar with the `toDual` definitions so I just stick to defining from `deriv`, I'll trust that they're equivalent as you've shown here :)

jstoobysmith:
Hey @zhikaip, nothing in particular. I think it is just it is good practice to have two sets of eyes on PRs in general and where possible. In case e.g. anything obvious is missed etc. So thanks :).

PR #622, opened 2025-06-22
feat: define `adjFDeriv` and provide basic API

lecopivo:
Define `adjFDeriv 𝕜 f x = (fderiv 𝕜 f x).adjoint`. The main purpose of this definition is to compute `gradient f x = adjFDeriv 𝕜 f x 1` because we can formulate composition theorem for `adjFDeriv_comp` but we can't do that for `gradient`.

`adjFDeriv` will naturally appear in [`HasVarAdjDeriv.fmap`](https://github.com/lecopivo/PhysLean/blob/8ab64cb1fa4b486f4361bd73382b5e0677c4c082/PhysLean/Mathematics/VariationalCalculus/HasVarAdjDeriv.lean#L164) and having API for it will allow up to easily derive Euler-Lagrange equations

Ideally this is PR is to mathlib but the situation around `WithLp` is a bit messy and I really do not feel like dealing with right now.


lecopivo:
Yay, thanks. Last sorry to go!

jstoobysmith:
three more I think, there are two in HasVarAdjDeriv :).

jstoobysmith (approved):
I've 'approved' this, and will merge. Many thanks for all of this!

PR #644, opened 2025-07-12
feat(Mathematics): define distributions and dirac delta

kckennylau:


PhysLean/Mathematics/Distribution/Basic.lean
```diff
@@ -0,0 +1,79 @@
...
+  at `0` and then take the inner product with `v`, i.e. `⟨v, η 0⟩`.
+
+-/
+
+open SchwartzMap NNReal
+noncomputable section
+
+/-- A distribution on `E` (normed vector space over `𝕜`) is a continuous linear map
+`𝓢(ℝ, E) →L[𝕜] 𝕜` where `𝒮(ℝ, E)` is the Schwarz space of smooth functions `ℝ → E` with rapidly
+decreasing iterated derivatives. This is notated as `ℝ →d[𝕜] E`. -/
```
jstoobysmith: The phrase 

```quote
the Schwarz space of smooth functions `ℝ → E` with rapidly
decreasing iterated derivatives.
```

would trip me up a bit I think, due to the `Schwarz space`. I think it something like the following may be better (if it is correct)

```quote
the space of smooth functions `ℝ → E` with rapidly
decreasing iterated derivatives (aka the Schwarz space).
```

PhysLean/Mathematics/Distribution/Basic.lean
```diff
@@ -0,0 +1,79 @@
...
+    refine ⟨s, C, fun η ↦ ?_⟩
+    obtain ⟨k, n, x, hkn, hη⟩ := hf η
+    have hs : s.Nonempty := ⟨(k, n), hkn⟩
+    refine hη.trans <| mul_le_mul_of_nonneg_left ((le_seminorm' 𝕜 k n η x).trans ?_) C.2
+    rw [Seminorm.finset_sup_apply]
+    refine (NNReal.coe_le_coe (r₁ := ⟨SchwartzMap.seminorm 𝕜 k n η, apply_nonneg _ _⟩)).2 ?_
+    convert s.le_sup hkn
+      (f := fun kn : ℕ × ℕ ↦ (⟨SchwartzMap.seminorm 𝕜 kn.1 kn.2 η, apply_nonneg _ _⟩ : ℝ≥0))
+
+/-- Dirac delta in a given direction (represented by `E →L[𝕜] 𝕜`). -/
```
jstoobysmith: In the module doc string you say:

```quote
(the direction `v` corresponds to the inner product `⟨v, -⟩`)
```

I think it would be nice to add this here too.

PhysLean/Mathematics/Distribution/Basic.lean
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Distribution
+
+/-- We construct a distribution from the following data:
+1. We take a linear map `f` that evaluates the given Schwartz function `η`. At this stage we don't
+   need `f` to be continuous.
+2. Recall that a Schwartz function `η` satisfies that
+   `{ |x| ^ k * ‖iteratedDeriv n η x‖ ∣ x : ℝ }` is bounded for any pair `(k, n) ∈ ℕ × ℕ`.
+3. We take a finite set `s` of pairs `(k, n) ∈ ℕ × ℕ` such that for each test function `η`, the
+   norm `‖f η‖` is controlled by `|x| ^ k * ‖(d^n/dx^n) η‖` for some `x : ℝ` and for some
```
jstoobysmith: `is controlled by` is confusing to me.

PhysLean/Mathematics/Distribution/Basic.lean
kckennylau: changed

PhysLean/Mathematics/Distribution/Basic.lean
kckennylau: changed, and added `diracDelta` (without `'`) that actually takes direction

PhysLean/Mathematics/Distribution/Basic.lean
kckennylau: changed to bounded

jstoobysmith (dismissed):
Approved. Let me know if you are happy for me to merge.

kckennylau:
I've changed diracDelta to use the delta in the library now. I'm happy for this to be merged.

jstoobysmith (dismissed):
Approved new changes and added you as an external collaborator to the project, so you should be able to merge. (Feel free to use the 'branches' approach rather than the 'forks' approach to github aswell if that is easier for you).

PhysLean/Mathematics/Distribution/Basic.lean
```diff
@@ -0,0 +1,149 @@
...
+- Generalise `derivative` to higher dimensional domain.
+
+-/
+
+open SchwartzMap NNReal
+noncomputable section
+
+/-- A distribution on `E` (normed vector space over `𝕜`) is a continuous linear map
+`𝓢(ℝ, E) →L[𝕜] 𝕜` where `𝒮(ℝ, E)` is the Schwarz space of smooth functions `ℝ → E` with rapidly
+decreasing iterated derivatives. This is notated as `ℝ →d[𝕜] E`. -/
```
jstoobysmith: I think this doc-string needs updating. I think it is also worth mentioning in this doc-string that this is a 'generalization' (using the word lightly) of functions from `E` (which is usually e.g. space or spacetime) to `F`.

PhysLean/Mathematics/Distribution/Basic.lean
```diff
@@ -0,0 +1,149 @@
...
+
+
+section RCLike
+
+/-- Definition of derivative of distribution: Let `u` be a distribution. Then its derivative is
+`u'` where given a test function `η`, `u' η := -u(η')`. This agrees with the distribution generated
+by the derivative of a differentiable function (with suitable conditions) (to be defined later),
+because of integral by parts (where the boundary conditions are `0` by the test functions being
+rapidly decreasing). -/
+def derivative : (ℝ →d[𝕜] 𝕜) →ₗ[𝕜] (ℝ →d[𝕜] 𝕜) where
```
jstoobysmith, line 120: Not saying for this PR, but I think we should think about how this can be generalized to 

- `(E →d[𝕜] F) →ₗ[𝕜] (E →d[𝕜] 𝕜)` giving the divergence 
- `(ℝ →d[𝕜] 𝕜) →ₗ[𝕜] (E →d[𝕜] 𝕜)` giving the gradient 
- Probably also something for the derivative in a given direction etc.

jstoobysmith (approved):

PR #650, opened 2025-07-16
feat(CM/Harmonic Oscillator): Fill in definition of force 

fenghourun:
~The type signature is a little wonky since `x = x(t)` 🤔~

Move over refactor from https://github.com/HEPLean/PhysLean/pull/604/files to use `Space 1`

### DIFF
--- a/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
+++ b/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
@@ -7,6 +7,7 @@ import PhysLean.Meta.TODO.Basic
 import PhysLean.Meta.Informal.SemiFormal
 import PhysLean.Meta.Linters.Sorry
 import PhysLean.SpaceAndTime.Time.Basic
+import PhysLean.SpaceAndTime.Space.Basic
 import Mathlib.Analysis.Calculus.Deriv.Add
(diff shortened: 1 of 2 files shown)


jstoobysmith:
Yeah, I think `potentialEnergy` needs to be changed not depend on a trajectory but to depend on just the position coordinate. I personally still think it would be best to work starting from #604 here, I don't think that PR builds mainly due to the replacement of `ℝ` with `Space 1`, which would be needed to be done at some point.

fenghourun:
@jstoobysmith Fair enough! I'll try to move over the  usage of `Space 1` from #604 👍

LodeVermeulen:
Is this not pretty much a duplicate of my PR https://github.com/HEPLean/PhysLean/pull/604? I am going to try to get that PR over the line this weekend, if all goes well.

jstoobysmith:
Maybe we could coordinate on the Zulip, there is still a bit to be done with the harmonic oscillator, e.g. the equation of motion stuff (which is now possible given the Euler-Lagrange equations), damping etc. But things will probably become complicated if two people are working on this at the same time (unless you work together). Alternatively, David, I'm happy to suggest other similar things to formalize.

fenghourun:
@LodeVermeulen @jstoobysmith let's move the discussion over to zulip! I'd be happy to tackle something else as well

PR #657, opened 2025-07-24
feat: harmonic oscillator initial time

jstoobysmith:
Update the harmonic oscillator initial conditions so that they contain an initial time `t₀` rather than assuming that this time is equal to zero.

### DIFF
--- a/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
+++ b/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
@@ -3,14 +3,9 @@ Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Joseph Tooby-Smith, Lode Vermeulen
 -/
-import PhysLean.Meta.TODO.Basic
 import PhysLean.Meta.Informal.SemiFormal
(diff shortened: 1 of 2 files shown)


zhikaip:
The changes look fine to me, but together with `TODO "IDXQA" "Modify the harmonic oscillator so it is centered on an arbitrary point
  `y` rather than on the origin."` I'm somewhat concerned whether generalizing everything will carry a lot of variables around and make the formulas intractable (e.g. every `x` and `t` will become `x - x₀` and `t - t₀`).

I'm just imagining possible solutions, maybe we want to have the most general case (i.e. arbitrary starting position and time etc.) and a simplified version (centered at origin, t = 0 etc.) that base upon this general case which makes the code more readable and beginner-friendly. This may be duplicate work so I'm not sure what's the best.

jstoobysmith:
Ok, thanks @zhikaip, I think I agree with this, and in fact given what you have said I don't think it is worth merging this PR. 

I'm trying to internalize what the `Time` type actually physically corresponds to. And I think the best way to interpret it is as: 
- Time in a given set of units, 
- Where the origin has been chosen. 

The choice of units and the choice of origin can depend on the problem (as long as things are consistent), and here, we are free to choose the `origin` to be the initial time of the oscillator. The units don't matter as long as you are consistent when choosing the units of `t : Time` and of the angular frequency `ω`.

PR #661, opened 2025-07-27
Create Matrix.det.exp

or4nge19:
This file proves the formula `det (exp A) = exp (tr A)` for matrices, also known as Lie's trace formula. Its immediate need is for several results related to Lorentz Algebra and Lorentz Group.

It needs some small clean-up (removing omits) and moving the relevant API to SchurTriangulation.  

edit: building on this, in ExponentialMap.lean we prove that the exponential of an element of the Lorentz algebra is a member of the restricted Lorentz group.


or4nge19:
I've cleaned-up the files as much as I could, there are a few lemmas that may be put in a better place (a generic Matrix folder? Matrix folders mirroring Mathlib organization?) But since Matrix.det.exp would go to mathlib we may keep things as they are until ported. 
I've removed a poisonous DecidableEq instance in SchurTriangulation that was creating diamonds with LinearOrder once utilised ExponentialMap (LinearOrder has its own Decidable instance).

PhysLean/Mathematics/DataStructures/Matrix/LieTrace.lean
```diff
@@ -0,0 +1,391 @@
...
+- `Matrix.det_exp`: The determinant of the exponential of a matrix is the exponential of its trace.
+-/
+
+/-!
+We give a higher priority to the canonical `1` and `*` instances coming
+from `Mathlib` so that every occurrence of `1 : Matrix _ _ _` and
+`A * B` uses the very same definitions.
+-/
+attribute [instance 100] Matrix.instHMulOfFintypeOfMulOfAddCommMonoid
+attribute [instance 100] Matrix.one
```
jstoobysmith: I think my only comment is, can we make these two attributes local?

jstoobysmith:
The linter error you are getting is that you have changed the location of the SchurTriangulation file, but not some of the places it was imported.

or4nge19:
haven't been able to fix toLorentzGroup_det_one yet, after the removal of DecidableEq from SchurTriangulation. Likely the issue is stupid, I'll try again tomorrow. I'd be grateful if you could have a quick look as you are more familiar with the SL2C API. I'll check again tomorrow, worse we keep the instance and will try to make it work downstream

jstoobysmith:
Ok, no worries, I will have a look tomorrow morning.

jstoobysmith (approved):
Nice! Good job fixing this, and many thanks for this PR. I've approved it, so feel free to merge when you are ready :).

PR #672, opened 2025-07-30
feat: More results related to units

jstoobysmith:
See: 

https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/physical.20units

For a conversation regarding the implementation here.


jstoobysmith:
There are certain improvements which can no-doubt be made to this formalization, these will be made at a latter date.

PR #673, opened 2025-07-31
Refactor Space API

fenghourun:


fenghourun:
@jstoobysmith What do you suggest as next steps to get a minimal PR working and compiling? Shall I go ahead and define a linearIsometry between `Space d` and `EuclideanSpace`, then fix the compile errors in the rest of the file?

jstoobysmith:
So I would next put an instance of `InnerProductSpace` on `Space d`, and then yes, define this linear isometry. From there, it is a bit unclear to me how much work would be needed to fix everything else. But I hope not much, but if so we can discuss options to make a working PR.

fenghourun:
> So I would next put an instance of `InnerProductSpace` on `Space d`, and then yes, define this linear isometry. From there, it is a bit unclear to me how much work would be needed to fix everything else. But I hope not much, but if so we can discuss options to make a working PR.

Sounds good, I've went ahead and begun working in that direction while using `Time/Basic.lean` as a reference. Taking a pause here, could you have a look to make sure the code style is in line with what's expected within the project? (Still in the process of learning LEAN so bear with me 😆 )

jstoobysmith:
Hey, absolutely no worries. I'm more than happy to answer questions or help guide you through things. 
What is currently in this PR looks really good and does match the coding style I would expect, so well done :).

Also I have an idea, which will mean you will be able to PR this sooner without having to fix all the annoying builds, and means we can work on this in smaller steps. That is: Instead of overwriting the definition `Space` we (in a file `Space.Struct`) could define `SpaceStruct` with your definition, and slowly build the API up, so that we can eventually replace `Space` with it. We could make it clear in the doc-strings that this is the plan. Let me know what you think?

fenghourun:
> Hey, absolutely no worries. I'm more than happy to answer questions or help guide you through things. What is currently in this PR looks really good and does match the coding style I would expect, so well done :).
> 
> Also I have an idea, which will mean you will be able to PR this sooner without having to fix all the annoying builds, and means we can work on this in smaller steps. That is: Instead of overwriting the definition `Space` we (in a file `Space.Struct`) could define `SpaceStruct` with your definition, and slowly build the API up, so that we can eventually replace `Space` with it. We could make it clear in the doc-strings that this is the plan. Let me know what you think?

Great idea I think an incremental migration here would be more manageable

fenghourun:
@jstoobysmith  I noticed some of the CI steps take quite a while to run, I think it might be a good idea to split them up into separate github actions workflows. That way they can be run in parallel and individual failures will not require running the entire pipeline again. 

I think the `build PhysLean` and `runLinter on PhysLean` can benefit the most from being split up. Let me know if you'd like me to tackle this afterwards!

jstoobysmith:
The problem is that a lot of the CI checks depend on the project being built in Lean, so e.g. `runLinter on PhysLean` won't work unless `build PhysLean` has already built the project. But I agree that the CI is annoyingly slow, and there are certain improvements that can be made. 

What I personally do is run the linters locally using 

```
lake exe lint_all
./scripts/lint-style.sh
```

This is usually faster since you don't have to keep building the whole project every time you run the linters (as the GitHub workflows have to do).

jstoobysmith (approved):
I reordered the author list, but otherwise this looks good to me! I've approved this, so feel free to merge when you are ready. Thanks!

PR #674, opened 2025-07-31
Add reflectionless potential implementation for quantum mechanics

casualPhysics:
Just the outline for the reflectionless potential


jstoobysmith:
Ok, so one thing I haven't mentioned thus far is the linters. These are automatic checks that run on the whole of PhysLean before we merge, they can be run locally via

```
lake exe lint_all
./scripts/lint-style.sh 
```
You can also use `#lint` temporarily at the bottom of the files you modified to pick up some of the errors. 

These are slightly more strict than the ones that happen on GitHub so you may see errors there which don't appear here.  But I think all you should have to do is: 

1. Add a doc-string to your definitions and structures (and each field part of structure corresponding to a definition) 
2. Import this file into the `PhysLean.lean` file in alphabetical order.

jstoobysmith (approved):
Nice! So this past the linters. I've approved it, and have also invited you as an outside collaborator to the project which, once you accept should give you permission to merge - feel free to when you are ready.

PR #689, opened 2025-08-13
feat: OfNat and Nat.Cast instances for Time

morrison-daniel:
This PR adds OfNat and Nat.Cast instances for use in explicit calculations, and the existing Zero and One instances are removed (replacements are automatically generated by the existence of the OfNat instance). Some proofs later are slightly modified to adapt, along with some clean up such as adding @[ext] tag and removing manual ext statements.

### DIFF
--- a/PhysLean/SpaceAndTime/Time/Basic.lean
+++ b/PhysLean/SpaceAndTime/Time/Basic.lean
@@ -118,7 +118,7 @@ instance : LE Time where
   le t1 t2 := t1.val ≤ t2.val
 
 lemma le_def (t1 t2 : Time) :
-  t1 ≤ t2 ↔ t1.val ≤ t2.val := Iff.rfl
+    t1 ≤ t2 ↔ t1.val ≤ t2.val := Iff.rfl
 
 /-!
 ## Basic operations on `Time`.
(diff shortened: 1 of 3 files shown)


jstoobysmith:
Not sure who I've mentioned this to before, so sorry if I am repeating myself, but for the linters it is quicker to run them locally, rather than waiting for the github workflows. They can be run via: 

```
lake exe lint_all
./scripts/lint-style.sh
```

PhysLean/SpaceAndTime/Time/Basic.lean
```diff
@@ -36,59 +36,82 @@ or origin.
...
 
-@[ext]
-lemma ext_of {t1 t2 : Time} (h : t1.val = t2.val) :
-    t1 = t2 := by
-  cases t1; cases t2; simp_all
+instance : NatCast Time where
+  natCast n := ⟨n⟩
 
-/-!
+instance {n : ℕ} : OfNat Time n where
```
jstoobysmith, line 56: Could we add a doc string to this similar to the following: 

```quote
The casting of a natural number to an element of `Time`. This corresponds to a choice of 1. zero point in time, and 2. a choice of metric on time (defining `1`). 
```

morrison-daniel: That's fine with me, I really just want it so it's easier to create explicit times for testing purposes.

jstoobysmith: Sounds good to me :).

jstoobysmith (approved):
Great many thanks for this! I've approved it, let me know if/when you want me to merge it.

morrison-daniel:
All good to merge!

PR #703, opened 2025-08-21
feat(CanonicalEnsemble):add differentialEntropy and properties of partitionFunction and entropy

or4nge19:
Proves positivity/normalisation properties: partitionFunction_pos, probability_nonneg, sum_probability_eq_one, and entropy_nonneg. Nonempty ι is required where sums must be nonzero.
Add differentialEntropy

### DIFF
--- a/PhysLean/StatisticalMechanics/CanonicalEnsemble/Basic.lean
+++ b/PhysLean/StatisticalMechanics/CanonicalEnsemble/Basic.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Matteo Cipollina, Joseph Tooby-Smith
+Authors: Joseph Tooby-Smith
 -/
 import PhysLean.Thermodynamics.Temperature.Basic
 import Mathlib.MeasureTheory.Measure.ProbabilityMeasure
@@ -355,11 +355,6 @@ lemma paritionFunction_eq_zero_iff (T : Temperature) [IsFiniteMeasure (𝓒.μBo
 
 open NNReal Constants
 
-lemma partitionFunction_comp_ofβ_apply (β : ℝ≥0) :
-    𝓒.partitionFunction (ofβ β) =
-    (𝓒.μ.withDensity (fun i => ENNReal.ofReal (exp (- β * 𝓒.energy i)))).real Set.univ := by
-  simp only [partitionFunction, μBolt, β_ofβ, neg_mul]
-
 /-!
 
 ## The probability measure
@@ -603,7 +598,7 @@ lemma probability_pos
   have hZpos := partitionFunction_pos (𝓒:=𝓒) (T:=T)
   simp [probability, div_pos, Real.exp_pos, hZpos]
 
-/- A general normalisation statement (integral of the probability density = 1)
+/- TODO: A general normalisation statement (integral of the probability density = 1)
 is already encoded via `μProd` being a probability measure; the finite sum
 version lives in `Finite` : `sum_probability_eq_one`. -/
 
--- a/PhysLean/StatisticalMechanics/CanonicalEnsemble/Basic.lean
+++ b/PhysLean/StatisticalMechanics/CanonicalEnsemble/Basic.lean
@@ -590,9 +590,6 @@ lemma meanEnergy_nsmul (n : ℕ) (T : Temperature)
 noncomputable def differentialEntropy (T : Temperature) : ℝ :=
   - kB * ∫ i, log (probability 𝓒 T i) ∂𝓒.μProd T
 
---@[deprecated differentialEntropy (since := "2025-08-22")]
---noncomputable def entropy (T : Temperature) : ℝ := differentialEntropy (𝓒:=𝓒) T
-
 /-- Probabilities are non-negative,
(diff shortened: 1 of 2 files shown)


PhysLean/StatisticalMechanics/CanonicalEnsemble/Finite.lean
```diff
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Joseph Tooby-Smith
+Authors: Joseph Tooby-Smith, Matteo Cipollina
```
jstoobysmith: Name order to alphabetical.

PhysLean/StatisticalMechanics/CanonicalEnsemble/Finite.lean
```diff
@@ -144,4 +144,68 @@ lemma entropy_of_fintype [IsFinite 𝓒] (T : Temperature) :
   simp [mul_comm]
   exact Integrable.of_finite
 
+/-- The partition function of a finite canonical ensemble is strictly positive.
+We require `[Nonempty ι]`; otherwise for an empty type the sum is zero. -/
+lemma partitionFunction_pos [IsFinite 𝓒] [Nonempty ι] (T : Temperature) :
```
jstoobysmith: Do you think you could prove these results in the general case, i.e. in the CanonicalEnsemble.Basic file?
If not I think we should put the lemmas there with a `sorry` proof and a note that they have being proved in the finite case.

or4nge19: thanks! have generalized where I could, see the comments in the new commits

PhysLean/StatisticalMechanics/CanonicalEnsemble/Basic.lean
```diff
@@ -569,4 +569,81 @@ lemma meanEnergy_nsmul (n : ℕ) (T : Temperature)
...
+        = - kB * (∫ i, Real.log (𝓒.probability T i) ∂𝓒.μProd T) := rfl
+  have hkB : 0 ≤ kB := kB_nonneg
+  have : 0 ≤ - kB * (∫ i, Real.log (𝓒.probability T i) ∂𝓒.μProd T) := by
+    have hIle0 : (∫ i, Real.log (𝓒.probability T i) ∂𝓒.μProd T) ≤ 0 := by
+      simpa [integral_const] using hIntLe
+    exact (mul_nonneg_of_nonpos_of_nonpos (neg_nonpos.mpr hkB) hIle0)
+  simpa [hent]
+
+/-!
+NOTE: Without the hypothesis `∀ i, probability ≤ 1` the (differential) entropy
```
jstoobysmith, line 641: I'm confused by this - do you have an example of where the probability could be greater than one? I.e. what conditions does on the measure allows this?

jstoobysmith: Ah, never mind! I understand, probability here is really the probability density function, there is no requirement on that being less then 1. Maybe we should rename it as such?

or4nge19: I believe the pointwise bound probability ≤ 1 is necessary in the general measure‑theoretic setup because probability is a density wrt an arbitrary measure, and densities can exceed 1 (e.g. uniform on a set of measure < 1). This bound is automatic for finite/counting measures, which is why the finite specialization gives unconditional nonnegativity. Perhaps we can change the statement to talk about “probability density” to make the semantics explicit

jstoobysmith: Yep - agreed :)

or4nge19: for continuous systems I think one normally refers to differential entropy which is not absolute but it depends on the choice of units or the underlying measure μ. It can be negative and does not have the same direct interpretation as an absolute measure of disorder. Here the physical quantities are the differences in entropy as they are independent of these choice

jstoobysmith: Ok, good, should we rename `entropy` in the continuous case as `differentialEntropy` and then in the finite case define `entropy` itself, through `differentialEntropy`. Or we can just get rid of entropy in the continous case completely for now.

jstoobysmith: Also in the `Basic.lean` file would you mind putting the lemmas in relevant sections of the file? So e.g. `partitionFunction_pos` should appear near the definition of `partitionFunction` and amongst the API which follows it.

or4nge19: let me think about it

jstoobysmith: Ok, sounds good to me.

jstoobysmith:
All looks good to me now! Many thanks! Once the linters pass I will approve for merge

or4nge19:
> All looks good to me now! Many thanks! Once the linters pass I will approve for merge

it looks much better now generalized as you suggested :)

jstoobysmith (approved):
Approved - feel free to merge when ready :).

PR #721, opened 2025-08-29
feat: import-graph analysis script

morrison-daniel:
Adds import-graph.py which uses the graph structure generated by the command `lake exe graph physlean.xdot_json`. Note that this requires [graphviz](https://graphviz.org/) to generate, see [here](https://github.com/leanprover-community/import-graph) for full details of import-graph. Allows analysis of imports, dependents, transitive imports, transitive dependents, subject area, and importance of files in the project.

### DIFF
--- a/scripts/import-graph.py
+++ b/scripts/import-graph.py
@@ -1,0 +1,120 @@
+import json
+import pandas as pd
+import matplotlib.pyplot as plt
+
+# JSON files to be imported, generate this with `lake exe graph physlean.xdot_json`
+# See `https://github.com/leanprover-community/import-graph` for more info
+graphFile = "physlean.xdot_json"
+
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Hey! This is great! Thank you. I've approved this, so let me know when you want to merge. 

Just to clarify, you need graphviz for `lake exe graph physlean.xdot_json`, not to run the .py file, correct?

morrison-daniel:
Yes, graphviz is what converts the graph to a json file so you only need it when generating the file. Once that file is generated you don't need graphviz to run the .py. I thought about including the json I generated but elected not to since it was already out of date and PhysLean changes quickly so users are bound to need to generate it themselves in the future.

Merge away!

PR #724, opened 2025-08-31
Feat: Added workflow to generate a treemap.png and save in doc folder

ShlokVaibhav:
Three files added: scripts/treemap_gen.py to generate the treemap file and scripts/treemap_requirements.txt to list python packages. workflow/PhysLeanTreemap.yml contains the workflow which installs python, dependencies, executes the file, generates the png and commits it to docs folder. 
Have verified intended functionality.


jstoobysmith (approved):
Many thanks for this Shlok! I've approved it, let me know if you want it to be merged now :).

PR #737, opened 2025-09-11
feat: update the harmonic oscillator

jstoobysmith:
Updating the harmonic oscillator to prove the Euler-lagrange equations.

### DIFF
--- a/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
+++ b/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
@@ -104,23 +104,23 @@ Here we both define and proof properties related to the angular frequency.
 
 -/
 
-/-- The angular frequence of the classical harmonic osscilator, `ω`, is defined
+/-- The angular frequency  of the classical harmonic osscilator, `ω`, is defined
   as `√(k/m)`. -/
 noncomputable def ω : ℝ := √(S.k / S.m)
 
-/-- The angular frequence of the classical harmonic osscilator is positive. -/
+/-- The angular frequency  of the classical harmonic osscilator is positive. -/
 @[simp]
 lemma ω_pos : 0 < S.ω := sqrt_pos.mpr (div_pos S.k_pos S.m_pos)
 
-/-- The square of the angular frequence of the classical harmonic osscilator is equal to `k/m`. -/
+/-- The square of the angular frequency  of the classical harmonic osscilator is equal to `k/m`. -/
 lemma ω_sq : S.ω^2 = S.k / S.m := by
   rw [ω, sq_sqrt]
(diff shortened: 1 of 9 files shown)


PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean
```diff
@@ -141,33 +139,90 @@ lemma force_eq_linear (x : Space 1) : force S x = - S.k • x := by
...
+open MeasureTheory ContDiff InnerProductSpace Time
 
-/-- A trajectory `x : ℝ → ℝ` satsifies the equation of motion if and only if
-  it is an extrema of the action.
+set_option linter.unusedVariables false in
+/-- The lagrangian of the harmonic oscillator is the kinetic energy minus the potential energy. -/
+@[nolint unusedArguments]
+noncomputable def lagrangian (t : Time) (x :  Space 1) (v : EuclideanSpace ℝ (Fin 1)) :
+   ℝ :=
+  1 / (2 : ℝ) * S.m * ⟪v, v⟫_ℝ - 1 / (2 : ℝ) * S.k * ⟪x, x⟫_ℝ
```
zhikaip: is there a reason why you're not using `kineticEnergy S xₜ t - potentialEnergy S (xₜ t)` here? is it just to avoid unfolding in the following parts or are there other considerations?

zhikaip: I would think the lagrangian should be defined as KE-PE, and do the unfolding in the following `lagrangian_eq` which is used in the rest of the file so it retains both the physical intuition and ease of use. but maybe I'm missing something here

jstoobysmith: No, its because the `lagrangian` needs to be a function of time, position and velocity (i.e. a function of phase space) for the variational calculus stuff, and Euler-Lagrange equations to work. Maybe this would be a good lemma to add though!

jstoobysmith: (Namely it is due to ClassicalMechanics.eulerLagrangeOp)

zhikaip: > No, its because the `lagrangian` needs to be a function of time, position and velocity (i.e. a function of phase space) for the variational calculus stuff, and Euler-Lagrange equations to work. Maybe this would be a good lemma to add though!

I see, indeed maybe it would be nice to include as a separate lemma then (i.e. this expression equals KE - PE)

jstoobysmith: I've included this lemma with more recent commits :).

PR #745, opened 2025-09-20
fix: autocomplete results blends in with the background in light theme

Komyyy:
Screenshot from 2025-09-21 02-53-20

This bug harms user experiences, so this PR fixes this.


jstoobysmith (approved):
Many thanks - I will merge now.

PR #774, opened 2025-10-17
Defined Equivalence relations for PMNS_Dirac 

PrabhodaCS:
.... and added some more diagPhase lemmas


jstoobysmith (approved):
Awesome many thanks for this - approved and merged 🎉

PR #775, opened 2025-10-18
Formalize basic definitions in cosmology. Showed one intermediate lemma to prepare formalizing an informal lemma.

luisgbariuan:


luisgbariuan:
Made changes that passes lint checker

PR #787, opened 2025-10-24
Refactor CarriesDimension

rlamy:
This refactors `CarriesDimension` by adding a generic `HasDim` class, together with a `dim M` notation. 

`CarriesDimension` becomes simply the combination of `HasDim` and `MulAction ℝ≥0 M`, and `ModuleCarriesDimension` can be entirely removed since the `Module` instance implies `MulAction`. Also, `WithDim` doesn't require `MulAction` any more, for consistency.

### DIFF
--- a/PhysLean/Units/Basic.lean
+++ b/PhysLean/Units/Basic.lean
@@ -270,27 +270,23 @@ The latter is need to prevent a typeclass diamond.
 
 -/
 
-/-- A type `M` carries a dimension `d` if every element of `M` is supposed to have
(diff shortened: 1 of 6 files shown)


jstoobysmith:
Many thanks for this PR, I will leave it open before merging for a little bit to see if anyone wants to comment, but otherwise it all looks good to me!

jstoobysmith (approved):
Will merge now. The linter here fails, but it passes when run locally. This linter has been updated, so should passed when merged.

PR #788, opened 2025-10-26
Wrote corresponding intermediate lemmata for spherical geometry case. Defined the first-order Friedmann equation, Hubble constant, and the deceleration parameter.

luisgbariuan:


PhysLean/Cosmology/FLRW/Basic.lean
```diff
@@ -78,25 +88,48 @@ def FLRW : Type := sorry
...
+We will leave it here for generality and accounting purposes.
+
+At time `t` the equation reads:
+`(a' / a)^2 = (8πG/3) ρ − k c^2 / a^2 + Λ c^2 / 3`.
+
+-/
+
+namespace FriedmannEquation
+
+def FirstOrderFriedmann (a: ℝ → ℝ) (ρ k Λ G c: ℝ) (t : ℝ) : Prop :=
```
jstoobysmith: This result needs a doc string

jstoobysmith (approved):
Great. Many thanks for this. I will merge it shortly. Maybe at some point in the future it would be worth replacing ` (t : ℝ)` in some of the equations with `t : Time`, and using the time derivative we have there. This should be an easy change, and can wait to another pr.

PR #817, opened 2025-11-24
Have added files for Landau n Lifshitz Ch1 Sec5 Q1,2,3 with introductory docstring

ShlokVaibhav:
Have added files for Landau n Lifshitz Ch1 Sec5 Q1,2,3 with introductory docstring. Have described the problem briefly, in each case the constraints are identified and few equations leading to the final lagrangian are mentioned. Names of the files were chosen to give an idea of the problem being solved. Looking to hear the feedback and improvements needed.

Help of VScode autocomplete was taken to autocomplete the equations (Though all equations were produced beforehand via pen n paper)

For Q3 part c), I am getting an extra minus sign in a term cos(gamma*t)cos(phi) which is absent from the textbook solution, I am attaching the solution, can someone take a look? 
The textbook solution has no minus sign for the middle term (highlighted in red) while my solution above has an extra minus sign
image


jstoobysmith:
Many thanks for this. Two comments: 

1. I would remove the underscores from the file names. 
2. I would add the standard copyright statement to each file (with your name). The linters should pick this up, and I'm not sure why they are not.

jstoobysmith (approved):
Great, many thanks for this! I've approved this PR. I will merge tomorrow morning (unless it allows you to do it - in which case feel free to).

PR #826, opened 2025-12-02
Added latex code to pendulum files

ShlokVaibhav:
Latex codes have been added to the docstring. The html pages look as follows:
image
image
image


jstoobysmith (approved):
Great, many thanks for this! I have approved it. I will also merge it.

PR #828, opened 2025-12-04
feat: Remove axioms related to units

jstoobysmith:
In this PR we remove a number of axioms related to units, and replace them with explicit definitions. 
This reasoning behind this change is discussed here:

https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/physical.20units/near/534907020

In the end the idea is that we do not introduce any axioms other than those in Mathlib. After this PR there are two axioms left related to the existence of Planck's constant and Boltzmann's constant. These could very easily be removed.

The use of axioms here is also reported to result in errors when interacting with @Aristotle-Harmonic, which this PR should fix (@rlamy).


rlamy:
I can confirm that this PR fixes the issues I had trying to use Aristotle.

jstoobysmith:
@rlamy  Perfect :)

PR #836, opened 2025-12-17
chore: Bump to v4.26.0

jstoobysmith:
Bump to v4.26.0. There are two places where I had to mark the results as @[sorryful] and two places where I increased the heartbeat. 

The places with the @[sorryful] attribute are discussed here: 

https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Fin.2Efind.3F.20and.20decide.20in.20v4.2E26.2E0/with/564068609

Since they don't really effect anything, I think it is easiest to wait until v4.27.0 where these can be fixed it seems.


jstoobysmith:
Closes #835

PR #845, opened 2025-12-27
feat: Implementing 6VZKA for Lorentz Algebra

NicolaBernini:
feat: Added the 6 Generator Matrices and added membership proofs for them being part of Lorentz Algebra

PR #847, opened 2025-12-28
Feat/todo 6 vzf2

NicolaBernini:
feat: Prove ofCliffordAlgebra surjectivity (partial progress on TODO 6VZF2)

It needs to be merged after #845


jstoobysmith (approved):
Great - approved, will merge later. Many thanks for this :).

PR #864, opened 2025-12-29
docs: Add reference for Hamilton's equations from SICM

rht:
I did it in the style of https://github.com/HEPLean/PhysLean/blob/f0b5bb0695aaf737c4f30ae68dd29e3c5a3d1009/PhysLean/Relativity/LorentzGroup/Basic.lean#L17, but put the authors name first before the book title because that's how the citation in BibTeX is done.


jstoobysmith (approved):
Great - approved, many thanks. I will merge this after the linters finish (in the future they will run automatically on your PRs).

PR #865, opened 2025-12-29
chore: fix linter warnings

pitmonticone:


jstoobysmith (approved):
Approved, thanks Pietro! I think you should be able to merge after the linters finish - so feel free too. Else I will do it later :).

PR #868, opened 2025-12-30
ci: Add codespell GH Actions

rht:


jstoobysmith (approved):
Great, many thanks for this, and changing it to an action. I've approved it and will merge it now.

PR #869, opened 2025-12-30
feat: Implementing TODO 6VZQF so Define homomorphism from  to the restricted Lorentz group.

NicolaBernini:
feat: Implementing TODO 6VZQF so Define homomorphism from  to the restricted Lorentz group.


jstoobysmith (approved):
Approved. Many thanks. Will merge shortly.

PR #871, opened 2025-12-30
feat: Specify total derivative equivalence for classical mechanics Lagrangian

rht:
This is to prepare for the derivation of the Lagrangian of a free particle found in Landau vol 1.

### DIFF
--- a/PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
+++ b/PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
@@ -63,11 +63,6 @@ namespace Lagrangian
 
 ## A. General Total Time Derivative
 
-A function δL(r, v, t) is a total time derivative if there exists F(r, t) such that
-δL(r, v, t) = fderiv ℝ F (r, t) (v, 1).
-
-By the chain rule, this equals ∂F/∂t(r, t) + ⟨∇ᵣF(r, t), v⟩.
-
 -/
 
 /-- A function δL(r, v, t) is a total time derivative if it can be written as d/dt F(r, t)
--- a/PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
+++ b/PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
@@ -1,0 +1,102 @@
+/-
+Copyright (c) 2025 Rein Zustand. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Rein Zustand, Joseph Tooby-Smith
+-/
+import Mathlib.Analysis.InnerProductSpace.EuclideanDist
+import Mathlib.Analysis.InnerProductSpace.PiL2
+import Mathlib.Topology.MetricSpace.Basic
+import Mathlib.Analysis.Calculus.Deriv.Basic
+import PhysLean.Mathematics.VariationalCalculus.HasVarGradient
+
+/-!
+
+# Gauge Equivalence of Lagrangians
+
+## i. Overview
+
+Two Lagrangians are physically equivalent if they differ by a total time derivative
+d/dt F(q, t). This is because the Euler-Lagrange equations depend only on extremizing
+the action integral, and total derivatives don't affect which paths are extremal.
+
+This module defines the key concept of a function being a total time derivative,
+which is essential for analyzing symmetries like Galilean invariance.
+
+## ii. Key insight
+
+A general function δL(r, v, t) is a total time derivative if there exists a function
+F(r, t) (independent of velocity) such that:
+  δL(r, v, t) = d/dt F(r, t) = ∂F/∂t + ⟨∇ᵣF, v⟩
+
+For the special case where δL depends only on velocity v (not position or time),
+this implies a strong constraint:
+  δL(v) = ⟨g, v⟩ for some constant vector g
+
+This is because:
+1. d/dt F(r, t) = ∂F/∂t + ⟨∇F, v⟩
+2. For δL to be r-independent, ∇F must be r-independent
+3. For δL to be t-independent, the time-dependent part must vanish
+4. The result is δL = ⟨g, v⟩ for constant g
+
+## iii. Key definitions
+
+- `IsTotalTimeDerivative`: General case for δL(r, v, t)
+- `IsTotalTimeDerivativeVelocity`: Velocity-only case, equivalent to δL(v) = ⟨g, v⟩
+
+## iv. References
+
+- Landau & Lifshitz, "Mechanics", §2 (The principle of least action)
+- Landau & Lifshitz, "Mechanics", §4 (The Lagrangian for a free particle)
+
+-/
+
+namespace ClassicalMechanics
+
+open InnerProductSpace
+
+namespace Lagrangian
+
+/-!
+
+## A. General Total Time Derivative
+
+A function δL(r, v, t) is a total time derivative if there exists F(r, t) such that
+δL(r, v, t) = ∂F/∂t(r, t) + ⟨∇ᵣF(r, t), v⟩.
+
+-/
+
+/-- A function δL(r, v, t) is a total time derivative if it can be written as d/dt F(r, t)
+    for some function F that depends on position and time but not velocity.
+
+    Mathematically: δL(r, v, t) = ∂F/∂t(r, t) + ⟨∇ᵣF(r, t), v⟩
+
+    This is the most general form of gauge equivalence for Lagrangian perturbations.
+    The key point is that F must be independent of velocity. -/
+def IsTotalTimeDerivative {n : ℕ}
+    (δL : EuclideanSpace ℝ (Fin n) → EuclideanSpace ℝ (Fin n) → ℝ → ℝ) : Prop :=
+  ∃ (F : EuclideanSpace ℝ (Fin n) → ℝ → ℝ)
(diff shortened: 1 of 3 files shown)


PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
```diff
@@ -0,0 +1,102 @@
...
+-/
+
+/-- A velocity-dependent function δL is a total time derivative (of a position-time function)
+    iff it has the form δL(v) = ⟨g, v⟩ for some constant vector g.
+
+    This characterization comes from the requirement that:
+    - d/dt F(r, t) = ∂F/∂t + ⟨∇F, ṙ⟩ = ∂F/∂t + ⟨∇F, v⟩
+    - For the result to be independent of r and t, we need ∇F = g (constant) and ∂F/∂t = 0
+    - Thus δL(v) = ⟨g, v⟩ -/
+def IsTotalTimeDerivativeVelocity {n : ℕ} (δL : EuclideanSpace ℝ (Fin n) → ℝ) : Prop :=
```
rht: This one arguably could have been placed in `./PhysLean/ClassicalMechanics/FreeParticle/LandauDerivation.lean` later. Not sure.

PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
```diff
@@ -0,0 +1,102 @@
+/-
+Copyright (c) 2025 Rein Zustand. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Rein Zustand, Joseph Tooby-Smith
```
jstoobysmith: I don't need to be in the author list of this :).

PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
```diff
@@ -0,0 +1,102 @@
...
+-/
+import Mathlib.Analysis.InnerProductSpace.EuclideanDist
+import Mathlib.Analysis.InnerProductSpace.PiL2
+import Mathlib.Topology.MetricSpace.Basic
+import Mathlib.Analysis.Calculus.Deriv.Basic
+import PhysLean.Mathematics.VariationalCalculus.HasVarGradient
+
+/-!
+
+# Gauge Equivalence of Lagrangians
```
jstoobysmith: I don't think "gauge equivalence" is the right word to use here, as it supposes there is a gauge transformation involved, where there might not be. Maybe we should rename this to something like Total derivatives of Lagrangians.

PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-- A velocity-dependent function δL is a total time derivative (of a position-time function)
+    iff it has the form δL(v) = ⟨g, v⟩ for some constant vector g.
+
+    This characterization comes from the requirement that:
+    - d/dt F(r, t) = ∂F/∂t + ⟨∇F, ṙ⟩ = ∂F/∂t + ⟨∇F, v⟩
+    - For the result to be independent of r and t, we need ∇F = g (constant) and ∂F/∂t = 0
+    - Thus δL(v) = ⟨g, v⟩ -/
+def IsTotalTimeDerivativeVelocity {n : ℕ} (δL : EuclideanSpace ℝ (Fin n) → ℝ) : Prop :=
+  ∃ g : EuclideanSpace ℝ (Fin n), ∀ v, δL v = ⟪g, v⟫_ℝ
```
jstoobysmith: Should this not be a theorem based on the previous definition?

PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
rht: I dug into the usage of the term. That the "Lagrangian is equivalent up to a total time derivative" is sometimes written as being gauge equivalent in older texts/papers and the field theory literature. But I agree that the total time derivative doesn't always show up because of a gauge transformation. Will change this.

rht: I renamed the file to TotalDerivativeEquivalence.lean, if you agree with this.

PhysLean/ClassicalMechanics/Lagrangian/GaugeEquivalence.lean
rht: Changed to a theorem that contains a `sorry`.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
```diff
@@ -0,0 +1,114 @@
...
+
+    If δL depends only on velocity and equals d/dt F(r, t) for some F,
+    then δL(v) = ⟨g, v⟩ for some constant vector g.
+
+    This characterization comes from the requirement that:
+    - d/dt F(r, t) = ∂F/∂t + ⟨∇F, ṙ⟩ = ∂F/∂t + ⟨∇F, v⟩
+    - For the result to be independent of r and t, we need ∇F = g (constant) and ∂F/∂t = 0
+    - Thus δL(v) = ⟨g, v⟩
+
+    WLOG, we assume `δL 0 = 0` since constants are total derivatives (c = d/dt(c·t))
```
rht, line 103: This should simplify the proof, but let me know if this is not recommended.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
```diff
@@ -0,0 +1,114 @@
...
+    and can be absorbed without affecting the equations of motion. -/
+theorem isTotalTimeDerivativeVelocity {n : ℕ}
```
jstoobysmith: Can we make this a lemma, and then will need a @[sorryful] attribute to satisfy the linters.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
jstoobysmith: If it can be proved without this, it is probably better to work in the more general case.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
rht: OK, I wasn't aware that the convention in Lean is that `theorem` is reserved for major, highly significant results.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
rht: Opus 4.5 was able to generate a clean proof only when I added this assumption. But I didn't include the proof in the PR because a human could probably do it better with/without the `δL 0 = 0` assumption, and that I couldn't grok the proof by myself at my current level.

For reference, the proof is
```lean
  -- Extract F and the key property: δL v = fderiv ℝ F (r, t) (v, 1) for all r, t
  obtain ⟨F, _, hδL⟩ := h
  -- Work with the derivative at the origin (0, 0)
  let dF := fderiv ℝ F (0, 0)
  -- Since δL 0 = 0, the time-derivative component vanishes
  have h_time : dF (0, 1) = 0 := by
    have := hδL 0 0 0
    simp only at this
    rw [← this, hδL0]
  -- Therefore δL v = dF (v, 0) (only the spatial part contributes)
  have h_spatial : ∀ v, δL v = dF (v, 0) := fun v => by
    have := hδL 0 v 0
    simp only at this
    rw [this, show (v, (1 : ℝ)) = (v, 0) + (0, 1) from by simp,
        dF.map_add, h_time, add_zero]
  -- Define the linear functional L : v ↦ dF (v, 0)
  let L : EuclideanSpace ℝ (Fin n) →L[ℝ] ℝ := dF.comp (ContinuousLinearMap.inl ℝ _ ℝ)
  -- By Riesz representation theorem, L v = ⟪g, v⟫ for some g
  use (InnerProductSpace.toDual ℝ (EuclideanSpace ℝ (Fin n))).symm L
  intro v
  rw [h_spatial v]
  -- Show ⟪toDual.symm L, v⟫ = L v = dF (v, 0)
  have h1 : ⟪(InnerProductSpace.toDual ℝ (EuclideanSpace ℝ (Fin n))).symm L, v⟫_ℝ = L v := by
    conv_lhs => rw [← InnerProductSpace.toDual_apply_apply]
    rw [LinearIsometryEquiv.apply_symm_apply]
  rw [h1]
  rfl
```

jstoobysmith (approved):
Great. This all looks good to me know. Thank you for being receptive to comments. I've approved and will merge shortly.

PhysLean/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
jstoobysmith: In Mathlib they do use `theorem` for everything (although I think this may change in the future), but yes in PhysLean we only use `theorem` for major results.

jstoobysmith:
Merged - we should probably eventually make this an actual equivalence relation on Lagrangians.

PR #873, opened 2026-01-01
feat: Implementing 7ESNL

NicolaBernini:
# Solution for TODO 7ESNL: Adding `inv_perserve_color` to Simp Database

## Problem

The lemma `PermCond.inv_perserve_color` at `PhysLean/Relativity/Tensors/Basic.lean:571-576` was identified as a useful simplification rule but could not be added to the simp database due to the `simpVarHead` linter warning.

The lemma states:
```lean
lemma PermCond.inv_perserve_color {n m : ℕ} {c : Fin n → C} {c1 : Fin m → C}
    {σ : Fin m → Fin n} (h : PermCond c c1 σ) (x : Fin n) :
    c1 (h.inv σ x) = c x
```

The issue is that the left-hand side `c1 (h.inv σ x)` has **`c1` as the head symbol**, which is a variable (parameter) rather than a constant. The `simpVarHead` linter warns against simp lemmas with variable heads because:
- They can cause performance issues in the simplifier
- The simplifier must attempt matching against many more terms when the head is a variable

## Solution Approach

After investigation, we determined that adding this lemma to the simp database is justified despite the variable head:

1. **Active Usage**: The lemma is used in 3 locations in the codebase:
   - `Relativity/Tensors/Basic.lean:592` (proof of `PermCond.symm`)
   - `Relativity/Tensors/Basic.lean:766` (in `permT_basis_repr_symm_apply` with `by simp`)
   - `Relativity/Tensors/ComplexTensor/OfRat.lean:169`

2. **Precedent**: The codebase already has a similar case at `Relativity/Tensors/Contraction/Pure.lean:201` where `@[simp, nolint simpVarHead]` is used for a lemma with variable head.

3. **Proof Automation**: The lemma is specifically used in `by simp` contexts (line 766), indicating it provides genuine simplification benefits.

## Justification for `nolint simpVarHead`

The `nolint simpVarHead` directive is justified because:
- The lemma is narrowly scoped to `PermCond` contexts, limiting potential performance impact
- It enables cleaner proofs in tensor permutation code
- There is existing precedent in the codebase for similar usage
- The variable head `c1` appears with specific structure `c1 (h.inv σ x)`, making pattern matching reasonably selective

### DIFF
--- a/PhysLean/Relativity/Tensors/Basic.lean
+++ b/PhysLean/Relativity/Tensors/Basic.lean
@@ -565,9 +565,7 @@ lemma PermCond.preserve_color {n m : ℕ} {c : Fin n → C} {c1 : Fin m → C}
   simp only [Function.comp_apply]
   rw [h.2]
 
-TODO "7ESNL" "We want to add `inv_perserve_color` to Simp database, however this fires the linter
-    simpVarHead. This should be investigated."
-
+@[simp, nolint simpVarHead]
 lemma PermCond.inv_perserve_color {n m : ℕ} {c : Fin n → C} {c1 : Fin m → C}
     {σ : Fin m → Fin n} (h : PermCond c c1 σ) (x : Fin n) :
     c1 (h.inv σ x) = c x := by


jstoobysmith (approved):
Great! Many thanks for the explanation, this makes a lot of sense, and makes it very clear what is happening here :). Approved, and will merge shortly.

PR #891, opened 2026-01-07
Define energy and energy dissipation rate-DHO04

rfikani:
Prove that energy is not conserved and derive the energy dissipation rate.


jstoobysmith (approved):
Looks good to me. Although you don't actually prove these things here - which would eventually be great to see in Lean! Nevertheless this is a useful PR, so I have approved it and will merge later :).

PR #894, opened 2026-01-09
fix sign error

MilesCranmer:
(first contribution as promised!)

Schwartz annihilation operator looks to have the wrong sign:


jstoobysmith (approved):
Nice! Many thanks Miles, and thanks for the chat yesterday.

(I've approved the PR and will merge it once the linters are done, so nothing more to do on your part here.)

Issue #880, opened 2026-01-05
API: Z-prime charges for Standard Model


jstoobysmith:
## Key data structure 

The key data structure here is the collection of integer charges one can assign to the Standard Model particles. Sets of charges which are multiples of each other should be equivalent.

- [ ] is defined 

## Need

Z-prime models are an important yet simple extension of the standard model. They have been utilized in many phenomenological problems. Having this data structure will allow us to reason more easily about these models. 

## Requirements

- [ ] The API shall contain an action of the group which permutes families of particles 
- [ ] The API shall contain a function returning the charges associated with different terms in the potential 
- [ ] The API shall contain a function given the anomaly cancellation conditions on the charges 
- [ ] The API shall contain an equivalence class on the charges by addition of hypercharge 
- [ ] The API shall contain common named sets of charges

## Corresponding file system

Lots of aspect of this API already exist in PhysLean, however, it is not structured around a key data structure, so needs to be rewritten. This content can be found as part of: 

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/Particles/StandardModel

## Parent APIs

#858

PR #900, opened 2026-01-10
TODO 6VZMW Add topological derivation interface for anomaly cancellation conditions

ValentinBredemestre:
Introduce a `TopologicalDerivation` structure recording anomaly cancellation conditions as linear, quadratic and cubic homogeneous forms arising from a formal derivation (e.g. triangle diagrams or index-theoretic methods).

Provide a canonical map from `TopologicalDerivation` to `ACCSystem`, together with basic simp lemmas, and update the document structure accordingly.

### DIFF
--- a/PhysLean/QFT/AnomalyCancellation/Basic.lean
+++ b/PhysLean/QFT/AnomalyCancellation/Basic.lean
@@ -579,50 +579,23 @@ Anomaly cancellation conditions can be derived formally from the gauge group and
 fermionic representations. They arise from triangle Feynman diagrams, and can also be
 obtained using index-theoretic or characteristic-class constructions.
 
-In this file, we do not formalize the full topological input. Instead, we record the
-output of such a derivation as linear, quadratic and cubic homogeneous forms on the
-space of charges. This is sufficient to build an `ACCSystem`.
+In this file, we do not formalize the full topological input. Instead, we define
+`TopologicalDerivation` as an abbreviation for `ACCSystem`, used to record the intended
+origin of the anomaly cancellation conditions.
 
 -/
 
 /--
-A package of anomaly cancellation conditions as they could arise from a formal derivation
-(e.g. from a gauge group and fermionic representations).
-
-This structure records the resulting homogeneous forms on the charge space.
+An abbreviation for `ACCSystem`, used to record the intended origin that the anomaly
+cancellation conditions arise from a formal derivation (e.g. from a gauge group and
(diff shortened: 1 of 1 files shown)


jstoobysmith:
Is there a difference between `ACCSystem` and `TopologicalDerivation` here? If not, I'm not sure we need the latter.

ValentinBredemestre:
Sorry for the late reply, I’ve been quite busy recently. : /

This is my first time handling a TODO in the project, so I’m still learning the conventions and best practices here. Thanks a lot for the comment and for pointing this out, I really appreciate it. = )

I’ve updated the code to avoid duplicating ACCSystem and instead introduced TopologicalDerivation as a lightweight abbreviation, whose purpose is only to record the intended origin of the anomaly cancellation conditions. I hope this aligns better with the overall design. = )

jstoobysmith:
No worries :). This is probably one of the harder TODOs one could have picked :). I think what is probably best, because the documentation you have included is nice is the following: 

1. Keep the TODO item in place, but add more detail based on your documentation for the "## J" subheading.
2. Don't define `TopologicalDerivation` explicitly, but add some information to the documentation string of `ACCSystem` stating that it is what should be derived (etc.). 

If this makes sense.

ValentinBredemestre:
Thanks a lot, that makes sense. = )

Before working on TODO 6VZMW more concretely, I just wanted to check what level of formalization you would expect there. In particular, would you envision something like a lightweight abstract structure encoding “gauge data” together with a map producing an ACCSystem, or is the intention more to keep this purely at the level of documentation for now?

I want to make sure any further work aligns well with the intended scope of this file and the project. = )

jstoobysmith:
Awesome, thanks for your edits.

```
In particular, would you envision something like a lightweight abstract structure encoding “gauge data” together with a map producing an ACCSystem, or is the intention more to keep this purely at the level of documentation for now?
```
I think the former. Ideally one would have a data type corresponding to a gauge field, and then be able to construct the triangle diagrams or use the topological index theorems to get from that to the ACC conditions. There are likely easier intermediate steps that could be done here though and maybe a start would be to break this TODO down into smaller TODO items.

jstoobysmith (approved):
Approved - and will merge now :).

jstoobysmith:
More generally, in this direction is this GitHub issue #880 which involves refactoring all of the code around these anomaly cancellation conditions, to make them centered on a key data structure.

PR #913, opened 2026-01-17
Introduce Z-prime charge assignments for Standard Model fermions

ValentinBredemestre:
Introduce the key data structure for Z′-models over the Standard Model: integer charge assignments for the chiral SM fermions with n families.

This provides a lightweight foundation for the Z′ API (issue #880), on which family symmetries, anomaly cancellation conditions, and equivalence relations will be built in subsequent work.

### DIFF
--- a/PhysLean/Particles/StandardModel/ZPrime/Charges.lean
+++ b/PhysLean/Particles/StandardModel/ZPrime/Charges.lean
@@ -1,0 +1,46 @@
+/-
+Copyright (c) 2025 Joseph Tooby-Smith. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joseph Tooby-Smith
+-/
+import Mathlib.Data.Int.Basic
+import Mathlib.Data.Fin.Basic
+/-!
+
+# Z-prime charges for the Standard Model
+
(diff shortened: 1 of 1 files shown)


PhysLean/Particles/StandardModel/ZPrime/Charges.lean
```diff
@@ -0,0 +1,46 @@
...
+`n` family copies. Thus a charge assignment is a function `Fin 5 → Fin n → ℤ`.
+
+This file is intentionally lightweight: it provides the underlying data structure and
+basic projections. Further API components (family permutations, potential-term charges,
+anomaly cancellation conditions, quotients by scaling/hypercharge) are developed in
+subsequent files/PRs.
+-/
+
+/-- Integer Z′-charges for the `n`-family Standard Model fermions (without RHN). -/
+abbrev SMZPrimeCharges (n : ℕ) : Type := Fin 5 → Fin n → ℤ
```
jstoobysmith, line 25: Looks, good, although I would make this a structure not an abbreviation. You could even make the fields of the abbreviation `Q`, `U`, `D` etc. I would also make a note in 

https://github.com/HEPLean/PhysLean/blob/master/PhysLean/Particles/StandardModel/AnomalyCancellation/Basic.lean

About this new definition of things

PR #918, opened 2026-01-22
feat: Complete TightBindingChain Hamiltonian hermiticity and eigenstate orthogonality proofs

physlean0:
## Summary

This PR completes two key sorryful lemmas in the tight binding chain model:

- **`hamiltonian_hermitian`**: Proves that the tight binding Hamiltonian is Hermitian by showing that the inner products satisfy `⟨Hψ, φ⟩ = ⟨ψ, Hφ⟩`
  - Adds helper lemma `localizedComp_adjoint`: adjoint of |m⟩⟨n| is |n⟩⟨m|
  - Adds helper lemma `localizedComp_self_adjoint`: diagonal terms are self-adjoint

- **`energyEigenstate_orthogonal`**: Proves energy eigenstates with distinct wavenumbers are orthogonal using the geometric series identity
  - The key insight is that the inner product reduces to ∑ωⁿ where ω is an N-th root of unity (from periodic boundary conditions)
  - Different wavenumbers give ω ≠ 1, and ∑ωⁿ = (ωᴺ - 1)/(ω - 1) = 0

These proofs formalize fundamental quantum mechanical results about the tight binding model commonly used in condensed matter physics.

## Test plan

- [x] `lake build PhysLean.CondensedMatter.TightBindingChain.Basic` compiles successfully
- [x] No new sorrys introduced
- [x] Removes 2 `@[sorryful]` annotations

### DIFF
--- a/PhysLean/CondensedMatter/TightBindingChain/Basic.lean
+++ b/PhysLean/CondensedMatter/TightBindingChain/Basic.lean
@@ -193,6 +193,13 @@ lemma localizedComp_apply_localizedState (m n p : Fin T.N) :
   rw [localizedComp, LinearMap.coe_mk, AddHom.coe_mk,
     orthonormal_iff_ite.mp T.localizedState_orthonormal n p, ite_smul, one_smul, zero_smul]
 
+/-- The adjoint of localizedComp |m⟩⟨n| is |n⟩⟨m|. -/
+lemma localizedComp_adjoint (m n : Fin T.N) (ψ φ : T.HilbertSpace) :
+    ⟪|m⟩⟨n| ψ, φ⟫_ℂ = ⟪ψ, |n⟩⟨m| φ⟫_ℂ := by
+  simp only [localizedComp, LinearMap.coe_mk, AddHom.coe_mk]
+  rw [inner_smul_left, inner_smul_right, inner_conj_symm]
+  ring
+
 /-!
 
 ## D. The Hamiltonian of the tight binding chain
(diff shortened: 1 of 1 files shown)


PhysLean/CondensedMatter/TightBindingChain/Basic.lean
```diff
@@ -212,10 +212,42 @@ noncomputable def hamiltonian : T.HilbertSpace →ₗ[ℂ] T.HilbertSpace :=
...
+/-- The adjoint of localizedComp |m⟩⟨n| is |n⟩⟨m|. -/
+lemma localizedComp_adjoint (m n : Fin T.N) (ψ φ : T.HilbertSpace) :
+    ⟪|m⟩⟨n| ψ, φ⟫_ℂ = ⟪ψ, |n⟩⟨m| φ⟫_ℂ := by
+  simp only [localizedComp, LinearMap.coe_mk, AddHom.coe_mk]
+  rw [inner_smul_left, inner_smul_right]
+  rw [inner_conj_symm]
+  ring
+
+/-- The diagonal term |n⟩⟨n| is self-adjoint. -/
+lemma localizedComp_self_adjoint (n : Fin T.N) (ψ φ : T.HilbertSpace) :
```
jstoobysmith: I think we can delete this, and replace it below just with `localizedComp_adjoint`

PhysLean/CondensedMatter/TightBindingChain/Basic.lean
```diff
@@ -212,10 +212,42 @@ noncomputable def hamiltonian : T.HilbertSpace →ₗ[ℂ] T.HilbertSpace :=
 
 -/
 
+/-- The adjoint of localizedComp |m⟩⟨n| is |n⟩⟨m|. -/
+lemma localizedComp_adjoint (m n : Fin T.N) (ψ φ : T.HilbertSpace) :
```
jstoobysmith: I think this can be moved up into section `C`.

jstoobysmith (commented):
Added some comments - if you could handle these it would be great, if not let me know and I can do them.

PhysLean/CondensedMatter/TightBindingChain/Basic.lean
```diff
@@ -478,10 +510,104 @@ noncomputable def energyEigenstate (k : T.QuantaWaveNumber) : T.HilbertSpace :=
...
+
+This is a fundamental quantum mechanical result: eigenstates of a Hermitian operator
+(the Hamiltonian) with distinct eigenvalues are orthogonal. Here we prove it directly
+using the periodic boundary conditions which quantize the wavenumbers.
+
+The key physical insight is that different wavenumbers k₁ ≠ k₂ give rise to different
+N-th roots of unity exp(i(k₂-k₁)a), and the sum of all N-th roots of unity equals zero. -/
 lemma energyEigenstate_orthogonal :
-    Pairwise fun k1 k2 => ⟪T.energyEigenstate k1, T.energyEigenstate k2⟫_ℂ = 0 := by sorry
+    Pairwise fun k1 k2 => ⟪T.energyEigenstate k1, T.energyEigenstate k2⟫_ℂ = 0 := by
```
jstoobysmith, line 516: I would also be interested in knowing if Claude could 'golf' this proof to make it shorter.

physlean0: Yes! The proof has been golfed from 84 lines to 47 lines (37 lines shorter). Key changes:

- Replaced `match k1, k2 with | ⟨_, hk1⟩, ⟨_, hk2⟩ =>` pattern with cleaner `obtain ⟨_, ⟨n1, rfl⟩⟩ := k1` destructuring
- Combined initial `simp only [energyEigenstate]` and `rw [sum_inner]` into single `simp only`
- Used `calc` block for `hω_pow` instead of separate `have heq` + `rw`
- Inlined `hm_eq` computation directly into `hm_int` proof
- Replaced `apply Finset.sum_congr rfl; intro i _` with `refine Finset.sum_congr rfl fun i _ => ?_`
- Removed verbose intermediate comments while preserving the key physical insights in the docstring

The proof structure remains the same (geometric series argument using periodic boundary conditions), just more concise.

jstoobysmith: Great! Would you be able to replace the proof in this PR with that shorter one?

jstoobysmith (approved):
Approved :). Many thanks for this first PR - hope it is the first of many! I will merge it shortly (sorry I think I missed that you had already replaced the proof).

PR #919, opened 2026-01-22
7RKA6: Replaced informal with sorryful

kastch:


kastch:
If the approach is fine, I'm going to similarly convert other informal lemmas there, before switching to proofs

jstoobysmith:
This approach is more than fine, and breaking prs up like this is nice and recommended

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -117,10 +117,26 @@ lemma toComplex_equivariant {n} {c : Fin n → realLorentzTensor.Color}
 
 -/
 
+/-- The `PermCond` condition is preserved under `colorToComplex`. -/
+@[simp] lemma PermCond_colorToComplex {n m : ℕ}
```
jstoobysmith: Can we make this start with a lower case P rather then upper case

kastch:
@jstoobysmith Thanks! I was more concerned about the style and approach of the lemma formalization itself, rather then the pr split. Glad to hear all is fine with that!

jstoobysmith (approved):
Approved this - many thanks, will merge now

PR #920, opened 2026-01-23
7RKFF: Replaced informal with sorryful

kastch:


PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -122,10 +122,46 @@ informal_lemma permT_toComplex where
...
+  refine Fin.addCases (fun i => ?_) (fun j => ?_) x
+  · -- left case: x = castAdd m i
+    -- here `simp` should expand `Fin.append` on castAdd
+    simp [Fin.append, Function.comp_apply]
+  · -- right case: x = natAdd n j
+    simp [Fin.append, Function.comp_apply]
+
+/-- `prodT` on the complex side, with colors written as `colorToComplex ∘ Fin.append ...`.
+This is `prodT` followed by a cast using `colorToComplex_append`. -/
+noncomputable def prodT_colorToComplex {n m : ℕ}
```
jstoobysmith: Could this be `prodTColorToComplex` or something with a snappier name? We follow a convention where definitions do not have underscores.

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -122,10 +122,46 @@ informal_lemma permT_toComplex where
...
+    {c : Fin n → realLorentzTensor.Color} {c1 : Fin m → realLorentzTensor.Color} :
+    ℂT(colorToComplex ∘ c) → ℂT(colorToComplex ∘ c1) → ℂT(colorToComplex ∘ Fin.append c c1) :=
+  fun x y =>
+    cast
```
jstoobysmith: I think permT would be better here than `cast`. In general `cast` is difficult to work with.

kastch: permT changes the ordering of tensor indices using a reindexing function, but here the indices stay the same and we only need to convert between two equal ways of describing the same color assignment, so we must transport across a type equality (a cast), not apply a permutation. I'm thinking maybe to have a separate thin layer with cast behind the  prodTColorToComplex (which in case would be much better for future use). I'll think, maybe I'll come to another solution eventually here.

jstoobysmith: I agree with this, however, permT can also handle this special case where the indices stay the same. In particular this seems to work

```lean
noncomputable def prodT_colorToComplex {n m : ℕ}
    {c : Fin n → realLorentzTensor.Color} {c1 : Fin m → realLorentzTensor.Color} :
    ℂT(colorToComplex ∘ c) → ℂT(colorToComplex ∘ c1) → ℂT(colorToComplex ∘ Fin.append c c1) :=
    fun x y => permT (S := complexLorentzTensor) id (by simp) <|
      prodT (S := complexLorentzTensor) x y
 ```
 
 I think this has the advantage (at least until a better solution is found) that there are lots of relevant machinery which can be used with `permT`, which can't be used with `cast`.

jstoobysmith (approved):
Many thanks for the changes - I've approved and will merge shortly.

PR #921, opened 2026-01-23
7RKFR: Replaced informal with sorryful

kastch:


jstoobysmith (dismissed):
Looks good to me! Many thanks for these - I think there are many improvement which could be made to this tensor/index-notation formalization - so its great to see steps in this direction! Will merge once linters finish (if they pass).

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -143,10 +143,41 @@ informal_lemma prodT_toComplex where
...
+    (complexLorentzTensor).τ (colorToComplex x) = colorToComplex ((realLorentzTensor).τ x) := by
+  cases x <;> rfl
+
+/-- The contraction condition is preserved under `colorToComplex`. -/
+@[simp]
+lemma contrCond_colorToComplex {n : ℕ}
+    {c : Fin (n + 1 + 1) → realLorentzTensor.Color} {i j : Fin (n + 1 + 1)}
+    (h : i ≠ j ∧ (realLorentzTensor).τ (c i) = c j) :
+    i ≠ j ∧ (complexLorentzTensor).τ ((colorToComplex ∘ c) i) = (colorToComplex ∘ c) j := by
+  refine And.intro h.1 ?_
```
jstoobysmith: linter is saying that simp only [*] should  be able to show this, in which case it may not be needed explicitly here and we could just used simp in `contrT_toComplex`.

kastch: The idea was to make it more stable for the potential interface changes, but I'm fine with minimality here, will push the fix in a few

jstoobysmith (approved):

PR #922, opened 2026-01-23
7RKGK: Replaced informal with sorryful

kastch:


jstoobysmith (approved):
Great! Looks good to me. Many thanks, will merge shortly.

PR #935, opened 2026-02-10
Remove sorry: Prove realLorentzTensor.prodT_toComplex

kastch:


jstoobysmith:
Great! This looks good. 

One thing I am getting a bit concerned about related to this file is that I think it is getting a bit complicated to read and understand (this is no fault of yours). I think we could do with some sub headings 
```lean 
/--

## ....
-/ 
```
which separate the different parts, and make it clearer what is being done where. Is this something you could have a go at in this PR? or at the least, I think we should put a TODO item related to this point.

kastch:
Hi @jstoobysmith ! 
I'll address it here, we need to do it with the file anyway, there are 3 more sorry results there and they would not be small. 

Should I just mark it with something like 
/-!

### prodT commutes with toComplex section

-/

or do you want me to do something way deeper?
Please let me know your thoughts.
Thanks!

jstoobysmith:
I think even this would be a good start, if we could do something similar with the other parts of the file, this would be good. 

I good example would be similar to:

https://github.com/lean-phys-community/PhysLean/blob/master/PhysLean/ClassicalMechanics/HarmonicOscillator/Basic.lean

kastch:
Hey @jstoobysmith ! I've added a brief meme/intro/contents at the beginning of the file and a few things across the text, feel free to review it!

jstoobysmith (approved):
Looks great! Many thanks for adding the comments. Approved, will merge shortly

Issue #887, opened 2026-01-06
API: Fluid


jstoobysmith:
## Key data structure 

The key data structure is a fluid specified by the distribution of its fluid velocity, density and pressure. 

- [ ] is defined

## Need

This is one of the core definitions fluid dynamics and can be used to define many other properties in this area.

## Requirements

- [ ] Flow out of a volume 
- [ ] Mass flux density 
- [ ] Equation of continuity 
- [ ] Euler's equation 
- [ ] Definition of an ideal fluid 
- [ ] Definition of the entropy flux density 
- [ ] Bernoulli's equation 
- [ ] Energy flux

## Corresponding file system

This does not yet have an API in PhysLean 

## References 

A good reference for this is 

- Landau and Lifshitz, Fluid Mechanics, Chapter 1.
## Parent APIs

#854 #855

mog1el:
I'm going to work on this for the next couple of weeks. If anyone is interested in collaborating, let me know!

jstoobysmith:
Closing as included in code as API-map, this should be followed instead of this PR.

PR #949, opened 2026-02-16
Initial IdealFluid structure

mog1el:
- [x] is well defined
- [x] Flow out of a volume
- [x] Mass flux density
- [x] Equation of continuity
- [x] Euler's equation
- [x] Definition of an ideal fluid
- [x] Definition of the entropy flux density
- [x] Bernoulli's equation
- [x] Energy flux

Addressing issue #887


PhysLean/FluidMechanics/Basic.lean
```diff
@@ -0,0 +1,23 @@
...
+import PhysLean.SpaceAndTime.Time.Basic
+/-!
+This module introduces the idea of an ideal fluid and the mass flux density. The rest of the early workings is to be seen in ./IdealFluid/
+-/
+open scoped InnerProductSpace
+
+structure IdealFluid where
+  density: Time → Space → ℝ
+  velocity: Time → Space→ Space
+  pressure: Time → Space → ℝ
```
jstoobysmith: You might want to think about the conditions that you put here. Do e.g. L&L suggest that the density need by a smooth function? Does it need to have certain properties with respect to integration? etc.

PhysLean/FluidMechanics/Basic.lean
```diff
@@ -0,0 +1,23 @@
...
+import PhysLean.SpaceAndTime.Space.Basic
+import PhysLean.SpaceAndTime.Time.Basic
+/-!
+This module introduces the idea of an ideal fluid and the mass flux density. The rest of the early workings is to be seen in ./IdealFluid/
+-/
+open scoped InnerProductSpace
+
+structure IdealFluid where
+  density: Time → Space → ℝ
+  velocity: Time → Space→ Space
```
jstoobysmith: I would make velocity a map to `EluclideanSpace _ _` rather than `Space`.

PhysLean/FluidMechanics/Basic.lean
```diff
@@ -0,0 +1,23 @@
+/-
+Copyright (c) 2026 Michał Mogielnicki. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Michał Mogielnicki
+-/
```
jstoobysmith: Would move this `.Basic.lean` file into the directory `IdealFluid`

jstoobysmith:
Please can you let me know when you want this reviewing again.

jstoobysmith:
So may main comment is that everything which is nominally a definition (so `def`, `structure` etc.) will need a documentation string for the linters to pass.

jstoobysmith:
Linters will fail as you need to add your new files to PhysLean.lean

jstoobysmith:
Hey, noticed some commits :). Let me know if you need any help.

PR #951, opened 2026-02-18
chore: remove unused simp lemmas

zhikaip:
I thought this will get picked up by the github build checks


jstoobysmith (approved):
Approved. Was this emitting a warning? Likely the github workflows don't break for just warnings (rather than errors), maybe this is something we can change.

zhikaip:
> Approved. Was this emitting a warning? Likely the github workflows don't break for just warnings (rather than errors), maybe this is something we can change.

Yes I noticed this during build

PR #959, opened 2026-02-24
feat: Improve docs on space

jstoobysmith:
Improving the documentation on `Space` in accordance with the conversation here: 

https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Space.20vs.20EuclideanSpace/with/575332888

### DIFF
--- a/PhysLean/SpaceAndTime/Space/Basic.lean
+++ b/PhysLean/SpaceAndTime/Space/Basic.lean
@@ -29,7 +29,7 @@ on the Lean Zulip, including:
 
 There is a choice between defining `Space d` as an `abbrev` of `EuclideanSpace ℝ (Fin d)`,
 as a `def` of a type with value `EuclideanSpace ℝ (Fin d)` or as a structure
-with a field `val : Fin d → ℝ`.
+with a field `val : Fin d → ℝ` :
 
 +---------------------------------------------------+---------+-------+-----------+
 |                                                   | abbrev  |  def  | structure |
@@ -39,8 +39,8 @@ with a field `val : Fin d → ℝ`.
 | requires reproving of lemmas from `EuclideanSpace`|   no    |  yes  |    yes    |
 +---------------------------------------------------+---------+-------+-----------+
 
-The `structure` is the most conservative choice, as everything needs redefining, however,
-there is are two benefits of this:
+The `structure` is the most conservative choice, as everything needs redefining. However,
+there is are two benefits of using it:
 
 1. It allows us to be precise about the instances we define on `Space d`, and makes
   future refactoring of those instances easier.
@@ -49,7 +49,7 @@ there is are two benefits of this:
 
 Currently `Space d` has the instances of `Module` (which requires the choice
 of a zero), `Norm` and `InnerProductSpace`.
-A future refactor should instead give `Space d` the instance of a `NormedAddTorsor` and a
+A future refactor should instead give `Space d` the instance of `NormedAddTorsor` and
 `MetricSpace` giving it directly the Euclidean distance.
 
 This has not been done yet since `fderiv` requires a `Module` instance.
--- a/PhysLean/SpaceAndTime/Space/Basic.lean
+++ b/PhysLean/SpaceAndTime/Space/Basic.lean
@@ -51,7 +51,7 @@ Currently `Space d` has the instance of a `Module` (which requires the choice
 of a zero), a future refactor should give `Space d` only the instance of an `AddTorsor`
 (an affine space) which does not require the choice of a zero, in such a way that `Space d` should be a special case of Riemannian manifolds. This has not been done
 yet since `fderiv` requires a `Module` instance.
-
+Similarly, `Norm` `InnerProductSpace` instances should be replaced by the `MetricSpace` instance giving directly the Euclidean distance.
 Because of this, one should be careful to avoid using the explicit zero in `Space d`,
 or adding two `Space d` values together. Where possible one should use
 the `VAdd (EuclideanSpace ℝ (Fin d)) (Space d)` instance instead.
--- a/PhysLean/SpaceAndTime/Space/Basic.lean
+++ b/PhysLean/SpaceAndTime/Space/Basic.lean
@@ -48,7 +48,7 @@ there is are two benefits of this:
   would not otherwise be possible if we reuse results from Mathlib.
 
 Currently `Space d` has the instance of a `Module` (which requires the choice
-of a zero), a future refactor should give `Space d` the instance of an `AddTorsor`
+of a zero), a future refactor should give `Space d` only the instance of an `AddTorsor`
 (an affine space) which does not require the choice of a zero. This has not been done
 yet since `fderiv` requires a `Module` instance.
 
--- a/PhysLean/SpaceAndTime/Space/Basic.lean
+++ b/PhysLean/SpaceAndTime/Space/Basic.lean
@@ -15,7 +15,7 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
 # Space
 
 In this module, we define the the type `Space d` which corresponds
-to a `d`-dimensional Euclidean space and prove some properties about it.
+to `d`-dimensional flat Euclidean space and prove some properties about it.
 
 PhysLean sits downstream of Mathlib, and above we import the necessary Mathlib modules
 which contain (perhaps transitively through imports) the definitions and theorems we need.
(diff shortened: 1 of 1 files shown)


PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -20,8 +20,45 @@ to a `d`-dimensional Euclidean space and prove some properties about it.
...
+The `structure` is the most conservative choice, as everything needs redefining, however,
+there is are two benefits of this:
+
+1. It allows us to be precise about the instances we define on `Space d`, and makes
+  future refactoring of those instances easier.
+2. It allows us to give the necessary physics context to results about `Space d`, which
+  would not otherwise be possible if we reuse results from Mathlib.
+
+Currently `Space d` has the instance of a `Module` (which requires the choice
+of a zero), a future refactor should give `Space d` the instance of an `AddTorsor`
```
eric-wieser:
```suggestion
of a zero), a future refactor should give `Space d` only the instance of an `AddTorsor`
```
since the instance already exists via `Module` right now

jstoobysmith: Fixed, thanks Eric.

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -33,8 +70,11 @@ TODO "HB6RR" "In the above documentation describe the notion of a type, and
 
 TODO "HB6VC" "Convert `Space` from an `abbrev` to a `def`."
 
-/-- The type `Space d` represents `d` dimensional Euclidean space.
-  The default value of `d` is `3`. Thus `Space = Space 3`. -/
+/-- The type `Space d` is the world-volume which corresponds to
+ `d` dimensional Euclidean space with a given (but arbitrary)
```
yoh-tanimoto:
```suggestion
 `d` dimensional (flat) Euclidean space with a given (but arbitrary)
```

yoh-tanimoto: and change the module docstring accordingly

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -20,8 +20,45 @@ to a `d`-dimensional Euclidean space and prove some properties about it.
...
+there is are two benefits of this:
+
+1. It allows us to be precise about the instances we define on `Space d`, and makes
+  future refactoring of those instances easier.
+2. It allows us to give the necessary physics context to results about `Space d`, which
+  would not otherwise be possible if we reuse results from Mathlib.
+
+Currently `Space d` has the instance of a `Module` (which requires the choice
+of a zero), a future refactor should give `Space d` only the instance of an `AddTorsor`
+(an affine space) which does not require the choice of a zero. This has not been done
```
yoh-tanimoto:
```suggestion
(an affine space) which does not require the choice of a zero, in such a way that `Space d` should be a special case of Riemannian manifolds. This has not been done
```
perhaps like this?

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -20,8 +20,45 @@ to a `d`-dimensional Euclidean space and prove some properties about it.
...
+yet since `fderiv` requires a `Module` instance.
+
```
yoh-tanimoto, line 56:
```suggestion
Similarly, `Norm` `InnerProductSpace` instances should be replaced by the `MetricSpace` instance giving directly the Euclidean distance.
```
if only "physical" quantities are allowed.

yoh-tanimoto (approved):

PhysLean/SpaceAndTime/Space/Basic.lean
eric-wieser: `NormedAddTorsor` is probably the right thing to mention.

jstoobysmith: I've updated this, would appreciate it if one of you could check what I have written.

yoh-tanimoto: looks good to me!

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -15,11 +15,49 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
...
+There is a choice between defining `Space d` as an `abbrev` of `EuclideanSpace ℝ (Fin d)`,
+as a `def` of a type with value `EuclideanSpace ℝ (Fin d)` or as a structure
+with a field `val : Fin d → ℝ`.
+
++---------------------------------------------------+---------+-------+-----------+
+|                                                   | abbrev  |  def  | structure |
++---------------------------------------------------+---------+-------+-----------+
+| allows casting from `EuclideanSpace`              |   yes   |  yes  |    no     |
+| carries instances from `EuclideanSpace`           |   yes   |  no   |    no     |
+| requires reproving of lemmas from `EuclideanSpace`|   no    |  no   |    yes    |
```
eric-wieser:
```suggestion
| requires reproving of lemmas from `EuclideanSpace`|   no    |  yes  |    yes    |
```
unless you want to get in trouble with lean. In new lean versions (4.29.0rcX) this is increasingly the case.

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -15,11 +15,49 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
...
+## Implementation details
+
+The exact implementation of `Space d` into PhysLean is discussed in numerous places
+on the Lean Zulip, including:
+
+- https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Space.20vs.20EuclideanSpace/with/575332888
+
+There is a choice between defining `Space d` as an `abbrev` of `EuclideanSpace ℝ (Fin d)`,
+as a `def` of a type with value `EuclideanSpace ℝ (Fin d)` or as a structure
+with a field `val : Fin d → ℝ`.
```
eric-wieser: Arguably `val : EuclideanSpace ℝ (Fin d)` is a nicer option here, which means you don't have to redefine all the structure.

jstoobysmith: But I think it means you have to use WithLp.ofLp etc. everywhere. Could one not get the same benefit in terms of structure by defining an equivalence from `Space d` to `EuclideanSpace ℝ (Fin d)`?

eric-wieser: Can you give an example of where you'd need to write `WithLp.ofLp`? I think it's registered as a function coercion so that you don't have to write it.

jstoobysmith: My bad, I meant `WithLp.toLp`, I agree that you shouldn't need `WithLp.ofLp`

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -15,11 +15,49 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
...
++---------------------------------------------------+---------+-------+-----------+
+|                                                   | abbrev  |  def  | structure |
++---------------------------------------------------+---------+-------+-----------+
+| allows casting from `EuclideanSpace`              |   yes   |  yes  |    no     |
+| carries instances from `EuclideanSpace`           |   yes   |  no   |    no     |
+| requires reproving of lemmas from `EuclideanSpace`|   no    |  yes  |    yes    |
++---------------------------------------------------+---------+-------+-----------+
+
+The `structure` is the most conservative choice, as everything needs redefining, however,
+there is are two benefits of this:
```
zhikaip:
```suggestion
there are two benefits of this:
```
extra word

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -15,11 +15,49 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
...
+
+1. It allows us to be precise about the instances we define on `Space d`, and makes
+  future refactoring of those instances easier.
+2. It allows us to give the necessary physics context to results about `Space d`, which
+  would not otherwise be possible if we reuse results from Mathlib.
+
+Currently `Space d` has the instances of `Module` (which requires the choice
+of a zero), `Norm` and `InnerProductSpace`.
+A future refactor should instead give `Space d` the instance of a `NormedAddTorsor` and a
+`MetricSpace` giving it directly the Euclidean distance.
```
zhikaip:
```suggestion
A future refactor should instead give `Space d` the instance of `NormedAddTorsor` and
`MetricSpace` giving it directly the Euclidean distance.
```
consistency with previous sentence

PhysLean/SpaceAndTime/Space/Basic.lean
zhikaip:
```suggestion
with a field `val : Fin d → ℝ`:
```

PhysLean/SpaceAndTime/Space/Basic.lean
```diff
@@ -15,11 +15,49 @@ import Mathlib.Analysis.InnerProductSpace.Calculus
...
+
++---------------------------------------------------+---------+-------+-----------+
+|                                                   | abbrev  |  def  | structure |
++---------------------------------------------------+---------+-------+-----------+
+| allows casting from `EuclideanSpace`              |   yes   |  yes  |    no     |
+| carries instances from `EuclideanSpace`           |   yes   |  no   |    no     |
+| requires reproving of lemmas from `EuclideanSpace`|   no    |  yes  |    yes    |
++---------------------------------------------------+---------+-------+-----------+
+
+The `structure` is the most conservative choice, as everything needs redefining, however,
```
zhikaip:
```suggestion
The `structure` is the most conservative choice, as everything needs redefining. However,
```

zhikaip (changes requested):
some minor suggestions otherwise looks good

PhysLean/SpaceAndTime/Space/Basic.lean
jstoobysmith: Fixed

PhysLean/SpaceAndTime/Space/Basic.lean
jstoobysmith: Fixed, and appended the end of that sentence

PhysLean/SpaceAndTime/Space/Basic.lean
jstoobysmith: Fixed

zhikaip (approved):
Thanks, approved.

PR #957, opened 2026-02-21
feat: Hilbert space & unbounded operators on Space

gloges:
Initial progress for single-particle QM on `Space d`, addressing #851.

`SpaceDHilbertSpace` and `SchwartzSubmodule` carry over from 1d with only minor changes (cf. [OneDimension/HilbertSpace](https://github.com/lean-phys-community/PhysLean/tree/master/PhysLean/QuantumMechanics/OneDimension/HilbertSpace)).
`UnboundedOperator` is defined using [LinearPMap](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/InnerProductSpace/LinearPMap.html) with additional density and closability hypotheses. Nothing in `UnboundedOperator` depends on the particular Hilbert space and can be easily generalized in the future.

A few sorries remain to be filled in; two I think I can do after understanding how to work with [InnerProductSpace/PiL2](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/InnerProductSpace/PiL2.html).

Issue #851, opened 2025-12-29
API: QM system on Space


jstoobysmith:
## Key data structure 

The key data structure here corresponds to Hilbert space of 'square integrable functions from `Space d` to the complex numbers`. Really one will need the equivalence class of such functions.

- [ ] is defined

## Need

This API will be used in, for example, the study of the classical harmonic oscillator, the reflectionless potential and other problems based on potentials in space.

## Requirements

- [ ] The API shall contain the definition of the position operator 
- [ ] The API shall contain the definition of the momentum operator 
- [ ] The key data structure shall carry the instance of a Hilbert space 
- [ ] The API shall contain the definition of a bra 
 
## Corresponding file system

Aspects of this API already exist, however, they are not structured neatly around the key data structure. 

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/QuantumMechanics/OneDimension

## Parent APIs

#854

PR #963, opened 2026-03-01
feat: Position & momentum unbounded operators

gloges:
Continues on from #957 and makes progress on #851.

Introduces `schwartzEquiv`, the `LinearEquiv` between Schwartz functions and their image in the Hilbert space, along with some basic properties. This bijection is then used to define three symmetric unbounded operators with Schwartz submodule for domain, using the previously-defined continuous linear maps on Schwartz functions. These are the position and momentum unbounded operators, defined component-wise, and the (regularized) radius operator to any real power.

There remains one sorry which will have to be revisited. Showing that the momentum operator on Schwartz functions is symmetric requires two results for `Space.deriv`: integration-by-parts over all of `Space d` and commutation through complex conjugation.

### DIFF
--- a/PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
+++ b/PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
@@ -12,11 +12,6 @@ import PhysLean.SpaceAndTime.Space.Derivatives.Basic
 # Momentum vector operator
 
 In this module we define:
-- The momentum operator on Schwartz maps, component-wise.
-- The momentum squared operator on Schwartz maps.
-
-
-In this module we define:
 - `momentumOperator i` acting on Schwartz maps `𝓢(Space d, ℂ)` as `-Iℏ∂ᵢ`.
 - `momentumOperatorSqr` acting on Schwartz maps `𝓢(Space d, ℂ)` as `∑ᵢ 𝐩[i]∘𝐩[i]`.
 - `momentumUnboundedOperator i`, a symmetric unbounded operator acting on the Schwartz submodule
@@ -25,7 +20,6 @@ In this module we define:
 We also introduce the following notation:
 - `𝐩[i]` for `momentumOperator i`
 - `𝐩²` for `momentumOperatorSqr`
-
 
 -/
 
--- a/PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
+++ b/PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
@@ -3,8 +3,10 @@ Copyright (c) 2026 Gregory J. Loges. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Gregory J. Loges
 -/
-import PhysLean.SpaceAndTime.Space.Derivatives.Basic
+import PhysLean.QuantumMechanics.DDimensions.Operators.Unbounded
+import PhysLean.QuantumMechanics.DDimensions.SpaceDHilbertSpace.SchwartzSubmodule
 import PhysLean.QuantumMechanics.PlanckConstant
+import PhysLean.SpaceAndTime.Space.Derivatives.Basic
 /-!
 
 # Momentum vector operator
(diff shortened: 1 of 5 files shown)


morrison-daniel:
Apologies for not getting to this sooner, but I have been thinking about your PR. I think everything you've done here is reasonable in terms of the existing API and how the literature is written but I wonder if it makes sense for us to change up how `UnboundedOperator` is defined to make things smoother for Lean formalization. What doesn't feel quite right to me is
```lean
/-- The position operators defined on the Schwartz submodule. -/
def positionOperatorSchwartz : schwartzSubmodule d →ₗ[ℂ] schwartzSubmodule d :=
  (schwartzEquiv (d := d)) ∘ₗ 𝐱[i].toLinearMap ∘ₗ (schwartzEquiv (d := d)).symm
```
Like I said, why you'd do this makes sense with how things are designed, but needing to use `schwartzEquiv` to move back and forth between the Schwartz subspace seems like a complication that is going to bite us in the long run. I wonder if it would make more sense to have `UnboundedOperator` between two different spaces, even if we only intend to use them between the same space because Lean sees the submodule it's defined on as a different space than the entire Hilbert space. So I'm thinking it might be smart to adjust the API before we get too deep.

Unfortunately that would delay this PR but I think it might save some headaches in the long run. Does this reasoning make sense, and do you think this would be a productive change?

jstoobysmith:
@morrison-daniel Could you give your suggested definition for UnboundedOperator here?

gloges:
> I wonder if it would make more sense to have UnboundedOperator between two different spaces, even if we only intend to use them between the same space

I think this change,
```lean
structure UnboundedOperator ... extends LinearPMap ℂ HS1 HS2 where ...
```
could be done relatively easily, but this wouldn't remove the need to map between Schwartz maps and their $$L^2$$ equivalence classes. Maybe I have misunderstood your comment?

morrison-daniel:
You are correct, I realized what I was suggesting wouldn't actually remove what I wanted to remove. I have a tendency to be overly cautious to whether something is being done in the best way which is overkill at this point... Ignore my previous statement I'll review, but it's already looking pretty good!

PhysLean/QuantumMechanics/DDimensions/Operators/Position.lean
```diff
@@ -113,23 +129,68 @@ lemma radiusRegPowOperator_comp_eq (hε : 0 < ε) (s t : ℝ) :
...
+/-- The (regularized) radius operators defined on the Schwartz submodule. -/
+def radiusRegPowOperatorSchwartz (ε s : ℝ) : schwartzSubmodule d →ₗ[ℂ] schwartzSubmodule d :=
+  (schwartzEquiv (d := d)) ∘ₗ (radiusRegPowOperator (d := d) ε s).toLinearMap ∘ₗ
+  (schwartzEquiv (d := d)).symm
+
+lemma radiusRegPowOperatorSchwartz_isSymmetric (ε s : ℝ) (hε : 0 < ε) :
+    (radiusRegPowOperatorSchwartz (d := d) ε s).IsSymmetric := by
+  intro ψ ψ'
+  obtain ⟨_, rfl⟩ := schwartzEquiv_exists ψ
+  obtain ⟨_, rfl⟩ := schwartzEquiv_exists ψ'
```
morrison-daniel:
```suggestion
  obtain ⟨_, rfl⟩ := schwartzEquiv.surjective ψ
  obtain ⟨_, rfl⟩ := schwartzEquiv.surjective ψ'
```

PhysLean/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
```diff
@@ -20,29 +20,45 @@ open MeasureTheory
 open InnerProductSpace
 open SchwartzMap
 
-/-- The continuous linear map including Schwartz functions into `SpaceDHilbertSpace d`. -/
-def schwartzIncl {d : ℕ} : 𝓢(Space d, ℂ) →L[ℂ] SpaceDHilbertSpace d := toLpCLM ℂ (E := Space d) ℂ 2
-
-lemma schwartzIncl_injective {d : ℕ} : Function.Injective (schwartzIncl (d := d)) :=
-  injective_toLp (E := Space d) 2
+variable {d : ℕ}
```
morrison-daniel, line 23:
```suggestion
variable (d : ℕ)
```
I wonder if it'd make sense to make this explicit because, while `d` can be inferred when applying to a specific function, we're using it a lot for operators and need to include `(d := d)` a lot. I can see it either way, you can choose which you prefer.

You can also decide if you'd like to make the same change on the operators.

PhysLean/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
```diff
@@ -20,29 +20,45 @@ open MeasureTheory
...
+  apply integral_congr_ae
+  filter_upwards [schwartzEquiv_coe_ae f, schwartzEquiv_coe_ae g] with _ hf hg
+  simp [hf, hg, mul_comm]
+
+lemma schwartzEquiv_ae_eq (h : schwartzEquiv f =ᵐ[volume] schwartzEquiv g) : f = g :=
+  (EmbeddingLike.apply_eq_iff_eq _).mp (SetLike.coe_eq_coe.mp (ext_iff.mpr h))
+
+lemma schwartzEquiv_exists (ψ : schwartzSubmodule d) : ∃ f, ψ = schwartzEquiv f := by
+  use schwartzEquiv.symm ψ
+  exact (LinearEquiv.apply_symm_apply _ _).symm
```
morrison-daniel: This is basically identical to `schwartzEquiv.surjective` so I think this is redundant.

PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
```diff
@@ -21,33 +29,67 @@ open Constants
...
+
+/-- The momentum operators defined on the Schwartz submodule. -/
+def momentumOperatorSchwartz : schwartzSubmodule d →ₗ[ℂ] schwartzSubmodule d :=
+  (schwartzEquiv (d := d)) ∘ₗ 𝐩[i].toLinearMap ∘ₗ (schwartzEquiv (d := d)).symm
+
+@[sorryful]
+lemma momentumOperatorSchwartz_isSymmetric : (momentumOperatorSchwartz i).IsSymmetric := by
+  intro ψ ψ'
+  obtain ⟨f, rfl⟩ := schwartzEquiv_exists ψ
+  obtain ⟨f', rfl⟩ := schwartzEquiv_exists ψ'
```
morrison-daniel:
```suggestion
  obtain ⟨f, rfl⟩ := schwartzEquiv.surjective ψ
  obtain ⟨f', rfl⟩ := schwartzEquiv.surjective ψ'
```

PhysLean/QuantumMechanics/DDimensions/Operators/Position.lean
gloges: Changed

PhysLean/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
gloges: Thanks, removed!

PhysLean/QuantumMechanics/DDimensions/Operators/Momentum.lean
gloges: Changed

PhysLean/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
gloges: I figured out that writing `schwartzEquiv.toLinearMap` correctly infers `d`, so I think we can keep it implicit for now. The only place that `(d := d)` remains is in `abbrev schwartzSubmodule`.

morrison-daniel (approved):

morrison-daniel:
Looks good! I'll merge if you're done making changes.

PR #969, opened 2026-03-04
feat: PositiveTemperature (closed for large commit size)

ichxorya:
Ref: #958 #861 #860 

In this PR I will construct relevant API for PositiveTemperature as discussed in #958


ichxorya:
@jstoobysmith I would like to ask for direction from here. As of now, the code in Temperature is compilable. However, when it comes to the dependency at CanonicalEnsemble, I don't know what would actually work. Can you break the necessary rework for me (i.e. in terms of types, theorems, lemmas) so that I can follow it to finish this PR.

@hungle0503 would you also mind looking at this too?

Note that a vibe-coded solution is available at https://github.com/leanprover-community/physlib/commit/c853862b3f6aca8024515bf20aff3a460fb98c35 but I think it is wrong since it defines this

```lean
/-- Construct a `PositiveTemperature` from a real-valued inverse temperature,
defaulting to a junk value for non-positive inputs. Used for `derivWithin` helpers
that need a total function `ℝ → PositiveTemperature`. -/
noncomputable def ofβ_real (b : ℝ) : PositiveTemperature :=
  if h : 0 < b then ofβ ⟨b, h⟩ else ofβ ⟨1, one_pos⟩
```

Thus, it shouldn't be right due to that junk value when we define the total function (could a partial function work?)

jstoobysmith:
Hi @ichxorya, the problem with this PR is that it is still too big, which makes it very hard to review properly. In the end it will be quicker for everyone if we split this up into multiple PRs (aim for like 200-500 lines of changes each). I would start with just submitting the changes in: 

[PhysLean/Thermodynamics/Temperature/Basic.lean‎](https://github.com/leanprover-community/physlib/pull/969#diff-a45f7a8cf0067283a4209011a5b764f0639ddc2746c655153835c17a4d1d462d)

ichxorya:
> Hi @ichxorya, the problem with this PR is that it is still too big, which makes it very hard to review properly. In the end it will be quicker for everyone if we split this up into multiple PRs (aim for like 200-500 lines of changes each). I would start with just submitting the changes in: 
> 
> [PhysLean/Thermodynamics/Temperature/Basic.lean‎](https://github.com/leanprover-community/physlib/pull/969#diff-a45f7a8cf0067283a4209011a5b764f0639ddc2746c655153835c17a4d1d462d)

Thank you, I would follow that!

PR #930, opened 2026-02-08
feat: QM operator commutators

gloges:
Continuing from #926:
1. positionOperator made sorry-free
2. deriv_smul is replaced by two related lemmas (deriv_const_smul is needed in momentum commutator proof)
3. Proofs added for commutators

PR #970, opened 2026-03-05
golf: QM commutators

gloges:
- Golfing for commutation lemmas (cf. #930).
- Improved docs for [Operators/Commutation.lean](PhysLean/QuantumMechanics/DDimensions/Operators/Commutation.lean).

In making some commutators `simp` lemmas a few proofs in [Hydrogen/LaplaceRungeLenzVector.lean](PhysLean/QuantumMechanics/DDimensions/Hydrogen/LaplaceRungeLenzVector.lean) broke; I have crudely fixed these but will revisit them for more golfing soon.

### DIFF
--- a/PhysLean/Mathematics/KroneckerDelta.lean
+++ b/PhysLean/Mathematics/KroneckerDelta.lean
@@ -49,11 +49,6 @@ lemma sum_kroneckerDelta' [AddCommGroup M] [Module ℂ M]
 
 lemma sum_kroneckerDelta_self [AddCommGroup M] [Module ℂ M] (c : ℂ) (f : M) :
     ∑ (i : Fin d), (c * δ[i,i]) • f = (d * c) • f := by
-  simp only [kroneckerDelta_self, Complex.ofReal_one, mul_one, Finset.sum_const, Finset.card_univ,
-    Fintype.card_fin]
-  induction d with
-  | zero => simp
-  | succ n hn =>
-    rw [succ_nsmul, hn, ← add_smul, Nat.cast_add_one, add_mul, one_mul]
+  simp [kroneckerDelta_self, ← Nat.cast_smul_eq_nsmul ℂ, smul_smul]
 
 end KroneckerDelta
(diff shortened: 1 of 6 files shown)


morrison-daniel:
I think this is a good example of something that should be broken into several smaller PRs for each set of changes. I think you could break it down like:
- Notation changes
- Documentation changes
- Kronecker delta API
- Commutator API (with the Hydrogen atom fixes)
I think we could do the first three at the same time and then the fourth would probably depend on the third.

I know it takes some more work on your part (I fell into this trap with my first mathlib contributions) but it's really helpful to do in the long run. Several PRs are much easier to review than one big one since we can approve small changes right away (like the notation and documentation) and it's easier to keep track of all the changes being made. It's also more likely that a PR affecting many files at once will have merge conflicts with other PRs.

Usually what I do is have git create a duplicate branch for each separate PR I want to make and then delete extraneous content from each branch, or copy paste bits of code into new branches. For PRs that depend on other PRs you can use the tag `blocked-by-PR` and link to the pre-requisite PR so reviewers know to wait, or sometimes I will create a draft PR while working through the first PRs.

It's extra work but it's a good habit to get into and reviewers will appreciate it!

gloges:
Thanks, yes this is a good opportunity for me to improve my git skills. The macro→notation and documentation changes should be straightforward, but I suspect that in making changes to kroneckerDelta other files may not build. How should I handle this? Should I just change these to sorries (to then be corrected/golfed in a later PR)?

morrison-daniel:
If one change breaks something else that's an instance where you want to group both changes into one PR. It's okay if multiple things need to happen at once, your goal is minimize the number of times this happens. Part of what we do in review is try to make PRs with code that can stand up to changes in the codebase so this is less likely to happen, but it's a bit inevitable at this point since we are doing lots of changes at the very base level of definitions.

I would say do your best to split these up and we'll deal with whatever you come up with. If you want help, feel free to try something and I can give feedback either here or on Zulip.

PS - The rule of thumb I got from mathlib reviewers is it's nice if you can get each PR under 100 lines of difference, and up to 200 lines is about as large as you want to get. Above 200 you really want to consider trying to break things up if possible. This is just a rough estimate and not hard cutoffs by any means, and things like documentation or starting a new file will inflate those numbers without making it harder to review.

PR #971, opened 2026-03-05
docs: Start review guidelines

jstoobysmith:
Added review guidelines, very early draft that needs additions, but hopefully something which can be added to.

### DIFF
--- a/docs/ReviewGuidelines.md
+++ b/docs/ReviewGuidelines.md
@@ -14,7 +14,7 @@ In another programming language these points analogous to not just making sure y
 have a function that does the right thing, but making sure that that function is
 as efficient as possible, uses the best algorithm available etc.
 
-## Organization
+## Organization
 
 - Lemmas and definitions appear in the correct place
 - Modules are easy to read and have a well-defined scope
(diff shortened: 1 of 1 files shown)


morrison-daniel:
I would suggest adding [What to consider when reviewing](https://leanprover-community.github.io/contribute/pr-review.html#what-to-consider-when-reviewing) as well. It's technically written for a reviewer and not an author but I found it useful (especially the examples) when I first got started.

At the end I think it's worth adding something to the end explaining that much of the work on a PR happens during review so it doesn't need to be perfect to start a request, though they obviously should have made effort on their own to stick to these guidelines. I think it's good to be encouraging to new contributors, the review process can be intimidating.

As for length considerations, I have been using the rule of thumb that
- < 50 lines = easy to check PR
- < 100 lines = average sized PR
- 100-200 lines = large PR, okay but try to break up if possible
- 200+ lines = very large PR, should be broken up into smaller pieces
Obviously these are not strict cutoffs but I've found it personally useful to judge where I'm at. Stuff like documentation and file organization can add lines without adding much complexity to review while definitions can be a few lines but usually require more effort than other additions of the same size. Often I will try to split changes in different files into different PRs.

To convince authors to follow length guidelines, it may be worth mentioning that limiting the length of a PR is good because
- Small changes can be merged faster and not get blocked by unrelated work
- Reduces the chance and impact of merge conflicts
- Easier for reviewers to understand

Final suggestion, perhaps it would be worth explaining the basics of the tag system since we don't have automated tools for it at the moment.

morrison-daniel (approved):
I fixed a couple spelling errors and am approving since it looks like no more changes are coming. Of course, feel free to keep editing if you like.

jstoobysmith:
Many thanks, will merge now. We can always add more things later.

PR #972, opened 2026-03-05
feat: AddTorsor instance on space

jstoobysmith:
Added an `AddTorsor` instance on `Space d` along with a number of other little results which are required to define this


jstoobysmith:
Put `awaiting-author` tag as one or two things I want to do here before review needed

jstoobysmith:
There is a lot here to review, but most of it is old code that has just been moved (e.g. everything in the Module file is existing code). What is actually new is the definition of `NormedAddTorsor`, `AddTorsor` and `VSub` in the `Space.Basic` file (see [here](https://github.com/leanprover-community/physlib/blob/a8da78242f93cca6f4ba094f46006ba73daed987/PhysLean/SpaceAndTime/Space/Basic.lean)), as well as documentation in the `Space.Basic` and `Space.Module` file.

morrison-daniel (approved):
I think it's smart to separate the general properties of `Space` from those that depend on a choice of origin and therefore module structure.

PR #978, opened 2026-03-07
feat: Unbounded operators from H1 to H2

gloges:
Generalizes the definition of unbounded operators to allow for different domain and codomain (cf. #957).


morrison-daniel (approved):
Looks good! I think this is a smart generalization to make.

PR #942, opened 2026-02-13
feat: Add surfaces

jstoobysmith:
Defining the line and spherical shell surfaces within the Space directory of PhysLean.

PR #995, opened 2026-03-16
feat: Add `gradSchwartz`

jstoobysmith:
Add the grad of a Schwartz map.

### DIFF
--- a/PhysLean/SpaceAndTime/Space/Derivatives/Grad.lean
+++ b/PhysLean/SpaceAndTime/Space/Derivatives/Grad.lean
@@ -45,6 +45,7 @@ of the input function with respect to each spatial coordinate.
   - B.3. The gradient as a sum over basis vectors
   - B.4. The underlying function of the gradient distribution
   - B.5. The gradient applied to a Schwartz function
+  - B.6. The gradident of a Schwartz map
 
 ## iv. References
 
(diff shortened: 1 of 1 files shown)


jstoobysmith:
t-space-time

jstoobysmith:
RFC

jstoobysmith:
Related to #942

PhysLean/SpaceAndTime/Space/Derivatives/Grad.lean
```diff
@@ -502,4 +503,38 @@ lemma distGrad_apply {d} (f : (Space d) →d[ℝ] ℝ) (ε : 𝓢(Space d, ℝ))
...
+        map_add' v1 v2 := by simp
+        map_smul' a v := by rw [smul_comm]; simp
+        cont := by fun_prop}
+      map_add' v1 v2 := by ext1 v; simp [add_smul]
+      map_smul' a v := by ext1 v; simp [smul_smul]
+      cont := continuous_clm_apply.mpr <| fun y => by
+        simp only [ContinuousLinearMap.coe_mk', LinearMap.coe_mk, AddHom.coe_mk]
+        fun_prop }
+    ∑ i, SchwartzMap.bilinLeftCLM smulCLM (.const (EuclideanSpace.single i 1))
+      (.evalCLM ℝ (Space d) ℝ (basis i) (.fderivCLM ℝ (E := _) (F := ℝ) η))
```
morrison-daniel: I'm guessing this looks this way because we need to move from Schwartz maps on `Space` to Schwartz maps on `EuclideanSpace`. I think for now let's move `smulCLM` to its own definition to make this definition easier to read and to make future lemmas easier to write. I believe there is talk of mathlib modifying the definition of the derivative to allow derivatives of maps on `AddTorsor` into maps on the underlying `Module` which would help clean this up significantly. I think it's a bit ways away so we should use a temp fix but it might be good for us to help push that effort forward.

morrison-daniel (commented):
Besides the challenges of swapping from `Space` to `EuclideanSpace` this looks good.

By the way, you can add/remove multiple tags in a single comment by putting them on different lines.

morrison-daniel:
awaiting-author

PhysLean/SpaceAndTime/Space/Derivatives/Grad.lean
jstoobysmith: Thanks - I golfed this, to get rid of `smulCLM` altogether.

jstoobysmith:
-awaiting-author

morrison-daniel (approved):

PR #983, opened 2026-03-11
feat: mfderiv for Space

jstoobysmith:
Add a new formula for derivatives on space in terms of the origin-independent manifold structure on `Space`. Added a number of other small lemmas needed along the way.

PR #996, opened 2026-03-16
feat: Properties of vadd

jstoobysmith:
This PR contains some properties of vadd on `Space d`. 

- Depends on #983


jstoobysmith:
#1001 replaces this

PR #999, opened 2026-03-20
feat(QM): AddZeroClass for UnboundedOperator

gloges:
- Defines zero as the zero map with full domain.
- Defines addition of unbounded operators, using two junk values:
  - When `D(U1) ∩ D(U2)` is not dense, use the zero with full domain.
  - When `D(U1) ∩ D(U2)` is dense but `U1 + U2` is not closable, use the zero map on `D(U1) ∩ D(U2)`.

  This second junk value will be needed so that smul distributes over addition in all cases. It ensures that  `c • (U1 + U2) = c • U1 + c • U2` holds for `(c : ℂ) = 0` when `D(U1) ∩ D(U2)` is dense, since the RHS is the zero map on `D(U1) ∩ D(U2)` regardless of whether `U1 + U2` is closable or not.

- `add_assoc` proves associativity holds when no junk values are encountered

(cf. #957,  #978)

### DIFF
--- a/PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
+++ b/PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
@@ -175,6 +175,7 @@ noncomputable instance : Add (UnboundedOperator H H') where
       else ⟨⟨U₁.domain ⊓ U₂.domain, 0⟩, hD, isClosable_of_zero rfl⟩
     else 0
 
+/-- The domain of `U₁ + U₂` is `D(U₁) ∩ D(U₂)` when it is dense. -/
 lemma add_domain_of_dense {U₁ U₂ : UnboundedOperator H H'}
     (hD : Dense (U₁.domain ⊓ U₂.domain : Set H)) : (U₁ + U₂).domain = U₁.domain ⊓ U₂.domain := by
   rw [HAdd.hAdd, instHAdd, Add.add, instAdd]
@@ -182,6 +183,7 @@ lemma add_domain_of_dense {U₁ U₂ : UnboundedOperator H H'}
   · simp only [hD, hC, ↓reduceDIte, LinearPMap.add_domain]
   · simp only [hD, hC, ↓reduceDIte]
 
+/-- The junk value for `U₁ + U₂` has domain all of `H` when `D(U₁) ∩ D(U₂)` is not dense. -/
 lemma add_domain_of_not_dense {U₁ U₂ : UnboundedOperator H H'}
     (hD : ¬Dense (U₁.domain ⊓ U₂.domain : Set H)) : (U₁ + U₂).domain = ⊤ := by
   rw [HAdd.hAdd, instHAdd, Add.add, instAdd]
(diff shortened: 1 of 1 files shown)


gloges:
t-quantum-mechanics

gloges:
RFC

PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
```diff
@@ -112,18 +115,150 @@ lemma inner_map_polarization' {T : UnboundedOperator H H} (x y : T.domain) :
...
+  zero_add U := by
+    apply UnboundedOperator.ext
+    rw [← zero_add U.toLinearPMap]
+    exact add_toLinearPMap_of_dense_closable (by simp [U.dense_domain]) (by simp [U.is_closable])
+  add_zero U := by
+    apply UnboundedOperator.ext
+    rw [← add_zero U.toLinearPMap]
+    exact add_toLinearPMap_of_dense_closable (by simp [U.dense_domain]) (by simp [U.is_closable])
+
+lemma add_assoc {U₁ U₂ U₃ : UnboundedOperator H H'}
```
jstoobysmith, line 242: Could we maybe put some comments on these lemmas. For instance, I think this is the statement that if addition does not give a junk value, then it is associative?

PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
```diff
@@ -112,18 +115,150 @@ lemma inner_map_polarization' {T : UnboundedOperator H H} (x y : T.domain) :
...
+    (hD : Dense (U₁.domain ⊓ U₂.domain ⊓ U₃.domain : Set H)) (hC₁₂ : (U₁.1 + U₂.1).IsClosable)
+    (hC₂₃ : (U₂.1 + U₃.1).IsClosable) : U₁ + U₂ + U₃ = U₁ + (U₂ + U₃) := by
+  have hD' : Dense (U₁.domain ⊓ (U₂.domain ⊓ U₃.domain) : Set H) := by simp_all [Set.inter_assoc]
+  have hD₁₂ : Dense (U₁.domain ⊓ U₂.domain : Set H) := Dense.mono Set.inter_subset_left hD
```
jstoobysmith:
```suggestion
  have hD₁₂ : Dense (U₁.domain ⊓ U₂.domain : Set H) := hD.mono Set.inter_subset_left
```

jstoobysmith: etc.

jstoobysmith (commented):
Looks mainly good to me. Couple of comments

jstoobysmith:
awaiting-author

PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
gloges: I've added doc strings to the lemmas. See also comment below.

gloges:
I realized that `add_assoc` could be generalized slightly; addition is associative when neither of the intermediate additions `U1 + U2` and `U2 + U3` is a junk value. The hypotheses should now be easier to supply, as they concern the two pairs separately.

gloges:
-awaiting-author

jstoobysmith (approved):
Approved - many thanks. Will merge shortly
ready-to-merge

jstoobysmith:
ready-to-merge

PR #1000, opened 2026-03-20
feat: Improve evaluation of tensors

jstoobysmith:
Add `[γ]` type syntax to indicate that the index `γ` should be considered as evaluated in a tensor. This is useful in extracting the components of a tensor. It will be used in e.g. electromagnetism.

For example if `γ : Fin (1 + d)` then something like `{A | [γ]}` is a real number rather than a tensor.


jstoobysmith:
RFC

kastch:
Hi @jstoobysmith ! Can't approve it, but definitely useful thing.

morrison-daniel (approved):
I'm not familiar with this part of physlib so I can't say much more than it seems fine stylistically. If you really want I can work through it and give a better review but I'm also fine approving so this isn't blocking you.

jstoobysmith:
Ok, thanks @morrison-daniel. Yeah, I think this is a tricky case, because I am aware that it is very hard to review.  Given that @kastch has also looked at it, I'm going to merge this. I'm also very sure it is correct on a functionality level, as it just replicates code which already exists.

PR #1004, opened 2026-03-25
feat: Adds a diffeormorphism for the harmonic oscillator

jstoobysmith:
Adds a diffeomorphism for the harmonic oscillator configuration space.

### DIFF
--- a/PhysLean/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
+++ b/PhysLean/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
@@ -168,7 +168,8 @@ lemma chartAt_invFun_tangent_eq_toEuclidTangent (x : TangentBundle I Configurati
     toEuclidTangent.symm := by
   ext1 p
   obtain ⟨p', rfl⟩ := toEuclidTangent.surjective p
-  simp
+  simp only [OpenPartialHomeomorph.symm_toPartialEquiv, OpenPartialHomeomorph.coe_coe_symm,
+    Equiv.symm_apply_apply]
   refine (OpenPartialHomeomorph.eq_symm_apply _ ?_ ?_).mp ?_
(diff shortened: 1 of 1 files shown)


PhysLean/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
```diff
@@ -255,6 +255,20 @@ instance : CompleteSpace ConfigurationSpace := by
   classical
   simpa using (FiniteDimensional.complete ℝ ConfigurationSpace)
 
+/-- The diffeomorphism between ℝ and ConfigurationSpace. -/
+noncomputable def toRealDiffeo : Diffeomorph 𝓘(ℝ, ℝ) 𝓘(ℝ, ℝ)  ℝ ConfigurationSpace ω where
+  toFun := fromRealCLM
+  invFun := toRealCLM
+  left_inv := by
+    intro t
+    cases t
```
zhikaip:
```suggestion
```
shouldn't need `cases t` here
(goes into CauSeq.Completion.Cauchy abs which is totally irrelevant here)

zhikaip (commented):
minor comment above

PhysLean/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
jstoobysmith: Thanks - agreed. I also realised it is probably a bit stupid to add this and not simultaneously convert to EuclideanSpace. So will do that as well here.

Bergschaf:
It might be possible to use `Homeomorph.chartedSpace` and `Homeomorph.isManifold` (from this [pr](https://github.com/leanprover-community/mathlib4/pull/40202)) instead of defining the `chartedSpace` manually. (The result that this Homeomorphism is then also a Diffeomorphism could probably also fit into mathlib...)

PR #1006, opened 2026-03-25
Adding QuantumInfo

Timeroot:
This merges in the Lean-QuantumInfo repository, previously hosted at: http://github.com/Timeroot/Lean-QuantumInfo

All of the git history has been "carried over" as a merging of two repositories, hence the very long number of commits dating back a year and a half.

Clearly this is not the standard size PR. :)


ClassicalInfo.lean
```diff
@@ -0,0 +1,5 @@
+/-
```
jstoobysmith: Do we need this file?

QuantumInfo/Regularized.lean
```diff
@@ -0,0 +1,164 @@
+/-
+Copyright (c) 2025 Alex Meiburg. All rights reserved.
+Released under MIT license as described in the file LICENSE.
```
jstoobysmith: Should we worry about the license here. I.e. MIT vs Apache

jstoobysmith (approved):
I've approved this, but will leave it to you to merge :).

Issue #941, opened 2026-02-13
API: Tensor components


jstoobysmith:
## Key data structure

The components of tensor, for example a 2-tensor $T^\mu_\nu$, the components would contain all specifications of the index. 

- [x] is defined as `ComponentIdx`

## Need

This is used to manipulate tensors and extract their components. It is also used with basis of tensors.

## What needs doing

- [ ] The data structure `ComponentIdx` and results related there to are currently spread throughout the `Tensors` directory. They should be moved into their own file system.

## Requirements

- [x] The API shall contain the combination of components based on taking the product of tensors. 
- [x] The API shall contain the restriction of components based on contraction of indices. 

## Corresponding file system

Currently these sit in the file system: 

https://github.com/lean-phys-community/PhysLean/tree/master/PhysLean/Relativity/Tensors

kastch:
Hi @jstoobysmith ! I'm going to start working on the item as we've discussed, what's the process, would you assign it to me? Thanks

jstoobysmith:
Hey @kastch, I assigned you this issue. Not sure if there was another part to your question?

PR #1007, opened 2026-03-26
Remove sorry: Prove realLorentzTensor.toComplex_equivariant

kastch:


PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -83,6 +83,36 @@ def colorToComplex (c : realLorentzTensor.Color) : complexLorentzTensor.Color :=
...
+    realLorentzTensor.basis realLorentzTensor.Color.up = Lorentz.contrBasisFin (d := 3) := rfl
+
+lemma basis_eq_coBasisFin :
+    realLorentzTensor.basis realLorentzTensor.Color.down = Lorentz.coBasisFin (d := 3) := rfl
+
+lemma FD_obj_up :
+    realLorentzTensor.FD.obj { as := realLorentzTensor.Color.up } = Lorentz.Contr 3 := rfl
+
+lemma FD_obj_down :
+    realLorentzTensor.FD.obj { as := realLorentzTensor.Color.down } = Lorentz.Co 3 := rfl
```
jstoobysmith: I think these could be done done with been a different file (maybe the same where `realLorentzTensor` is defined?

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -235,11 +444,141 @@ Finally we record that `toComplex` is equivariant for the natural action of
...
+      erw [Lorentz.complexCoBasis_ρ_apply]
+      exact (toComplex_equivariant_slot_repr_down k h Λ b i).symm
+  · simp [P]
+  · intro r t ht
+    dsimp [P] at ht ⊢
+    rw [toComplex_map_smul, Tensor.actionT_smul (S := complexLorentzTensor), ht,
+      Tensor.actionT_smul (S := realLorentzTensor), toComplex_map_smul]
+  · intro t1 t2 h1 h2
+    dsimp [P] at h1 h2 ⊢
+    rw [map_add, Tensor.actionT_add (S := complexLorentzTensor), h1, h2,
```
jstoobysmith, line 507: What is your opinion on whether these proofs can be golfed or split into smaller lemmas which may be more generally useful?

jstoobysmith (commented):
Sorry - was going to review this yesterday but didn't quite get around to it.

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
kastch: Hi @jstoobysmith ! I've extracted some lemmas that make sense to reuse, others are more like one-purpose and technical, to make it through all lean formalities, but if you think that it requires more decomposition - please let me know.

PhysLean/Relativity/Tensors/ComplexTensor/SlotRepr.lean
```diff
@@ -0,0 +1,86 @@
...
+
+@[expose] public section
+
+namespace complexLorentzTensor
+
+open TensorSpecies Tensor Lorentz Lorentz.SL2C Matrix MatrixGroups Complex CategoryTheory Module
+
+variable {n : ℕ} {c : Fin n → complexLorentzTensor.Color}
+
+/-- Basis coefficient for the Lorentz action on a contravariant (`up`) vector slot,
```
jstoobysmith: Could we make these doc-strings a bit more descriptive, I'm not entirely sure what these lemmas are doing here.

PhysLean/Relativity/Tensors/ComplexTensor/SlotRepr.lean
```diff
@@ -0,0 +1,111 @@
...
+  simp only [complexLorentzTensor.basis_up_eq]
+  erw [Lorentz.complexContrBasis_reindex_apply_eq_fin4]
+  simp_rw [complexLorentzTensor.basis_eq_complexContrBasisFin4,
+    Lorentz.complexContrBasisFin4_eq_reindex]
+  rw [Basis.repr_reindex_apply, Basis.reindex_apply]
+  simp_rw [complexLorentzTensor.FD_obj_up]
+  conv_lhs => erw [← LinearMap.toMatrix_apply]
+  exact (Lorentz.complexContrBasis_ρ_apply (M := Λ)
+    (i := finSumFinEquiv.symm (Fin.cast (by rw [h]; rfl) (i k)))
+    (j := finSumFinEquiv.symm (Fin.cast (by rw [h]; rfl) (b k))))
```
jstoobysmith: I don't actually think you need the `k` in this lemma. I.e. just define what would be equivalent to `c k`, `b k` and `i k`.  This should generalize this, and it should still work where you need it to. 

I also think it might be worth working out why all the `erw`'s are needed, as they make the proof pretty difficult. I think it is down to some def-eq abuse in some of the lemmas, in particular `complexLorentzTensor.basis_eq_complexContrBasisFin4`.

kastch: Totally agreed. I've simplified lemmas, and moved the whole file to Vector folder for better organization.

PhysLean/Relativity/Tensors/ComplexTensor/Vector/SlotRepr.lean
```diff
@@ -0,0 +1,111 @@
...
+
+open TensorSpecies Tensor Lorentz Lorentz.SL2C Matrix MatrixGroups Complex CategoryTheory Module
+
+/--
+Component formula for the standard contravariant vector slot `Color.up`.
+
+For `b, i : Fin 4`, the `i`-component of `ρ Λ` on `basis Color.up b` equals the corresponding entry
+of `LorentzGroup.toComplex (SL2C.toLorentzGroup Λ)` in `Fin 1 ⊕ Fin 3` coordinates.
+-/
+lemma repr_ρ_basis_vector_up (Λ : SL(2, ℂ)) (b i : Fin 4) :
```
jstoobysmith: Thanks for adding these doc-strings.  I thin we could put these in ...ComplexTensor.Basic for now

PhysLean/Relativity/Tensors/RealTensor/ToComplex.lean
```diff
@@ -235,11 +434,80 @@ Finally we record that `toComplex` is equivariant for the natural action of
 -/
 
 /-- The map `toComplex` is equivariant. -/
-@[sorryful]
 lemma toComplex_equivariant {n} {c : Fin n → realLorentzTensor.Color}
```
jstoobysmith, line 436: would opening the namespace `complexLorentzTensor` help golf these results?

kastch: Hi @jstoobysmith ! Actually it Is open already, but that does not helps too much there - I can use it only for a few things, most of all the long namings are due to either structure fields or due to having the realLorenzTensor (which is where the code resides) and complexLorenzTensor opened at the same time (thus I can't just use Color instead of complexLorentzTensor.Color for instance - it would default to the Real one).
THat definitely could be achieved if needed but with larger refactoring.
If larger refactoring is really needed I suggest doing it as a separate task in Tensors (maybe the one I'm going to make the next PR on - https://github.com/leanprover-community/physlib/issues/941 ) but please let me know your thoughts.

jstoobysmith (approved):
Approving this. Many thanks for iterating with me on this. Will merge shortly

PR #1010, opened 2026-03-28
feat: EM in three dimensions (basic relations)

zhikaip:
Attempt at restating EM results in vector calculus notation linking the more abstract tensorial formulation to the more elementary vector description which both serve demonstration purposes and potentially useful for applied scenarios. this is an example file demonstrating how I think this should look like. I'm aware this is a basically a exact restatement of `electricField_eq` and `magneticField_eq`, but maybe it is still useful as a landing point to this directory.

Next PR will be to state the Maxwell equations in vector calculus form.


PhysLean/Electromagnetism/ThreeDimension/Basic.lean
```diff
@@ -0,0 +1,59 @@
...
+# Fields from potentials
+
+In this section we rewrite the electric and magnetic fields associated to an
+electromagnetic potential in terms of the scalar and vector potentials, using
+the standard vector-calculus expressions.
+
+-/
+
+/-- The electric field written in terms of the scalar and vector potentials as `- ∇ φ - ∂ₜ A`. -/
+theorem electricField_eq_3D :
```
jstoobysmith, line 47: Lemma instead of theorem but otherwise looks good

zhikaip: Hmm I would consider these two be fundamental relations that although simple they're important enough. I can understand that this is more like a def, would you say `theorem` is reserved for results as consequences of the physical setup with non-trivial proofs? But at the same time I feel like the difference should be judged by conceptual importance not by complexity, and here I think they are key in EM.

My perspective is that somewhere down the line we should have a "theorem list", which gives a sufficiently clear overview of what results are in Physlib, and I'd argue these two should be one of the first things people see looking at this module.

jstoobysmith: Ok, I actually think this "theorem list" idea makes a lot of sense. In this sense I agree that these should be theorems. Maybe we should include this in some documentation somewhere, and make a "lemma vs theorem" thread on the Zulip about this. 

Will approve.

jstoobysmith (approved):
Approved - feel free to merge when you are ready, else I will merge later :)

PR #1011, opened 2026-03-28
feat(QM) : Add DistribSMul instances for UnboundedOperator

gloges:
Follow-up to #999 in which addition of unbounded operators was defined.

Adds `SMul ℕ` (with the usual recursive definition) and `DistribSMul ℂ` instances to `UnboundedOperator H H'`.
The proof of `smul_add` amounts to checking several cases to account for the junk values for addition.

### DIFF
--- a/PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
+++ b/PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
@@ -278,17 +278,15 @@ lemma smul_mem_graph_of_mem_smul_graph {E F : Type*} [NormedAddCommGroup E] [Inn
     [NormedAddCommGroup F] [InnerProductSpace ℂ F] {f : LinearPMap ℂ E F} {c : ℂ} (hc : c ≠ 0)
     {x : E × F} (h : x ∈ (c • f).graph.topologicalClosure) :
     (x.1, c⁻¹ • x.2) ∈ f.graph.topologicalClosure := by
-  have {y : f.domain} {z : F} : c • f y = z ↔ f y = c⁻¹ • z := by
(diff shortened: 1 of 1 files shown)


gloges:
t-quantum-mechanics

gloges:
RFC

PhysLean/QuantumMechanics/DDimensions/Operators/Unbounded.lean
```diff
@@ -268,6 +269,97 @@ lemma add_assoc {U₁ U₂ U₃ : UnboundedOperator H H'}
...
+
+Scalar multiplication by natural numbers is defined recursively as usual with `(0 : ℕ) • U = 0` and
+`n.succ • U = n • U + U`. Note that `(0 : ℕ) • U` has domain `⊤` irrespective of the domain of `U`.
+
+Scalar multiplication by complex numbers is inherited from `LinearPMap`. Note that `(c : ℂ) • U` has
+the same domain as `U` for all constants; in particular, `(0 : ℂ) • U ≤ 0` with equality if and only
+if `U.domain = ⊤`.
+-/
+
+noncomputable instance : SMul ℕ (UnboundedOperator H H') := ⟨nsmulRec⟩
```
jstoobysmith: Do we need this? In particular, is it different from the scalar multiplication induced by that by the complex numbers

gloges: They differ only at `n = 0` and `c = 0`; probably best to leave out `SMul ℕ`  to avoid confusion. I've removed this instance and the corresponding comments.

jstoobysmith (approved):
Approved - many thanks for this.

PR #1012, opened 2026-03-29
feat: differentiability lemmas

zhikaip:
Two sets of changes:
1. replace cross product symbol `×` (\x) with `⨯` (\X or \crossproduct). 
2. This PR is preparing for Maxwell's equations in 3D, and these are the minimal set of lemmas I had to add. I'm not sure if these are idiomatic, but hopefully later we could restructure everything in the space and time derivative operators so that these can just be `fun_prop`. 

Two observations:
1. I think a lot of these just require twice-differentiability i.e. `(hf : ContDiff ℝ 2 f)`, but in potential etc. many of them are just assumed to be smooth. Should we make an conscious effort to keep everything to the minimum assumptions?
3. I think the import graphs here are quite long and messy, e.g. making a change in `Space.Derivatives.Basic` take a long time to build everything between this and EM. I'm aware we want to base everything on space and EM does involve a lot of things, but maybe improvements can be made here. I think Mathlib had some efforts on reducing transitive import chains and we can look into that.


jstoobysmith (dismissed):
Looks good to me - approved

jstoobysmith:
```quote 
I think a lot of these just require twice-differentiability i.e. (hf : ContDiff ℝ 2 f), but in potential etc. many of them are just assumed to be smooth. Should we make an conscious effort to keep everything to the minimum assumptions?
```
If possible, I think this would be nice - however I think the variational gradient result does rely on smoothness, so this can likely only be done so far. 

```quote 
I think the import graphs here are quite long and messy, e.g. making a change in Space.Derivatives.Basic take a long time to build everything between this and EM. I'm aware we want to base everything on space and EM does involve a lot of things, but maybe improvements can be made here. I think Mathlib had some efforts on reducing transitive import chains and we can look into that.
```
Yeah - we should see what results sit between the EM files and Space.Derivatives.Basic. I think it really should be just the e.g. grad, curl and div files.

PhysLean/SpaceAndTime/Time/Derivatives.lean
```diff
@@ -111,6 +111,12 @@ lemma deriv_contDiff_of_contDiff {M : Type}
   · fun_prop
   · fun_prop
 
+lemma deriv_contDiff_of_space {M : Type}
+    [NormedAddCommGroup M] [NormedSpace ℝ M] (f : Time → Space d → M) (hf : ContDiff ℝ ∞ ↿f) :
+    ContDiff ℝ ∞ fun (x : Space d) => (∂ₜ fun t => f t x) t := by
```
zhikaip: actually this can be made n + 1 and n as well - generally should we have both versions with n and smooth?

jstoobysmith: I think this is really down to the question about whether \infinty + 1 = \infinity (by definition) in Lean, if so then we should only need the n version I think.

zhikaip: I think it is indeed rfl, I've changed it to n+1.

PhysLean/SpaceAndTime/Space/Derivatives/Grad.lean
```diff
@@ -395,6 +396,21 @@ lemma integrable_isDistBounded_inner_grad_schwartzMap_spherical{dm1 : ℕ}
 
 /-!
 
+### A.14. Differentiability of gradient
+
+-/
+
+open ContDiff
+
+lemma contDiff_grad {n} {f : Space → ℝ} (hf : ContDiff ℝ (n + 1) f) :
```
jstoobysmith, line 406: should we mark these with `@[fun_prop]`?

zhikaip: I think the two I added (deriv_contDiff_of_space  and contDiff_grad) can both be fun_prop? not 100% sure, see commit below

jstoobysmith: Yep, I think this should work here.

jstoobysmith (approved):
Approved.

PR #1013, opened 2026-03-29
feat: Maxwell Equations in 3D

zhikaip:
depends on

- [x] #1012

As mentioned in #1012, ideally the proofs in `faradayLaw` is just fun_prop.

The statement of `isExtrema_iff_gauss_ampere_magneticFieldMatrix` could probably be cleaned up, and there are minor inconsistencies e.g. `ElectromagneticPotential` is implicit but `LorentzCurrentDensity` is a explicit variable.

### DIFF
--- a/PhysLean.lean
+++ b/PhysLean.lean
@@ -42,6 +42,7 @@ public import PhysLean.Electromagnetism.Kinematics.VectorPotential
 public import PhysLean.Electromagnetism.PointParticle.OneDimension
 public import PhysLean.Electromagnetism.PointParticle.ThreeDimension
 public import PhysLean.Electromagnetism.ThreeDimension.Basic
+public import PhysLean.Electromagnetism.ThreeDimension.MaxwellEquations
 public import PhysLean.Electromagnetism.Vacuum.Constant
 public import PhysLean.Electromagnetism.Vacuum.HarmonicWave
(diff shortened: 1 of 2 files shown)


zhikaip:
tried tagging `vectorPotential_contDiff` with fun_prop but still got error saying "fun_prop bug: function expected, got `V : ElectromagneticPotential, type ctor app". I avoided tagging them for now and applied them manually but maybe we could look into this.

Otherwise I think this is ready once #1012 is merged.

jstoobysmith:
Yeah - this `fun_prop` issue is something which the change to the definition of `ElectromagneticPotential` in: 

https://github.com/leanprover-community/physlib/tree/EMUpdates

fixes.

jstoobysmith:
I've forgotten how to do this, but this changes of this PR is currently showing the changes from the already merged PR. I think there is a way to update the history, so it only shows the new changes.

zhikaip:
I think I managed after some git gymnastics, I was experimenting how to split PRs but seems like it's a lot messier than I thought. will try to figure it out next time

jstoobysmith:
Seems like this force-push works, I made a discussion on the zulip about this, because I hope there is an easier way (maybe some repo settings we need to change).

jstoobysmith (approved):
Looks good - approved.

PR #1018, opened 2026-04-01
feat: Move variational calculus

jstoobysmith:
Variational calculus in PhysLib is specifically designed for the physicists. Eventually we will likely move `Mathematics` into `ForMathlib`, but  variational calculus should not be considered in this way.


morrison-daniel (approved):
Looks good

PR #994, opened 2026-03-15
refactor(QM) : commutators

gloges:
Improves / golfs commutator proofs (cf. #930) making use of Kronecker delta api (cf. #979).
Also adds vanishing commutators as simp lemmas. (This breaks a few LRL proofs that I will fix shortly.)

PR #979, opened 2026-03-08
feat: Kronecker delta api

gloges:
Introduces a basic api for the Kronecker delta function.
- I've kept to lemmas which adhere to the Einstein summation rules that indices occur at most twice in each term and free indices match across `=` (unless one side is zero). Hence no `δ[i,j] • f i j = δ[i,j] • f i i` lemma.
- Commutators.lean and LaplaceRungeLenzVector.lean required some fixes, which I have done crudely (w/ sorries for some of the more computation-heavy lemmas); I will revisit these soon.

PR #1019, opened 2026-04-02
refactor(QM): Improve/golf LRL proofs

gloges:
Makes progress in improving/golfing the proofs for the LRL vector, taking advantage of the Kronecker delta api (c.f. #979) and commutator simp lemmas (`@[simp]` added in #994).

### DIFF
--- a/PhysLean/QuantumMechanics/DDimensions/Hydrogen/LaplaceRungeLenzVector.lean
+++ b/PhysLean/QuantumMechanics/DDimensions/Hydrogen/LaplaceRungeLenzVector.lean
@@ -94,20 +94,19 @@ lemma lrlOperator_eq'' (ε : ℝˣ) (i : Fin H.d) : H.lrlOperator ε i = ∑ j, 
   ring
 
 /-- The operator `𝐱ᵢ𝐩ᵢ` on Schwartz maps. -/
-private def positionDotMomentum {d} := ∑ i : Fin d, 𝐱[i] ∘L 𝐩[i]
+private def positionDotMomentum (d : ℕ) := ∑ i : Fin d, 𝐱[i] ∘L 𝐩[i]
 
 /-- The operator `𝐱ᵢ𝐩²` on Schwartz maps. -/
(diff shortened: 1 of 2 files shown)


gloges:
t-quantum-mechanics

gloges:
RFC

PhysLean/QuantumMechanics/DDimensions/Hydrogen/LaplaceRungeLenzVector.lean
```diff
@@ -94,20 +93,19 @@ lemma lrlOperator_eq'' (ε : ℝˣ) (i : Fin H.d) : H.lrlOperator ε i = ∑ j,
   ring
 
 /-- The operator `𝐱ᵢ𝐩ᵢ` on Schwartz maps. -/
-private def positionDotMomentum {d} := ∑ i : Fin d, 𝐱[i] ∘L 𝐩[i]
+private def positionDotMomentum (d : ℕ) := ∑ i : Fin d, 𝐱[i] ∘L 𝐩[i]
```
jstoobysmith, line 96: I wonder if there is a way to define this type of dot product in general, so we wouldn't need a separate definition for each of them.

jstoobysmith (approved):
Approved - will merge shortly

PhysLean/QuantumMechanics/DDimensions/Hydrogen/LaplaceRungeLenzVector.lean
gloges: You mean to subsume not only `positionDotMomentum` but also `∑ i, 𝐩[i] ∘L 𝐩[i]`, `∑ i, ∑ j, 𝐋[i,j] ∘L 𝐋[i,j]`, etc? Yeah, that could be nice - I will think on how to do this.

jstoobysmith: Yeah exactly. If you can get all three with the same definition it would be great, but even just the first two would be nice.

PR #1021, opened 2026-04-03
Add MultiIndex API for local field-theory foundations

juanjfndz:
## Summary

This PR is the first step in a planned local Classical Field Theory development for PhysLib, aimed at a formalization of the local Euler--Lagrange criterion appearing as Theorem 5.2 in Cortés--Haupt, Chapter 5.

- local jets,
- total derivatives,

This first PR only introduces the reusable `MultiIndex` API needed later to index jet coordinates and iterated derivatives.

## What This PR Adds

This PR adds [PhysLean/Mathematics/DataStructures/MultiIndex.lean](/Users/juanjosefernandezmorales/classical-field-theory/physlib/PhysLean/Mathematics/DataStructures/MultiIndex.lean), with a small API for finitely supported multi-indices on `Fin n`.

Concretely, it provides:
- the type `MultiIndex n := Fin n → ℕ`,
- the order `|I| = Σ_i I_i`,
- the zero multi-index,
- increment and add operations,
- basic lemmas about order and coordinates.

## Why This Is Needed

Later PRs will use multi-indices to write the local jet coordinates

`u^a_I`

and iterated derivatives

`∂^I = ∂_1^{I_1} ... ∂_n^{I_n}`.

They also appear in the local Euler--Lagrange expression

`E_a(L) = Σ_{|I| ≤ k} (-D)^I (∂L / ∂u^a_I)`.

So even though this first PR is small and purely reusable, it is the indexing layer that the rest of the local field-theory stack will build on.

## Design Notes

I have placed this in `Mathematics.DataStructures` rather than under `ClassicalFieldTheory`, since it is meant to be reusable outside the field-theory namespace as well. That follows the earlier Zulip suggestion to keep the reusable pieces out of the domain-specific layer where possible.

## Build

Locally checked with:
- `python3 scripts/lint-style.py PhysLean/Mathematics/DataStructures/MultiIndex.lean`
- `lake build PhysLean.Mathematics.DataStructures.MultiIndex`
- `lake build PhysLean`


PhysLean/Mathematics/DataStructures/MultiIndex.lean
```diff
@@ -0,0 +1,113 @@
...
+
+namespace PhysLean
+
+/-!
+## A. The basic type of multi-indices
+
+-/
+
+/-- A multi-index on `n` source coordinates. -/
+abbrev MultiIndex (n : ℕ) := Fin n → ℕ
```
jstoobysmith: I personally would make this an `structure` with a single field corresponding to `Fin n -> Fin d` where `d` is the dimension. Using a `structure` initially will make things more difficult, as you have to define instances on it for e.g. `Add`, but, it gives you much more flexibility. Everytime I haven't used a structure initially I have gone on to regret it.

Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean
```diff
@@ -0,0 +1,113 @@
+/-
```
jstoobysmith, line 1: Would put this in `Space.Derivatives.MultIndex` - we have some plans to eventually change what goes into the `Mathematics` file.

juan-fernandez-gotherlabs:
Thanks — both comments made sense to me.

I have now reworked this locally so that:
- `MultiIndex` is a `structure` (with underlying map `Fin n → ℕ`) rather than an `abbrev`,
- the file lives under `PhysLean/SpaceAndTime/Space/Derivatives/MultiIndex.lean`,
- and the import-order issue from CI is also fixed.

I will push that update to the PR branch next.

Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean
```diff
@@ -0,0 +1,131 @@
...
+
+/-!
+## A. The basic type of multi-indices
+
+-/
+
+/-- A multi-index on `n` source coordinates. -/
+structure MultiIndex (n : ℕ) where
+  /-- The coordinates of the multi-index. -/
+  toFun : Fin n → ℕ
```
jstoobysmith: I think this really should be a map to Fin d not the natural numbers, as the indices only go up to d. Happy to be told otherwise

jstoobysmith: Never mind, I got this wrong. Here `n` is the dimension - so maybe we should rename that `d` to match the convention elsewhere

jstoobysmith:
We renamed ./PhysLean to ./Physlib which is why there is a merge conflict on this PR now. If you could fix it, that would be great, but if you need help feel free to let me know.

Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean
juanjfndz: I was using the standard analysis/PDE convention for multi-indices, namely
`α : Fin d → ℕ`, equivalently `α ∈ ℕ^d`, where each direction `i = 1, ..., d`
is assigned a natural number `α_i` recording how many times one differentiates
in that direction.

Accordingly,
`|α| = Σ_i α_i`
is the total order of differentiation, and
`∂^α = ∂_1^{α_1} ... ∂_d^{α_d}`.

For example, if `d = 3` and `α = (2, 0, 1)`, then
`|α| = 2 + 0 + 1 = 3`
and
`∂^α = ∂_1^2 ∂_3 = ∂_1 ∂_1 ∂_3`.

For example:
- MathWorld: https://mathworld.wolfram.com/Multi-IndexNotation.html

So the intention here was that the directions themselves are indexed by the ambient
space dimension, while the components record derivative multiplicities rather than
coordinate labels.

That said, I do take the point that this should be kept conceptually aligned with
the intended derivative notation, and I will be careful to keep that convention explicit
in the follow-up API so that we do not blur derivative multiplicities with coordinate labels.

juanjfndz:
I have now updated the PR branch to the current `Physlib` tree as well.

Concretely:
- the `MultiIndex` API remains under `Space.Derivatives`,
- the branch is now aligned with the `PhysLean -> Physlib` rename,
- and the local checks now pass again (`lint-style.py`, `lake build Physlib.SpaceAndTime.Space.Derivatives.MultiIndex`, and `lake exe check_file_imports`).

So this one should be back in a sane state now.

Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean
```diff
@@ -0,0 +1,131 @@
...
+deriving DecidableEq
+
+namespace MultiIndex
+
+variable {n : ℕ}
```
jstoobysmith: mentioned above, but could we replace `n` here with `d` since it is meant to be the dimension of `Space`.

juan-fernandez-gotherlabs: Good point — I have now changed this to `d` in the `MultiIndex` API so that it matches the ambient space-dimension convention more closely.

jstoobysmith (approved):
Great! Many thanks! This now looks good to me. I've approved this PR, but you might have to take it out of "draft" mode.

jstoobysmith:
Thanks - I will merge shortly :).

juanjfndz:
Many thanks!

I'll move on to the next one tomorrow:
- PR 2: Iterated derivatives

jstoobysmith:
Great! Looking forward to it!

PR #1014, opened 2026-03-31
feat(QM): Add radius power operator

gloges:
Defines a linear operator which maps a Schwartz function `ψ` to `x ↦ ‖x‖ ^ s * ψ x`.

The resulting function is smooth away from `x = 0` and strongly measurable, and is also square-integrable as long as `s` is not too negative.

PR #1024, opened 2026-04-05
feat(QM): Add radius operator limits

gloges:
Adds lemmas which relate `radiusRegPowOperator` to `radiusPowOperator` as the regulator around `x = 0` is removed.
These are point-wise limits, but when used in conjunction with `radiusPowOperator_apply_memHS` (c.f. #1014) one is able to determine if $(|x|^2 + \epsilon^2)^{s/2} \psi$ remains square-integrable in the strict limit $\epsilon \to 0$.


gloges:
t-quantum-mechanics

gloges:
RFC

jstoobysmith:
We renamed ./PhysLean to ./Physlib which is why there is a merge conflict on this PR now. If you could fix it, that would be great, but if you need help feel free to let me know.

gloges:
Thanks, all merged.

jstoobysmith (approved):
Approved - many thanks. Will merge shortly

PR #1027, opened 2026-04-07
Add iterated Space derivatives indexed by multi-indices

juanjfndz:
## Summary

This PR is the second step in a planned local Classical Field Theory development for PhysLib, aimed at a formalization of the local Euler--Lagrange criterion appearing as Theorem 5.2 in Cortés--Haupt, Chapter 5.

- local jets,
- total derivatives,

The previous PR, #1021, introduced the reusable `MultiIndex` API. This PR extends that API with the canonical ordered list of directions associated to a multi-index, and then uses it to define iterated coordinate derivatives on `Space d`.

## What This PR Adds

This PR adds `Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean`, updates `Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean`, and adds the corresponding import in `Physlib.lean`.

Concretely, it provides:
- `MultiIndex.tail`,
- `MultiIndex.toList`,
- `MultiIndex.length_toList`,
- the definition `Space.iteratedDeriv`,
- the notation `∂^[I] f`,
- and the first basic lemmas identifying the order-zero and one-step cases.

In particular, the API already gives the expected specializations

`∂^[0] f = f`

and

`∂^[increment I 0] f = ∂[0] (∂^[I] f)`

as well as the single-direction case

`∂^[increment 0 i] f = ∂[i] f`.

## Why This Is Needed

The previous PR, #1021, provided the combinatorial indexing layer. This PR supplies both the canonical ordered realization of a multi-index and the corresponding analytic notion of iterated coordinate derivative.

Later PRs will use iterated derivatives to define the local jet evaluation

`j^k_x f`

by collecting the derivative coordinates of a field up to order `k`.

These derivatives are also the local analytic input for the jet coordinates

`u^a_I`

and hence for local lagrangians depending on derivatives up to order `k`.

So this PR is the first analytic bridge between the reusable indexing object from `MultiIndex` and the later local field-theory constructions.

## Scope Of This PR

This PR is intentionally limited to the core definition and its first basic recursion/specialization lemmas.

It does not yet add:
- linearity lemmas,
- smoothness preservation lemmas,
- or support lemmas for iterated derivatives.

Those will appear in the follow-up PR, so that the basic notion of iterated derivative can be reviewed separately before adding the later technical API needed for local jets and first variation.

## Build

Locally checked with:
- `python3 scripts/lint-style.py Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean`
- `lake build Physlib.SpaceAndTime.Space.Derivatives.MultiIndex Physlib.SpaceAndTime.Space.Derivatives.Iterated`
- `lake build Physlib`

### DIFF
--- a/Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean
+++ b/Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean
@@ -20,98 +20,18 @@ of coordinate directions, and the iterated derivative is then defined by repeate
 
 ## ii. Key results
 
-- `Physlib.MultiIndex.toList` : the canonical ordered list of directions encoded by a multi-index.
 - `Space.iteratedDeriv` : iterated coordinate derivatives on `Space d`.
+- `∂^[I] f` : notation for the iterated derivative indexed by the multi-index `I`.
 
 ## iii. Table of contents
 
-- A. The combinatorics of iterated directions
-- B. Iterated derivatives on `Space d`
+- A. Iterated derivatives on `Space d`
 
 ## iv. References
 
 -/
 
 @[expose] public section
-
-namespace Physlib
-
-namespace MultiIndex
-
-variable {d : ℕ}
-
-/-!
-## A. The combinatorics of iterated directions
-
--/
-
-/-- The tail of a multi-index on `d + 1` coordinates, dropping the `0`-th coordinate. -/
-def tail (I : MultiIndex d.succ) : MultiIndex d := ⟨fun i => I i.succ⟩
-
-/-- The canonical ordered list of coordinate directions encoded by a multi-index. -/
-def toList : {d : ℕ} → MultiIndex d → List (Fin d)
-  | 0, _ => []
-  | _ + 1, I => List.replicate (I 0) 0 ++ (toList (tail I)).map Fin.succ
-
-@[simp]
-lemma tail_zero : tail (0 : MultiIndex d.succ) = 0 := by
-  ext i
-  rfl
-
-@[simp]
-lemma tail_increment_zero (I : MultiIndex d.succ) : tail (increment I 0) = tail I := by
-  ext i
-  simp [tail, increment]
-
-@[simp]
-lemma tail_increment_succ (I : MultiIndex d.succ) (i : Fin d) :
-    tail (increment I i.succ) = increment (tail I) i := by
-  ext j
-  by_cases h : j = i
-  · subst h
-    simp [tail, increment]
-  · simp [tail, increment, h]
-
-@[simp]
-lemma toList_zero : toList (0 : MultiIndex d) = [] := by
-  induction d with
-  | zero => rfl
-  | succ d ih =>
-      simp [toList, ih]
-
-lemma length_toList (I : MultiIndex d) : I.toList.length = I.order := by
-  induction d with
-  | zero =>
-      simp [toList, MultiIndex.order]
-  | succ d ih =>
-      simp [toList, tail, MultiIndex.order, Fin.sum_univ_succ, ih]
-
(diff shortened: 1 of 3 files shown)


Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean
```diff
@@ -0,0 +1,149 @@
...
+
+variable {d : ℕ}
+
+/-!
+## A. The combinatorics of iterated directions
+
+-/
+
+/-- The tail of a multi-index on `d + 1` coordinates, dropping the `0`-th coordinate. -/
+def tail (I : MultiIndex d.succ) : MultiIndex d := ⟨fun i => I i.succ⟩
```
jstoobysmith: These results about MultiIndex are probably best placed in that file not this one.

Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean
```diff
@@ -0,0 +1,149 @@
...
+/-!
+## B. Iterated derivatives on `Space d`
+
+-/
+
+/-- The iterated coordinate derivative on `Space d` indexed by a multi-index. -/
+noncomputable def iteratedDeriv [AddCommGroup M] [Module ℝ M] [TopologicalSpace M]
+    (I : MultiIndex d) (f : Space d → M) : Space d → M :=
+  I.toList.foldr (fun i g => deriv i g) f
+
```
jstoobysmith, line 51: I think it would be nice to define some notation for this, so we don't have to use `iteratedDeriv` everywhere.

juanjfndz: That makes sense. Before changing the API, two notation candidates that seem most natural to me here are:

- `∂^[I] f`, to stay close to the standard mathematical notation `∂^I f` while still making the multi-index argument visually explicit in Lean;
- `∂[I] f`, as the lighter-weight option, in continuity with the existing first-order notation `∂[i] f`.

My preference would currently be the first one, but I would be happy to go with either if you think one fits PhysLib style better.

jstoobysmith: I think the first one is good, keeping the brackets might make it easier to define.

juanjfndz:
Thanks — I have now updated this accordingly.

Concretely:
- the purely combinatorial `MultiIndex` material (`tail`, `toList`, and the associated lemmas) has been moved into `Physlib/SpaceAndTime/Space/Derivatives/MultiIndex.lean`,
- and the iterated-derivative notation has been added in the form `∂^[I]`.

Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean
```diff
@@ -0,0 +1,72 @@
...
+@[inherit_doc iteratedDeriv]
+macro "∂^[" I:term "]" : term => `(iteratedDeriv $I)
+
+@[simp]
+lemma iteratedDeriv_zero [AddCommGroup M] [Module ℝ M] [TopologicalSpace M]
+    (f : Space d → M) : iteratedDeriv (0 : MultiIndex d) f = f := by
```
jstoobysmith: could use the notation here aswell.

juanjfndz:
Sorry — this should have been consistent already.

I have now made the notation consistent in `Iterated.lean` as well, and updated the PR description so that it matches the current diff.

jstoobysmith (approved):
Great - approved. Many thanks, will merge shortly

PR #1028, opened 2026-04-08
Extend iterated Space derivatives with regularity and support lemmas

juanjfndz:
## Summary

This PR is the third step in a planned local Classical Field Theory development for PhysLib, aimed at a formalization of the local Euler--Lagrange criterion appearing as Theorem 5.2 in Cortés--Haupt, Chapter 5.

The overall local stack is intended to proceed through:
- multi-indices and iterated derivatives,
- admissible compactly supported variations,
- local jets,
- total derivatives,
- local lagrangians and the action functional,
- the local Euler--Lagrange operator,
- and finally the first-variation criterion.

The Zulip discussion for this development is here:
- https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLib/topic/ClassicalFieldTheory/

The previous two PRs introduced the reusable `MultiIndex` API in #1021 and the core definition of iterated coordinate derivative in #1027. This PR adds the small regularity, linearity, and support lemmas needed to use iterated derivatives later in the local jet and first-variation arguments.

## What This PR Adds

This PR extends `Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean` with:
- `Space.iteratedDeriv_add`,
- `Space.iteratedDeriv_const_smul`,
- `Space.iteratedDeriv_contDiff`,
- `Space.deriv_iteratedDeriv_commute`,
- `Space.tsupport_deriv_subset`,
- `Space.tsupport_iteratedDeriv_subset`,
- and `Space.iteratedDeriv_eq_zero_of_notMem_tsupport`.

These results are stated using the notation `∂^[I] f` introduced in the previous PR.

## Why This Is Needed

The next stage of the development packages iterated coordinate derivatives into local jet data. For that, the iterated-derivative API already needs three basic kinds of control:

- algebraic compatibility, so that jet evaluation behaves well with linear combinations of fields;
- regularity preservation, so that smooth fields still yield smooth derivative coordinates;
- and support control, so that compactly supported variations remain manageable after differentiation.

These are also exactly the ingredients later reused in the first-variation and integration-by-parts arguments.

## Scope Of This PR

This PR is intentionally limited to the reusable analytic API around iterated spatial derivatives.

It does not yet introduce:
- local jets,
- field variations,
- or any field-theory-specific definitions.

Those appear in the following PRs, once this derivative layer is in place.

## Build

Locally checked with:
- `python3 scripts/lint-style.py Physlib/SpaceAndTime/Space/Derivatives/Iterated.lean`
[...]

PR #1033, opened 2026-04-10
Add admissible local variations from test functions

juanjfndz:
## Summary

This PR is the next step in a planned local Classical Field Theory development for PhysLib, aimed at a formalization of the local Euler--Lagrange criterion appearing as Theorem 5.2 in Cortés--Haupt, Chapter 5.

- local jets,
- total derivatives,

The previous PRs, #1021, #1027, and #1028, introduced the reusable `MultiIndex` API and the basic iterated-derivative layer on `Space d`. This PR adds the corresponding notion of admissible compactly supported variation, built directly from `IsTestFunction`.

## What This PR Adds

This PR adds `Physlib/ClassicalFieldTheory/Local/Variation.lean`, together with its import in `Physlib.lean`.

Concretely, it provides:
- the structure `AdmissibleVariation d U` for compactly supported smooth maps `Space d → U`,
- coercion to the underlying function,
- the induced `IsTestFunction` and `HasCompactSupport` facts,
- the associated `CompactlySupportedContinuousMap`,
- and the basic algebraic structure on admissible variations.

It also includes the coordinatewise test-function lemma

`AdmissibleVariation.coord`

in the specialization to codomain `Space m`.

## Why This Is Needed

In the local variational problem, the admissible perturbations are compactly supported smooth maps

`η : Space d → Space m`.

This PR packages the reusable notion first at the level of a general real normed vector-space codomain `U`, and the later local field-theory PRs will specialize it to `U = Space m`.

The next PRs then use these variations to define the varied field

`f + s • η`

and later the local action variation and criticality condition.

## Scope Of This PR

This PR is intentionally limited to the variation object itself and its first basic operations.

It does not yet define:
- varied fields,
- the action functional,
- action variation,
- or criticality.

Those come in the next local-action PR, so that the variation layer can be reviewed first on its own.

## Build

Locally checked with:
- `python3 scripts/lint-style.py Physlib/ClassicalFieldTheory/Local/Variation.lean`
- `lake build Physlib.ClassicalFieldTheory.Local.Variation`

### DIFF
--- a/Physlib/ClassicalFieldTheory/Local/Variation.lean
+++ b/Physlib/ClassicalFieldTheory/Local/Variation.lean
@@ -1,0 +1,136 @@
+/-
+Copyright (c) 2026 Juan Jose Fernandez Morales. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Juan Jose Fernandez Morales
+-/
+module
+
+public import Physlib.Mathematics.VariationalCalculus.IsTestFunction
+/-!
+# Admissible local variations
+
+## i. Overview
+
+This module packages the admissible variations used in the local first-variation problem.
+
+For the first local stage, a variation is admissible if it is a smooth compactly supported map
+`Space n → Space m`. This reuses the existing `IsTestFunction` predicate rather than introducing a
+second support calculus.
+
+## ii. Key results
+
+- `ClassicalFieldTheory.Local.AdmissibleVariation` : compactly supported smooth variations.
+- `ClassicalFieldTheory.Local.IsAdmissibleVariation` : predicate form of the same notion.
+
+## iii. Table of contents
+
+- A. Admissible variations
+- B. Basic operations on admissible variations
+
+## iv. References
+
+-/
+
+@[expose] public section
+
+open MeasureTheory
+open Physlib
+
+namespace ClassicalFieldTheory
+namespace Local
+
+/-!
+## A. Admissible variations
+
+-/
+
+/-- Predicate form of admissible local variations. -/
+abbrev IsAdmissibleVariation (η : Space n → Space m) : Prop := IsTestFunction η
+
+/-- An admissible local variation is a smooth compactly supported map `Space n → Space m`. -/
+structure AdmissibleVariation (n m : ℕ) where
+  toFun : Space n → Space m
+  isAdmissible : IsAdmissibleVariation toFun
+
+namespace AdmissibleVariation
+
+variable {n m : ℕ}
+
+/-!
+## B. Basic operations on admissible variations
+
+-/
+
+instance : CoeFun (AdmissibleVariation n m) (fun _ => Space n → Space m) where
+  coe η := η.toFun
+
+/-- The underlying test-function property of an admissible variation. -/
+lemma isTestFunction (η : AdmissibleVariation n m) : IsTestFunction (η.toFun) := η.isAdmissible
+
+lemma hasCompactSupport (η : AdmissibleVariation n m) : HasCompactSupport (η.toFun) :=
+  η.isTestFunction.supp
+
+noncomputable def toCompactlySupportedContinuousMap (η : AdmissibleVariation n m) :
+    CompactlySupportedContinuousMap (Space n) (Space m) :=
+  η.isTestFunction.toCompactlySupportedContinuousMap
+
+/-- The zero admissible variation. -/
+noncomputable def zero : AdmissibleVariation n m where
+  toFun := fun _ => 0
+  isAdmissible := IsTestFunction.zero
+
+noncomputable instance : Zero (AdmissibleVariation n m) := ⟨zero⟩
+
+@[simp]
+lemma zero_apply (x : Space n) : (0 : AdmissibleVariation n m) x = 0 := rfl
+
+/-- Negation of admissible variations. -/
+noncomputable def neg (η : AdmissibleVariation n m) : AdmissibleVariation n m where
+  toFun := fun x => -η x
+  isAdmissible := η.isTestFunction.neg
+
+noncomputable instance : Neg (AdmissibleVariation n m) := ⟨neg⟩
+
+@[simp]
+lemma neg_apply (η : AdmissibleVariation n m) (x : Space n) : (-η) x = -η x := rfl
+
+/-- Sum of admissible variations. -/
+noncomputable def add (η ξ : AdmissibleVariation n m) : AdmissibleVariation n m where
+  toFun := fun x => η x + ξ x
+  isAdmissible := η.isTestFunction.add ξ.isTestFunction
+
+noncomputable instance : Add (AdmissibleVariation n m) := ⟨add⟩
+
+@[simp]
+lemma add_apply (η ξ : AdmissibleVariation n m) (x : Space n) : (η + ξ) x = η x + ξ x := rfl
+
+/-- Difference of admissible variations. -/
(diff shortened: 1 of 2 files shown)


Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+namespace ClassicalFieldTheory
+namespace Local
+
+/-!
+## A. Admissible variations
+
+-/
+
+/-- Predicate form of admissible local variations. -/
+abbrev IsAdmissibleVariation (η : Space n → Space m) : Prop := IsTestFunction η
```
jstoobysmith: Do we need this abbreviation? seems like the only place it is used is in the definition of `isAdmissible`, and there we could just put `IsTestFunction`

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+
+/-- An admissible local variation is a smooth compactly supported map `Space n → Space m`. -/
+structure AdmissibleVariation (n m : ℕ) where
+  toFun : Space n → Space m
```
jstoobysmith: The definition parts of a structure need doc-strings.

jstoobysmith: Also, I think having `Space m` here as the target is probably not what is generally wanted. Would any normed vector space here work?

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+
+lemma hasCompactSupport (η : AdmissibleVariation n m) : HasCompactSupport (η.toFun) :=
+  η.isTestFunction.supp
+
+noncomputable def toCompactlySupportedContinuousMap (η : AdmissibleVariation n m) :
+    CompactlySupportedContinuousMap (Space n) (Space m) :=
+  η.isTestFunction.toCompactlySupportedContinuousMap
+
+/-- The zero admissible variation. -/
+noncomputable def zero : AdmissibleVariation n m where
```
jstoobysmith: Not sure you need this definition, if you have the instance below as well, could just define things in the instance.

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+  toFun := fun _ => 0
+  isAdmissible := IsTestFunction.zero
+
+noncomputable instance : Zero (AdmissibleVariation n m) := ⟨zero⟩
+
+@[simp]
+lemma zero_apply (x : Space n) : (0 : AdmissibleVariation n m) x = 0 := rfl
+
+/-- Negation of admissible variations. -/
+noncomputable def neg (η : AdmissibleVariation n m) : AdmissibleVariation n m where
```
jstoobysmith: again, don't likely need a definition and an `instance`.

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+  isAdmissible := IsTestFunction.smul_left (f := fun _ : Space n => c) (g := η.toFun)
+    (by fun_prop) η.isTestFunction
+
+noncomputable instance : SMul ℝ (AdmissibleVariation n m) := ⟨smul⟩
+
+@[simp]
+lemma smul_apply (c : ℝ) (η : AdmissibleVariation n m) (x : Space n) :
+    (c • η) x = c • η x := rfl
+
+lemma coord (η : AdmissibleVariation n m) (a : Fin m) :
```
jstoobysmith: attribute with `@[fun_prop]`?

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,136 @@
...
+
+-/
+
+instance : CoeFun (AdmissibleVariation n m) (fun _ => Space n → Space m) where
+  coe η := η.toFun
+
+/-- The underlying test-function property of an admissible variation. -/
+lemma isTestFunction (η : AdmissibleVariation n m) : IsTestFunction (η.toFun) := η.isAdmissible
+
+lemma hasCompactSupport (η : AdmissibleVariation n m) : HasCompactSupport (η.toFun) :=
```
jstoobysmith: attribute above two with `@[fun_prop]` if can.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Yes, agreed. I removed `IsAdmissibleVariation` and now use `IsTestFunction` directly in the structure field.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Done. I added docstrings to the structure fields, and also updated the module/PR description to match the generalized version of the API.

juanjfndz: Yes, this was the right generalization. The structure is now `AdmissibleVariation n U` for a general real normed vector space `U`, and the coordinate lemma is kept only in the specialization to `Space m`.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Agreed. I removed the separate convenience definitions here and now define the algebraic structure directly through the instances.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Done. I added `@[fun_prop]` to the reusable test-function field, and also on the coordinate lemma in the `Space m` specialization.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Same change here: I removed the separate definition and keep only the instance-level construction.

Physlib/ClassicalFieldTheory/Local/Variation.lean
juanjfndz: Done here as well: `@[fun_prop]` is now on the corresponding reusable lemmas.

Physlib/ClassicalFieldTheory/Local/Variation.lean
```diff
@@ -0,0 +1,126 @@
...
+
+/-!
+## A. Admissible variations
+
+-/
+
+/-- An admissible local variation is a smooth compactly supported map `Space n → U`. -/
+structure AdmissibleVariation (n : ℕ) (U : Type*) [NormedAddCommGroup U] [NormedSpace ℝ U] where
+  /-- The underlying variation function. -/
+  toFun : Space n → U
```
jstoobysmith: Sorry, one last comment, an we use ‘d’ throughout instead of ‘n’ again?

juanjfndz: Yes, done. I have now switched this PR to `d` throughout, and I also aligned the same convention in the remaining local stack so that the follow-up PRs stay consistent on this point.

jstoobysmith (approved):
Approved. Many thanks for this.

juan-fernandez-gotherlabs:
> Approved. Many thanks for this.

Many thanks for all the reviews!
Let's continue ^^

PR #1035, opened 2026-04-12
chore: lake shake on physlib

zhikaip:
Ran `lake shake --add-public --keep-implied --keep-prefix --explain Physlib`

shake is not perfect yet and there are various places where it broke the code or I have to run #min_imports manually for comparison. Also can't guarantee that there are no mistakes but maybe this is good enough as a first pass. Hopefully later on we can also integrate shake into CI so we don't have to do this manually anymore.


jstoobysmith (approved):
Great! Many thanks. Approved

PR #1037, opened 2026-04-15
refactor(QM): angular momentum notation

gloges:
Changes the notation for the angular momentum operator (and its square) to mirror that for position / momentum operators and make use of `dotProduct` in the LRL vector definition.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Approved - very nice. Thanks. Will merge shortly

PR #1039, opened 2026-04-15
Add bounded derivative indices

juanjfndz:
## Summary

This PR restarts the jet layer of the local Classical Field Theory development by introducing only the reusable bounded derivative-index object.

- bounded derivative indices,
- coordinate-level jet points,
- fiber/affine jet API,
- jet regularity/support lemmas,
- total derivatives,

The earlier draft of this PR tried to introduce the local jet object directly. Following review feedback, this PR has been reshaped so that it now contains only the reusable derivative infrastructure piece. The jet-point layer will follow in subsequent PRs.

## What This PR Adds

This PR adds `Physlib/SpaceAndTime/Space/Derivatives/DerivativeIndex.lean`, together with its import in `Physlib.lean`.

Concretely, it provides:
- `Physlib.DerivativeIndex d k := { I : MultiIndex d // I.order ≤ k }`,
- `Fintype`, `DecidableEq`, and `Zero` instances,
- and the basic `@[simp]` lemmas for the zero derivative index.

## Why This Is Needed

Later local jet coordinates, finite sums over jet coordinates, and the local Euler--Lagrange formulas all use bounded multi-indices `I` with `|I| ≤ k`.

So this PR isolates the finite reusable indexing object first, before reintroducing the jet-point layer itself.

## Scope Of This PR

This PR is intentionally limited to bounded derivative indices.

It does not yet add:
- jet coordinates,
- jet points,
- affine fiber lines in jet space,
- or jet regularity/support lemmas.

Those will come in the next PRs.

## Build

Locally checked with:
- `python3 scripts/lint-style.py Physlib/SpaceAndTime/Space/Derivatives/DerivativeIndex.lean`
- `lake build Physlib.SpaceAndTime.Space.Derivatives.DerivativeIndex`
- `lake build Physlib`


Physlib/ClassicalFieldTheory/Local/LocalJet.lean
```diff
@@ -0,0 +1,102 @@
...
+/-- The zero multi-index as a derivative index of any order. -/
+def zeroIndex (d k : ℕ) : DerivativeIndex d k := ⟨0, by simp [Physlib.MultiIndex.order_zero]⟩
+
+/-!
+## B. Local jets
+
+-/
+
+/-- Local jets of order `k` for fields `Space d → Space m`. -/
+structure LocalJet (d m k : ℕ) where
```
jstoobysmith: I'm a bit confused about what this definition is meant to represent. More specifically how this is a `local jet`. To me it looks it represents a point in the jet bundle when the base is `Space d` and the fibers are `Space M` (should probably generalize to `U`).

Physlib/ClassicalFieldTheory/Local/LocalJet.lean
```diff
@@ -0,0 +1,102 @@
...
+namespace ClassicalFieldTheory
+namespace Local
+
+/-!
+## A. Derivative indices
+
+-/
+
+/-- The multi-indices on `d` coordinates of order at most `k`. -/
+abbrev DerivativeIndex (d k : ℕ) := { I : MultiIndex d // I.order ≤ k }
```
jstoobysmith: Would make own file in `Space.Derivatives`.

jstoobysmith:
Hey, just wanted to check in about this PR. No rush, just making sure things don't slip through the net.

juan-fernandez-gotherlabs:
It has been a busy week, and I want to look carefully at exactly how to handle the bundle setup. I’m still working toward the same goal. Thanks for the ping!

Physlib/ClassicalFieldTheory/Local/LocalJet.lean
juanjfndz: Sorry for the delay on this — I’ve been fairly tied up, and I also wanted to think more carefully about how this should actually be handled rather than replying too quickly.

I think your comments are pointing to a real issue in how I sliced this layer.

I agree that `DerivativeIndex` should not first appear inside `ClassicalFieldTheory.Local`. It looks like reusable derivative infrastructure, so my current plan is to move it out into its own file under `Space.Derivatives` first.

On the jet side, I think the problem is that what I put in this PR is really closer to a coordinate-level jet point for maps `Space d → Space m` than to something that should just be called `LocalJet` without further qualification. What I was trying to model was a point of the locally trivialized `k`-jet bundle, but I think the current name/API makes that too easy to read in a stronger sense than intended.

So, before I reshuffle the stack, my current thought is:
- first extract `DerivativeIndex` into `Space.Derivatives`,
- then reintroduce the jet layer as a clearer coordinate-level `JetPoint`-style object,
- and only after that add the affine / evaluation / regularity pieces in follow-up PRs.

I am also thinking about whether the bare coordinate container should generalize from `Space m` to a target `U`, while still keeping the later Euler--Lagrange coordinate API specialized to `Space m` for the current theorem.

If that direction sounds right to you, I’ll restructure the stack that way rather than trying to patch this PR in place.

Physlib/ClassicalFieldTheory/Local/LocalJet.lean
juanjfndz: Yes — agreed. I think this should move out first as reusable derivative infrastructure, rather than appearing for the first time in the local jet layer.

Physlib/ClassicalFieldTheory/Local/LocalJet.lean
jstoobysmith: This sounds like a good plan to me.

juanjfndz:
I have now reshaped this PR to contain only the reusable `DerivativeIndex` layer.

The coordinate-level `JetPoint` object, and then the affine / regularity pieces on top of it, will follow in the next PRs.

juanjfndz:
Hi @jstoobysmith — this one has now been reshaped down to just the reusable `DerivativeIndex` layer, following the plan above. If you get a chance to take another look, that would be very helpful; if this cut looks good, I’ll continue with the `JetPoint` core in the next PR.

jstoobysmith (approved):
Approved this - many thanks. Will merge shortly.

PR #1040, opened 2026-04-16
feat(EM): Rewrite deriv as tensorDeriv

jstoobysmith:
Added some lemmas for electromagnetism to rewrite some derivatives as `tensorDeriv`'s. 
Used this to golf some proofs, and setting of things to golf other proofs.


zhikaip (approved):
looks much cleaner, thanks. 

I will still say something like 
```
((Tensor.basis (S := realLorentzTensor d) (Fin.append ![Color.down] ![Color.up])).map
      (Tensorial.toTensor (M := (Lorentz.CoVector d) ⊗[ℝ] (Lorentz.Vector d))).symm) (e.symm (μ, ν))
```
is still quite unreadable, and we probably should think about how to improve these.

ideally we never need to work at this level? perhaps some sort of tactic to normalise tensor expressions? I imagine this is quite difficult but I do wonder how feasible this is.

PR #1041, opened 2026-04-16
chore: Remove unneeded maxHeartbeats

jstoobysmith:
Removing two maxHeartbeats modifications which are actually unneeded.


zhikaip (approved):
this is straightforward so I'll just approve

PR #1043, opened 2026-04-17
refactor: Tensor components

jstoobysmith:
I got rid of some unneeded definitions related to the product of tensor components, and simplified an existing definition. Ends up in shorter cleaner proofs.


morrison-daniel (approved):
Looks good to me

PR #1051, opened 2026-04-21
refactor(QuantumInfo): Some Line lengths

jstoobysmith:
Starting the process of reducing the line lengths of some of the results in QuantumInfo. With the help of the VS-code extension`Rewrap`


rodolfor-s (approved):
Looks good to me.

PR #1055, opened 2026-04-22
feat(QM): add polynomially-bounded Schwartz maps

gloges:
Defines polynomially-bounded Schwartz maps and the corresponding submodule in the Hilbert space.
These are Schwartz maps with the decay bounds `‖x‖ ^ k * ‖f x‖ ≤ C` extended to negative values for `k` (up to some threshold), although phrased in terms of (extended) natural numbers rather than integers.

The proof that these submodules are dense in `SpaceDHilbertSpace` is a bit involved and I have split it off into a separate PR.

### DIFF
--- a/Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
+++ b/Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
@@ -24,6 +24,12 @@ open MeasureTheory
 open InnerProductSpace
 open SchwartzMap
 
+/-!
+## A. Schwartz submodule
+-/
+
+section
+
 variable {d : ℕ}
 
 set_option backward.isDefEq.respectTransparency false in
@@ -49,9 +55,15 @@ set_option backward.isDefEq.respectTransparency false in
 def schwartzEquiv : 𝓢(Space d, ℂ) ≃ₗ[ℂ] schwartzSubmodule d :=
   LinearEquiv.ofInjective schwartzIncl.toLinearMap (injective_toLp (E := Space d) 2)
 
-variable (f g : 𝓢(Space d, ℂ))
+variable (f g : 𝓢(Space d, ℂ)) (ψ : schwartzSubmodule d)
 
 lemma schwartzEquiv_coe_ae : (schwartzEquiv f) =ᵐ[volume] f := coeFn_toLp f 2 volume
+
+lemma schwartzEquiv_symm_coe_ae : schwartzEquiv.symm ψ =ᵐ[volume] ψ := by
+  nth_rw 2 [← schwartzEquiv.apply_symm_apply ψ]
+  exact (schwartzEquiv_coe_ae _).symm
+
+lemma schwartzEquiv_apply_coe : ↑(schwartzEquiv f) = schwartzIncl f := by simp [schwartzEquiv]
 
 lemma schwartzEquiv_inner :
     ⟪schwartzEquiv f, schwartzEquiv g⟫_ℂ = ∫ x : Space d, starRingEnd ℂ (f x) * g x := by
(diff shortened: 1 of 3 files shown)


gloges:
t-quantum-mechanics

Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
```diff
@@ -62,6 +106,98 @@ lemma schwartzEquiv_inner :
 lemma schwartzEquiv_ae_eq (h : schwartzEquiv f =ᵐ[volume] schwartzEquiv g) : f = g :=
   (EmbeddingLike.apply_eq_iff_eq _).mp (SetLike.coe_eq_coe.mp (ext_iff.mpr h))
 
+lemma schwartzIncl_ker : schwartzIncl.ker = (⊥ : Submodule ℂ 𝓢(Space d, ℂ)) := by
+  ext; simp [← schwartzEquiv_apply_coe]
+
+end
+
+/-!
+## B. Polynomially-bounded Schwartz submodule
```
jstoobysmith: Should we put them in their own file?

Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
```diff
@@ -62,6 +106,98 @@ lemma schwartzEquiv_inner :
...
+-/
```
jstoobysmith: May also be good to put a small explanation of why they are needed.

Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
```diff
@@ -62,6 +106,98 @@ lemma schwartzEquiv_inner :
...
+  zero_mem' := fun _ _ ↦ ⟨1, by simp⟩
+  smul_mem' := by
+    intro c f hf k hk
+    obtain ⟨C, hC_pos, hC⟩ := hf k hk
+    refine ⟨(1 + ‖c‖) * C, by positivity, fun x ↦ ?_⟩
+    rw [smul_apply, norm_smul, mul_rotate', mul_comm ‖f x‖]
+    exact le_trans (mul_le_mul_of_nonneg_left (hC x) (norm_nonneg c)) (by linarith)
+
+/-- The linear map `schwartzIncl` with domain restricted to `polyBddSchwartzMap d a`. -/
+def polyBddSchwartzIncl {d : ℕ} {a : ℕ∞} : polyBddSchwartzMap d a →ₗ[ℂ] SpaceDHilbertSpace d :=
```
jstoobysmith: Would it be possible to make these continuous linear  maps? Does it matter here?

gloges: Added! I had not needed it in the density proof, but it was easy enough to make this change.

Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
gloges: I've added to the docs in the new file.

Physlib/QuantumMechanics/DDimensions/SpaceDHilbertSpace/SchwartzSubmodule.lean
gloges: Sure thing - I've separated the file into two (a few residual changes remain in `SchwartzSubmodule.lean`).

jstoobysmith:
Another quick question here: Why define this which Schwartz maps and not more general square integrable functions (which I think might be more common here)?

gloges:
My understanding is that for something like `[p, 1/rˢ]` elements of the domain must be both differentiable and "nice" around the origin, hence bounded Schwartz maps. But perhaps one could instead have polynomially-bounded L2 functions (changing the inequalities to only hold a.e.) and then define `polyBddSchwartzSubmodule = polyBddSubmodule ∩ schwartzSubmodule`.

jstoobysmith:
Ah, ok, this makes sense to me. Thank you. Will approve and merge now.

jstoobysmith (approved):

PR #1059, opened 2026-04-25
refactor: fix import error with Mathlib

jstoobysmith:
Put distributions in the namespace `Physlib` to prevent an error mentioned here 

https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Conflict.20when.20importing.20Mathlib.20and.20Physlib/with/590839433

when importing Physlib and Mathlib at the same time. 

More long term solutions to prevent such things are mentioned there.


Physlib/SpaceAndTime/TimeAndSpace/Basic.lean
```diff
@@ -53,6 +53,7 @@ We put these results in the namespace `Space` by convention.
 -/
 
 @[expose] public section
+open Physlib
```
zhikaip, line 57:
```suggestion
open Physlib

```
The OCD in me wants the new line (or removing it) to be consistent. Is it possible to have linters that check this or is opening namespaces too ad-hoc?

Physlib/SpaceAndTime/Space/Derivatives/Div.lean
```diff
@@ -40,6 +40,7 @@ properties about it.
 -/
 
 @[expose] public section
```
zhikaip, line 42:
```suggestion
@[expose] public section

```
also for this line, I'm being a bit pedantic here but would be nice if consistent. I think the default from modulising has an extra line gap. Just raising it, it's fine if you think it's better if we don't have this line, maybe we can do a pass on the whole of Physlib later

jstoobysmith:
@zhikaip should have addressed these spacing issues here. I agree that there should be a linter for this and general linters around `open`.

zhikaip (approved):
Many thanks, let's go with this for now.

PR #1062, opened 2026-04-28
refactor: TODO item hashes

jstoobysmith:
Replace the TODO item tags with hashes based on String.hash, making it easier for users to write TODO items without having to worry about the unique tag.


zhikaip (approved):
I think this is useful, thanks

PR #1063, opened 2026-04-28
feat: Start split of Distributional EM

jstoobysmith:
Added a Basic.lean file for Distributional EM with the idea that a future PR will move the distribution stuff to this file (I would add a TODO item here, however waiting for #1062). I'm hoping that this doc-string will help clarify what distributional EM is, and how e.g. the integral forms fit in.

### DIFF
--- a/Physlib/Electromagnetism/Distributional/Basic.lean
+++ b/Physlib/Electromagnetism/Distributional/Basic.lean
@@ -76,10 +76,11 @@ through `S` itself is not directly accessible in the framework, since a perfectl
 it can be recovered as a limit of weighted fluxes over progressively sharper test functions.
 
 In this setting, Gauss's law states precisely that the flux with weighting
-`φ` equals the charge measured with the same weighting `φ`. Notionally this is
-`- ∫ E x · ∇ φ x dx = ∫ ρ x φ x dx`.
+`φ` equals the charge measured with the same weighting `φ` (with
+an additional factor of `1/ε₀`). Notionally this is
+`- ∫ E x · ∇ φ x dx = ∫ ρ x φ x dx / ε₀`.
 In the distributional setting, this 'integral' form of Gauss's law
-is exactly equivalent to the differential form `∇ · E = ρ`,
+is exactly equivalent to the differential form `∇ · E = ρ / ε₀`,
 due to the way derivatives are defined by integration by parts.
 
 -/
--- a/Physlib/Electromagnetism/Distributional/Basic.lean
+++ b/Physlib/Electromagnetism/Distributional/Basic.lean
@@ -13,74 +13,89 @@ public import Physlib.Mathematics.VariationalCalculus.HasVarAdjDeriv
 
 ## i. Overview
 
-In electromagnetism, charge and current distributions are often idealised as objects that
-are not smooth functions: point particles, infinitely thin wires, charged surfaces, and so on.
-A point charge, for example, is conventionally written as `ρ(x) = q δ³(x - x₀)`, where `δ³`
-is the Dirac delta. These idealisations are not functions in the usual sense, and cannot be
-described within a formulation of electromagnetism that only allows smooth functions, because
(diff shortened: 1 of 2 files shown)


Physlib/Electromagnetism/Distributional/Basic.lean
```diff
@@ -0,0 +1,86 @@
...
+
+# Distributional Electromagnetic Potential
+
+## i. Overview
+
+In electromagnetism, charge and current distributions are often idealised as objects that
+are not smooth functions: point particles, infinitely thin wires, charged surfaces, and so on.
+A point charge, for example, is conventionally written as `ρ(x) = q δ³(x - x₀)`, where `δ³`
+is the Dirac delta. These idealisations are not functions in the usual sense, and cannot be
+described within a formulation of electromagnetism that only allows smooth functions, because
```
zhikaip: I think "Cannot be described within a formulation of electromagnetism that only allows smooth functions" is too strong, I do want to build smooth version on the punctured space (which links to this), and this shouldn't be the only formulation in Physlib which forces us to always carry distributions around

zhikaip: maybe be a bit explicit, something like "the fields of singular sources can be treated classically on the appropriate punctured domain, but the sources themselves and the differential form of Maxwell's equations including the singularity require distributions."

Physlib/Electromagnetism/Distributional/Basic.lean
```diff
@@ -0,0 +1,86 @@
...
+some smooth sensitivity profile. The idealised pointwise value is then replaced by the
+weighted average `∫ f(x) φ(x) dx`, and a "field" is identified with the rule `φ ↦ ∫ f φ dx`
+that sends each test function to a number.
+
+When the test functions are taken to be `Schwartz maps` (smooth functions which, together
+with all their derivatives, decay faster than any polynomial at infinity), the resulting
+mathematical object is called a tempered distribution. Fields and charge or current
+distributions are then specified by their action on test functions rather than by their
+pointwise values. Ordinary smooth functions `f` still fit into this framework via the rule
+`φ ↦ ∫ f φ dx`, but the framework is strictly larger: it also accommodates point particles,
```
zhikaip: "the framework is strictly larger" is too strong as well, maybe some thing like "the framework accommodates objects that smooth functions cannot represent (point particles, surface charges, derivatives of step functions), at the cost of giving up pointwise values and pointwise products."

Physlib/Electromagnetism/Distributional/Basic.lean
```diff
@@ -0,0 +1,86 @@
...
+even when the sources are singular.
+
+Because pointwise values are no longer available, some constructions from classical
+electromagnetism cease to be well defined in the distributional setting. The Lagrangian
+density `ℒ = - Aᵤ Jᵘ - ¼ Fᵤᵥ Fᵘᵛ` is one such example: it relies on pointwise products of
+distributions (such as `A` with `J`, or `F` with itself), and in general the product of two
+distributions is not defined. For a point charge, for instance, both `A` and `J` are singular
+at the location of the charge, and their product has no distributional meaning.
+
+Other constructions must be reformulated rather than abandoned. The flux of `E` out of a
```
zhikaip: this section may be we can leave it for now, but I'd hope that after things are more fleshed out we can rewrite it to show how the distributional pairing relates to the functional API

zhikaip (commented):
I think the framing should be that Distributions are "necessary", but not necessarily "better". it is needed to handles things properly but we shouldn't rely on it to do everything. (I'll keep looking for a way to do EM stuff the "engineering" or "undergraduate" way and hope we meet somewhere in the middle)

jstoobysmith:
@zhikaip Tried to rewrite to incorporate your comments here and other thoughts.

jstoobysmith:
@zhikaip  Maybe we should also write something explicit here "This is one formalization of Electromagnetism, and there are others which suite other needs. These will also appear in Physlib."

PR #1066, opened 2026-04-30
feat(QuantumInfo): discharge sorries in ClassicalInfo.Capacity and Regularized

dennj:
ClassicalInfo/Capacity: implement BlockCode.encoder/decoder and their
length lemmas via a private mapBlocks helper, closing 4 sorries.

Regularized: prove InfRegularized.to_SupRegularized,
SupRegularized.to_InfRegularized, and InfRegularized.of_Subadditive
via a private limsup_eq_neg_liminf_neg helper, closing 3 sorries.


QuantumInfo/Finite/Capacity.lean
```diff
@@ -122,19 +123,19 @@ end εApproximates
 
 section AchievesRate
 
-/-- Every quantum channel achieves a rate of zero. -/
-theorem achievesRate_0 (Λ : CPTPMap d₁ d₂) : Λ.AchievesRate 0 := by
+/-- Every quantum channel between nonempty spaces achieves a rate of zero. -/
+theorem achievesRate_0 (Λ : CPTPMap d₁ d₂) [Nonempty d₁] [Nonempty d₂] : Λ.AchievesRate 0 := by
```
jstoobysmith: Do you think it is possible here to prove this in its original form? Ie without assuming NonEmpty

QuantumInfo/Regularized.lean
```diff
@@ -120,27 +121,43 @@ end SupRegularized
...
 section real
 
+private def realNegOrderIso : ℝ ≃o ℝᵒᵈ where
+  toEquiv := Equiv.neg ℝ
+  map_rel_iff' := by
+    intro a b
+    change (-b : ℝ) ≤ -a ↔ a ≤ b
+    simpa using neg_le_neg_iff
+
+private theorem limsup_eq_neg_liminf_neg {fn : ℕ → ℝ} {_lb _ub : ℝ}
```
jstoobysmith: Maybe move to where Filter.atTop.limsup are defined or to a ForMathlib

jstoobysmith:
There are some merge conflicts here now, but I think these should be relatively easy to fix.

QuantumInfo/Finite/Capacity.lean
Timeroot: This change is correct, the theorem was missing these assumptions. (There is no CPTPMap from a nonempty space to an empty one, which makes this kind of thing required often.) But only Nonempty d1 is required -- which is what was dennj did in his other PR. So we'll take that one.

QuantumInfo/Regularized.lean
Timeroot: I'm okay keeping this here for now. It should be upstreamed to Mathlib, yeah

Timeroot:
Fixed the merge conflicts. LGTM, will merge

Timeroot (approved):

PR #1067, opened 2026-04-30
feat(CPTPMap): close 6 sorries around piProd

dennj:
MatrixMap:
- exists_kraus: prove via explicit Kraus decomposition over ((B × A) × A) × B (now requires [StarRing R] [Fintype B]).
- Basis.piTensorProduct: drop the local opaque stub and use Mathlib's PiTensorProduct.Basis.
- Add choi_matrix_piProd characterising the Choi matrix of piProd.

Unbundled:
- Add IsTracePreserving.piProd.
- Discharge IsCompletelyPositive.piProd via choi_matrix_piProd and Matrix.PosSemidef.piProd.

CPTP:
- Fill CPTPMap.piProd.TP using IsTracePreserving.piProd (needs maxRecDepth 1000).
- fin_1_piProd: replace the sorry-in-type statement with the real equality using ofEquiv (Equiv.funUnique (Fin 1) _) and prove it.

### DIFF
--- a/QuantumInfo/Finite/CPTPMap/MatrixMap.lean
+++ b/QuantumInfo/Finite/CPTPMap/MatrixMap.lean
@@ -7,6 +7,7 @@ module
 
 public import Mathlib.LinearAlgebra.TensorProduct.Matrix
 public import Mathlib.LinearAlgebra.PiTensorProduct
+public import Mathlib.LinearAlgebra.PiTensorProduct.Basis
 public import Mathlib.Data.Set.Card
 public import Mathlib.Algebra.Module.LinearMap.Basic
 public import QuantumInfo.ForMathlib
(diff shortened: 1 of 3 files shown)


QuantumInfo/Finite/CPTPMap/MatrixMap.lean
```diff
@@ -358,43 +395,42 @@ theorem id_kron_submatrix [CommSemiring R] (f : B → A) :
...
-
 variable {R : Type*} [CommSemiring R]
-variable {ι : Type u} [DecidableEq ι] [fι : Fintype ι]
+variable {ι : Type u} [DecidableEq ι] [Fintype ι]
 variable {dI : ι → Type v} [∀i, Fintype (dI i)] [∀i, DecidableEq (dI i)]
 variable {dO : ι → Type w} [∀i, Fintype (dO i)] [∀i, DecidableEq (dO i)]
 
 /-- Finite Pi-type tensor product of MatrixMaps. Defined as `PiTensorProduct.tprod` of the
   underlying Linear maps. Notation `⨂ₜₘ[R] i, f i`, eventually. -/
 noncomputable def piProd (Λi : ∀ i, MatrixMap (dI i) (dO i) R) : MatrixMap (∀i, dI i) (∀i, dO i) R :=
```
jstoobysmith, line 405: Is this definition equivalent to the existing one? What is the reason for the change?

QuantumInfo/Finite/CPTPMap/Unbundled.lean
jstoobysmith, line 1: At some point, I think these CPTP files should be broken up. Maybe we should add a TODO item for this at some point.

jstoobysmith (commented):
Two comments - though one is really just a comment, the other a question.

jstoobysmith (approved):
Ok, I'm approving this one, as it seems like it is undeniably an improvement,  I also now understand the change in the definition of piProd

PR #1068, opened 2026-05-01
feat: Basic field strength equalities

jstoobysmith:
Adding some lemmas which rewrite the field strength in a more familiar form. These will eventually be used to remove all the `Tensor` stuff from EM.

### DIFF
--- a/Physlib/Electromagnetism/Kinematics/FieldStrength.lean
+++ b/Physlib/Electromagnetism/Kinematics/FieldStrength.lean
@@ -78,12 +78,12 @@ TODO "Currently the API for the field strength tensor has the definition
 
 ## A. The field strength tensor
 
-We define the field strength tensor `F_μ^ν` in terms of the derivative of the
+We define the field strength tensor `F^{μν}` in terms of the derivative of the
 electromagnetic potential `A^μ`. We then prove that this tensor transforms correctly
 under Lorentz transformations.
 
 -/
-/-- The field strength from an electromagnetic potential, as a tensor `F_μ^ν`. -/
+/-- The field strength from an electromagnetic potential, as a tensor `F^{μν}`. -/
 noncomputable def toFieldStrength {d} (A : ElectromagneticPotential d) :
     SpaceTime d → Lorentz.Vector d ⊗[ℝ] Lorentz.Vector d := fun x =>
   Tensorial.toTensor.symm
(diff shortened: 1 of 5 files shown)


Physlib/Electromagnetism/Kinematics/FieldStrength.lean
```diff
@@ -120,10 +145,149 @@ lemma toTensor_toFieldStrength {d} (A : ElectromagneticPotential d) (x : SpaceTi
...
+
+These equalities for the field strength tensor are in terms of vector basis.
+They match some of the familiar forms one might expect to see the field strength
+tensor in.
+
+-/
+
+/-- The statement that `F = F^{μν} eᵤ ⊗ eᵥ` written explicitly, with
+  the components extracted via `toField`. -/
+lemma toFieldStrength_eq_sum_basis_eval {d} {A : ElectromagneticPotential d} :
```
zhikaip, line 164: does this work for a general `SpaceTime d → Lorentz.Vector d ⊗[ℝ] Lorentz.Vector d` type?

To me it seems a bit uneconomic to have these results proven specifically for one physical object only where presumably it's just mathematical manipulations, but maybe there are difficulties I'm not aware of

Physlib/Electromagnetism/Kinematics/FieldStrength.lean
zhikaip:
```suggestion
We define the field strength tensor `F_μν` in terms of the derivative of the
```

do you mind fixing the indices throughout?

Physlib/Electromagnetism/Kinematics/FieldStrength.lean
zhikaip: maybe this is a more general discussion and out of scope for this particular PR, if that's the case I'm happy to approve it first

zhikaip (commented):
I think the docstring should be fixed, whether the mathematical properties can be generalised can be left for future work

Physlib/Electromagnetism/Kinematics/FieldStrength.lean
jstoobysmith: I added a TODO item about this. I tried to do this but it needs a bit more API I think.

Physlib/Electromagnetism/Kinematics/FieldStrength.lean
jstoobysmith: Done.

zhikaip (approved):
thanks for the changes, approved. 

I've been thinking whether `FieldStrength` should get its own type, but maybe that's a much larger design decision to be made. Either way I think separating the maths from the physics as much as possible (generalising the former, then applying) is probably the way to go throughout physlib

PR #1074, opened 2026-05-02
feat(HermitianMat/Proj): add projector positivity, kernel, and posPart=0 lemmas

dennj:
proof wanted to theorem

### DIFF
--- a/QuantumInfo/ForMathlib/HermitianMat/Proj.lean
+++ b/QuantumInfo/ForMathlib/HermitianMat/Proj.lean
@@ -8,6 +8,7 @@ module
 public import QuantumInfo.ForMathlib.HermitianMat.CFC
 
 public import Mathlib.Analysis.CStarAlgebra.Classes
+public import Mathlib.Analysis.InnerProductSpace.Positive
 public import Mathlib.Analysis.SpecialFunctions.ContinuousFunctionalCalculus.PosPart.Basic
 
 /-!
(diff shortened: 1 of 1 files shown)


QuantumInfo/ForMathlib/HermitianMat/Proj.lean
```diff
@@ -8,6 +8,7 @@ module
 public import QuantumInfo.ForMathlib.HermitianMat.CFC
```
jstoobysmith, line 9: I think this file should be named “Projector”. If easy could we change it? No worries if not.

QuantumInfo/ForMathlib/HermitianMat/Proj.lean
```diff
@@ -56,6 +57,26 @@ theorem projector_add_orthogonal : projector S + projector Sᗮ = 1 := by
   erw [ Subtype.mk_eq_mk ];
   ext i j; simp [ LinearMap.toMatrix_apply, Matrix.one_apply ] ;
 
+theorem projector_nonneg : 0 ≤ projector S := by
```
jstoobysmith, line 60: Most of the QuantumInfo “theorem” will eventually be changed to “lemma” could we use “lemma” here and throughout the changes

jstoobysmith (commented):
Couple of small comments - otherwise this looks good

jstoobysmith (approved):
Let me actually approve this one, as my comments can be left to a future PR. I think this is undeniably an improvement, and therefore should be merged. I will do so shortly.

PR #1075, opened 2026-05-03
feat(HermitianMat/Rpow): prove Lieb-Thirring and trace subadditivity

dennj:


Timeroot:
This looks great. The compound matrix theory is already somewhat developed at QuantumInfo/ForMathlib/Majorization, I think for essentially the same purposes. It would be preferable to use / extend that, I think it'll have a lot of the common lemmas then.

dennj:
-awaiting-author

I did the refactor

Timeroot (approved):
Looks good, thanks!

PR #1079, opened 2026-05-04
feat(ClassicalMechanics/HarmonicOsciallator): a few refactors and a new lemma

Bergschaf:
Hi,
I am an undergraduate physics student with a bit of lean experience and I just came across physlib. I was looking around the files and found a few things that could be expressed a bit cleaner (namely using @[ext] annotations). 
I also added a new lemma about the periodicity of the trajectory of a harmonic oscillator. For this, i introduced the definition of the period T of a harmonic oscillator. Please give me some feedback if this is desirable here or if there are better design choices.

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
@@ -269,7 +269,6 @@ lemma trajectory_zero : trajectory S 0 = fun _ => 0 := by
 
 /-!
 
-
 ### B.3. Smoothness of the trajectories
 
 The trajectories for any initial conditions are smooth functions of time.
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
@@ -25,7 +25,6 @@ prove that they satisfy the equation of motion, and prove some properties of the
 
 - A. The initial conditions
   - A.1. Definition of the initial conditions
-    - A.1.1. Extensionality lemma
   - A.2. Relation to other types of initial conditions
   - A.3. The zero initial conditions
     - A.3.1. Simple results for the zero initial conditions
@@ -80,29 +79,16 @@ We start by defining the type of initial conditions for the harmonic oscillator.
 -/
 
 /-- The initial conditions for the harmonic oscillator specified by an initial position,
-  and an initial velocity. -/
-structure InitialConditions where
+  and an initial velocity.
+
+The `@[ext]` attribute provides an extensionality lemma for `InitialConditions`.
+That is, a lemma which states that two initial conditions are equal if their
+initial positions and initial velocities are equal. -/
+@[ext] structure InitialConditions where
   /-- The initial position of the harmonic oscillator. -/
   x₀ : EuclideanSpace ℝ (Fin 1)
   /-- The initial velocity of the harmonic oscillator. -/
   v₀ : EuclideanSpace ℝ (Fin 1)
-
-/-!
-
-#### A.1.1. Extensionality lemma
-
-We prove an extensionality lemma for `InitialConditions`.
-That is, a lemma which states that two initial conditions are equal if their
-initial positions and initial velocities are equal.
-
--/
-
-@[ext]
-lemma InitialConditions.ext {IC₁ IC₂ : InitialConditions} (h1 : IC₁.x₀ = IC₂.x₀)
-    (h2 : IC₁.v₀ = IC₂.v₀) : IC₁ = IC₂ := by
-  cases IC₁
-  cases IC₂
-  simp_all
 
 /-!
 
@@ -147,7 +133,7 @@ This is useful when the natural reference point for a problem is not at time zer
 
   The conditions can be converted to the standard `InitialConditions` format (at `t=0`)
   using the `toInitialConditions` function. -/
-structure InitialConditionsAtTime where
+@[ext] structure InitialConditionsAtTime where
   /-- The time at which the initial conditions are specified. -/
   t₀ : Time
   /-- The position at time t₀. -/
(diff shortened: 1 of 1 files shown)


Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,51 @@ We give some open TODOs for the classical harmonic oscillator.
 
 -/
 
+end InitialConditions
+
+/--
+The period of a Harmonic Osciallator is `2 * π / ω`.
+-/
+noncomputable def T (S : HarmonicOscillator) : ℝ := 2 * π / S.ω
```
jstoobysmith: I would rename this to `period`, and then definition notation that lets you use `T`. This hasn't been done elsewhere in this module, but it likely should be done at some point.

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,51 @@ We give some open TODOs for the classical harmonic oscillator.
...
+
+lemma T.def : S.T = 2 * π / S.ω := rfl
+
+lemma T.pos : 0 < S.T := by
```
jstoobysmith: `period_pos`

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,51 @@ We give some open TODOs for the classical harmonic oscillator.
...
 -/
 
+end InitialConditions
+
+/--
+The period of a Harmonic Osciallator is `2 * π / ω`.
+-/
+noncomputable def T (S : HarmonicOscillator) : ℝ := 2 * π / S.ω
+
+lemma T.def : S.T = 2 * π / S.ω := rfl
```
jstoobysmith: `period_eq`

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,51 @@ We give some open TODOs for the classical harmonic oscillator.
...
+
+/--
+The trajectory of the harmonic oscialltor is periodic with period of `2 * π / ω`.
+-/
+lemma trajectory_periodic (IC : InitialConditions) :
+    Function.Periodic (IC.trajectory S) S.T := by
+  simp only [Function.Periodic, InitialConditions.trajectory, add_val]
+  intro t
+  have h1 : Real.cos (S.ω * (t.val + S.T)) = Real.cos (S.ω * t) :=
+    calc
```
jstoobysmith: I suspect this can be golfed to be made shorter, does `grind` work here? No worries if you can't get it shorter though.

jstoobysmith (commented):
Couple of comments, but looks like a very nice PR :).

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
Bergschaf: Thanks for your quick review! grind does not work here, the anoying thing about this proof is, that omega and 1/omega have to be next to each other for the lemma to work, which is why I am using calc. However, I managed to get down to just one `have`

jstoobysmith: Does `field_simp` work then?

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,42 @@ We give some open TODOs for the classical harmonic oscillator.
 
 -/
 
+end InitialConditions
+
+/--
+The period of a Harmonic Osciallator is `2 * π / ω`.
```
jstoobysmith: This misspelling of Oscillator also will need fixing for the linters.

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
jstoobysmith: Will need documentation I think.

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -880,14 +849,42 @@ We give some open TODOs for the classical harmonic oscillator.
...
+-/
+noncomputable def period (S : HarmonicOscillator) : ℝ := 2 * π / S.ω
+
+noncomputable abbrev T := HarmonicOscillator.period
```
jstoobysmith: scoped notation "T" => HarmonicOscillator.period 

probably safer here.

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
Bergschaf: Oh, yes, that works. It needs to know that omega is not equal to zero though.

Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
```diff
@@ -872,6 +841,33 @@ lemma trajectory_velocity_eq_zero_iff (IC : InitialConditions) (t : Time) :
...
     apply ω_ne_zero
     exact Ne.symm (Nat.zero_ne_add_one 1)
+end InitialConditions
+
+/--
+The period of a harmonic oscillator is `2 * π / ω`.
+-/
+noncomputable def period (S : HarmonicOscillator) : ℝ := 2 * π / S.ω
+
+scoped notation "T" => HarmonicOscillator.period
```
jstoobysmith, line 852: One last thing: I think you are going to need `@[inherit_doc period]` i.e. 

```suggestion
@[inherit_doc period]
scoped notation "T" => HarmonicOscillator.period
```

jstoobysmith (approved):
Approved - many thanks and congrats on your first PR :). Will merge shortly.

PR #1080, opened 2026-05-05
feat: Differentiability of the field strength

jstoobysmith:
We add some basic API about the differentiability of the field-strength tensor. This is then used to golf some down-stream proofs.


zhikaip (approved):
Lgtm

PR #1082, opened 2026-05-05
feat(QuantumInfo): Finish Data Processing Inequality

Timeroot:
This adds the formalization of the data processing inequality for the sandwiched relative Renyi entropy for quantum states. This was the final sorry needed to finish the proof of the Generalized Quantum Stein's Lemma. The quantum-specific stuff is in DPI.lean, but then it depends on several other operator inequalities.

Some of the operator inequalities are proved fresh in QuantumInfo, but some come from a separate project of Hayata Yamasaki's group. They're wanting to eventually publish their project separately (before, I believe, eventually upstreaming) -- but they've agreed to allow it here in the interim. (I think this is so that, primarily, they can get proper academic credit when they have their project done.) In order to make the credit here abundantly clear, their 10 files are in a dedicated folder `QuantumInfo/ForMathlib/HayataGroup`. It has been edited lightly to (1) move it to the module system, (2) remove long lines, and (3) port to the current version of Mathlib.


jstoobysmith (approved):
Approved - feel free to merge when you are ready.

PR #1085, opened 2026-05-06
docs: Add documentation for Tensors

jstoobysmith:
Add a module documentation string for `Tensors` giving the overall picture of things.


morrison-daniel (approved):
I think this is good information, if you want I think giving an example to go along with the description would be helpful. But I'm fine with merging as is.

jstoobysmith:
@morrison-daniel Will merge as it is, and will keep a note to come back and add an example. Many thanks.

PR #1088, opened 2026-05-08
feat(QM): Add multiplication unbounded operators

gloges:
Introduces multiplication unbounded operators on `SpaceDHilbertSpace`.
A function `f : Space d → ℂ` gives an operator mapping `ψ ↦ f • ψ` and with domain exactly those `ψ` for which `f • ψ` is an element of the Hilbert space. A measurability assumption on `f` ensures the domain is dense.

This can be used, for example, to define the potential operator `ψ ↦ V • ψ` for a quantum system on `Space d`.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good to me! Will merge shortly. Thanks.

PR #1090, opened 2026-05-12
refactor(QM): unbounded operators

gloges:
Introduces `IsUnbounded` as a property for LinearPMaps: see [#Physlib > Rethinking unbounded operators](https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/Rethinking.20unbounded.20operators/with/593923945).

_Unbounded_v2.lean_ is a reworking of _Unbounded.lean_ to have unbounded operators (characterized by `IsUnbounded`) on par with `HasDenseDomain`, `IsClosable`, `IsSymmetric` and `IsSelfAdjoint`.

As a trial, I have switched multiplication operators over to the new system.


gloges:
t-quantum-mechanics

gloges:
RFC

morrison-daniel:
I'll take a closer look later, but how has this compared to the other version in your opinion? It looks like it hasn't made that much of a difference so far but I like that the multiplication operator can be written as `ℳ f` instead of `ℳ f hf` and having to carry the measurability condition everywhere with the function, it seems closer to what you would usually see.

gloges:
Thanks! I prefer the new version for how the `IsUnbounded` lemmas mirror those for `IsSymmetric` &c. With some `@[simp]` and `@[aesop]` tags the 'obvious' manipulations should be handled easily.
The `mulOperator` lemmas are essentially just a repackaging of what was there, but dropping the `hf` in `ℳ f hf` is a very nice bonus and makes it [analogous](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Distribution/SchwartzSpace/Basic.html#SchwartzMap.smulLeftCLM_apply) to `SchwartzMap.smulLeftCLM` and `HasTemperateGrowth`.

gloges:
At this point everything has been transferred and the old `UnboundedOperator` file removed.

Physlib/QuantumMechanics/DDimensions/Operators/Position.lean
```diff
@@ -128,46 +128,52 @@ lemma normRegularizedPow_hasTemperateGrowth (d : ℕ) (ε : ℝˣ) (s : ℝ) :
...
 @[simp]
-lemma radiusRegPowOperator_zero {d : ℕ} (ε : ℝˣ) :
+lemma radiusRegPowCLM_zero {d : ℕ} (ε : ℝˣ) :
     𝐫₀ ε 0 = ContinuousLinearMap.id ℂ 𝓢(Space d, ℂ) := by
   ext
   simp
 
 set_option backward.isDefEq.respectTransparency false in
-lemma positionOperatorSqr_eq {d : ℕ} (ε : ℝˣ) :
+lemma positionCLMSqr_eq {d : ℕ} (ε : ℝˣ) :
```
morrison-daniel:
```suggestion
lemma positionSqCLM_eq {d : ℕ} (ε : ℝˣ) :
```
I think we should align with Mathlib notation that "sq" is the abbreviation of "square". Also, it seems the convention you're starting is that the end of name is the kind of mapping it is so "CLM" goes last. For example, above you have "momentumSqOperator".

Physlib/QuantumMechanics/DDimensions/Operators/Momentum.lean
```diff
@@ -82,43 +82,64 @@ lemma momentumOperator_apply (ψ : 𝓢(Space d, ℂ)) (x : Space d) : 𝐩 i ψ
...
-  ofSymmetric' (SchwartzSubmodule.dense d) (momentumOperatorSchwartz_isSymmetric i)
+  exact integral_mul_fderiv_eq_neg_fderiv_mul_of_integrable hI₂ hI₃ hI₄ (by fun_prop) (by fun_prop)
+
+lemma momentumOperator_isUnbounded : (𝐏 i).IsUnbounded := by
+  refine (LinearPMap.IsSymmetric.isUnbounded_iff_hasDenseDomain ?_).mpr ?_
+  · exact momentumOperator_isSymmetric i
+  · exact momentumOperator_hasDenseDomain i
+
+/-- The square of the momentum operator. -/
+def momentumSqrOperator : SpaceDHilbertSpace d →ₗ.[ℂ] SpaceDHilbertSpace d :=
```
morrison-daniel:
```suggestion
def momentumSqOperator : SpaceDHilbertSpace d →ₗ.[ℂ] SpaceDHilbertSpace d :=
```
See below

Physlib/QuantumMechanics/DDimensions/Operators/Unbounded.lean
```diff
@@ -55,557 +36,389 @@ Definitions:
...
+    ← adjoint_graph_eq_graph_adjoint h.1, mem_submodule_adjoint_iff_mem_submoduleToLp_orthogonal]
+  rintro ⟨y, Uy⟩ hy
+  simp only [neg_zero, WithLp.prod_inner_apply, inner_zero_right, add_zero]
+  exact hx y (mem_domain_of_mem_graph hy)
 
-/-!
-## H. Generalized eigenvectors
--/
+@[simp]
+lemma IsUnbounded.closure_adjoint_eq_adjoint [CompleteSpace H] (h : U.IsUnbounded) :
```
morrison-daniel:
```suggestion
lemma IsUnbounded.adjoint_closure_eq_adjoint [CompleteSpace H] (h : U.IsUnbounded) :
```

morrison-daniel (commented):
Minor naming convention suggestions

morrison-daniel (approved):

PR #1100, opened 2026-05-15
feat: TensorSpecies.ofFunction

jstoobysmith:
Add `TensorSpecies.ofFunction`. This is the first step in de-categorifying the Tensor notions in Physlib. It gives an illustration of what should be done.


morrison-daniel (approved):
Looks good, didn't worry about the proofs since we intend to make this the definition

PR #1101, opened 2026-05-15
feat: Add EM Potential constructors

jstoobysmith:
Adding constructors of the electromagnetic potential for the function case. Giving ways to generate an element of `EMPotential` from scalar and vector potentials.

### DIFF
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -164,7 +164,7 @@ noncomputable def ofPotentials {d} (c : SpeedOfLight) (ϕ : Time → Space d →
     ElectromagneticPotential d :=
   ofScalarPotential c ϕ + ofVectorPotential c A
 
-/-- The creation of of an electromagnetic potential from  static potentials. -/
+/-- The creation of of an electromagnetic potential from static potentials. -/
 noncomputable def ofStaticPotentials {d} (c : SpeedOfLight) (ϕ : Space d → ℝ)
     (A : Space d → EuclideanSpace ℝ (Fin d)) : ElectromagneticPotential d :=
   ofStaticScalarPotential c ϕ + ofStaticVectorPotential c A
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -135,7 +135,7 @@ noncomputable def ofScalarPotential {d} (c : SpeedOfLight)
     (ϕ : Time → Space d → ℝ) : ElectromagneticPotential d where
   val x μ :=
     match μ with
-    | Sum.inl 0 => (timeSlice c).symm  ϕ x
+    | Sum.inl 0 => (timeSlice c).symm ϕ x
     | Sum.inr _ => 0
 
 /-- The creation of an electromagnetic potential from a static scalar potential. -/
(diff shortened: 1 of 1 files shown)


Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -121,7 +125,78 @@ lemma smul_apply {d} (r : ℝ) (A : ElectromagneticPotential d) (x : SpaceTime d
...
+    (A : Space d → EuclideanSpace ℝ (Fin d)) : ElectromagneticPotential d :=
+  ofVectorPotential c (fun _ => A)
+
+/-- The creation of an electromagnetic potential from the non-relativistic potentials. -/
+noncomputable def ofPotentials {d} (c : SpeedOfLight) (ϕ : Time → Space d → ℝ)
+    (A : Time → Space d → EuclideanSpace ℝ (Fin d)) :
+    ElectromagneticPotential d :=
+  ofScalarPotential c ϕ + ofVectorPotential c A
+
+/-- The creation of of an electromagnetic potential from  static potentials. -/
```
zhikaip:
```suggestion
/-- The creation of of an electromagnetic potential from static potentials. -/
```
extra space

Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -121,7 +125,78 @@ lemma smul_apply {d} (r : ℝ) (A : ElectromagneticPotential d) (x : SpaceTime d
...
+/-- The creation of an electromagnetic potential from a static vector potential. -/
+noncomputable def ofStaticVectorPotential {d} (c : SpeedOfLight)
+    (A : Space d → EuclideanSpace ℝ (Fin d)) : ElectromagneticPotential d :=
+  ofVectorPotential c (fun _ => A)
+
+/-- The creation of an electromagnetic potential from the non-relativistic potentials. -/
+noncomputable def ofPotentials {d} (c : SpeedOfLight) (ϕ : Time → Space d → ℝ)
+    (A : Time → Space d → EuclideanSpace ℝ (Fin d)) :
+    ElectromagneticPotential d :=
+  ofScalarPotential c ϕ + ofVectorPotential c A
```
zhikaip: is the addition better or just match the whole thing?

If someone wants to use this for a concrete example, this will require an extra step to add the two term together

Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -121,7 +125,78 @@ lemma smul_apply {d} (r : ℝ) (A : ElectromagneticPotential d) (x : SpaceTime d
...
+noncomputable def ofStaticPotentials {d} (c : SpeedOfLight) (ϕ : Space d → ℝ)
+    (A : Space d → EuclideanSpace ℝ (Fin d)) : ElectromagneticPotential d :=
+  ofStaticScalarPotential c ϕ + ofStaticVectorPotential c A
+
+/-- The electromagnetic potential from a static electric and a static magnetic field.
+  There is no canonical choice here, so this depends on choice. -/
+noncomputable def ofStaticElectricMagneticField (c : SpeedOfLight)
+    (E : Space 3 → EuclideanSpace ℝ (Fin 3))
+    (B : Space 3 → EuclideanSpace ℝ (Fin 3))
+    (hE : Differentiable ℝ E) (hB : ContDiff ℝ 1 B)
```
zhikaip, line 180: why are these two `(hE : Differentiable ℝ E) (hB : ContDiff ℝ 1 B)` not both just `Differentiable`?

(I did try to experiment a bit, and I found there's a stack overflow bug in `continuous_uncurry_fderiv_homotopyOperatorIntegrand` in `Curl.lean` when calling `fun_prop` if I change it to `(Differentiable ℝ f)`, so I'm guessing it's this? I can't find any discussion on zulip, if you're familiar with the proof maybe you can produce a mwe and raise it?)

jstoobysmith: I don't think this is an issue with `fun_prop`. I think the problem is that in `continuous_uncurry_fderiv_homotopyOperatorIntegrand` you are showing that the derivative of the homotopy operator is continuous, and for that you actually need `ContDiff ℝ 1 B` (differentiable with a continuous derivative), not just `Differentiable ℝ E`. I'm not sure if loosening `ContDiff ℝ 1 B` here is possible.

Physlib/Electromagnetism/Kinematics/EMPotential.lean
jstoobysmith: Yeah, I think matching is better - it will have better properties with e.g. `rfl`. I've updated this.

zhikaip (approved):
maybe static could have a separate API in the future, but we can update later. many thanks for this

PR #1103, opened 2026-05-16
Prove Lorentz algebra generator lemmas

NicolaBernini:
# Overview 

Prove Lorentz algebra generator lemmas


jstoobysmith (approved):
Looks good to me - approved. Will merge shortly

PR #1105, opened 2026-05-16
feat: Adding a units example involving derivatives

NicolaBernini:
# Overview

Adding a units example involving derivatives


jstoobysmith (approved):
Looks good, many thanks - will merge soon.

PR #1111, opened 2026-05-20
feat: Other implementation

jstoobysmith:
Adding the ability to link to other implementations within Physlib.


morrison-daniel:
Can you explain why this is better than a link in a doc-string? Would we pull these out for the website or something similar?

PR #1112, opened 2026-05-20
feat(FluidDynamics): add Navier-Stokes conservative and convective forms

FloWsnr:
## Summary

This PR introduces a first Navier-Stokes API for `Physlib.FluidDynamics`.
The Navier-Stokes form is intentionally kept as the general balance-law without yet specializing to a Newtonian constitutive law, viscosity assumptions, or boundary conditions.

## Changes
- Add `Space.tensorDiv` for matrix-valued spatial fields, with basic API lemmas for evaluation, zero, constants, addition, and scalar multiplication. This is needed for the general balance-law. We use the already present vector div `Physlib.SpaceAndTime.Space.Derivatives.Div` inside the tensor div to be maximally compatible.
- Add `Physlib.FluidDynamics.NavierStokes.Basic` with typed aliases for mass density, velocity, stress tensor, and body-force fields.
- Define the continuity equation, conservative momentum equation, convective momentum equation.
- Add momentum-density, momentum-flux, convective-term, and material-acceleration helpers.
- Prove the algebraic bridge between conservative and convective momentum forms, including equivalence theorems under the continuity equation and differentiability assumptions.
- Export the new modules from `Physlib.lean`.

## Notes

Feel free to disagree with the general direction


jstoobysmith:
I would take a look at #949 here, which is in a similar direction. I think the structure for an `IdealFluid` used there is the right one which should be followed.

FloWsnr:
Hi Joseph! I'm sorry, I missed that PR.. I will take a look and adjust. I'll ping you here once the code is ready for review again.

FloWsnr:
@jstoobysmith I checked out #949 . The PRs are very much compatible.
I am trying to introduce a more general formulation of fluids and therefore suggest to start with a minimal FluidState:

```lean
structure FluidState (d : ℕ) where
rho : MassDensity d
velocity : VelocityField d
```

we can then extend this structure when I work with more specific physics, e.g.

```lean
structure MomentumBalanceData (d : ℕ) extends FluidState d where
stress : StressTensor d
bodyForce : BodyForce d

structure ThermodynamicFluidState (d : ℕ) extends FluidState d where
pressure : Time → Space d → ℝ
entropy : Time → Space d → ℝ
enthalpy : Time → Space d → ℝ
```

This way, the basic mass and momentum equations do not need to carry thermodynamic fields such as entropy or enthalpy when they are not used. Also, enthalpy is often derived from other thermodynamic data together with an
equation of state, rather than being independent primitive data. If we put it in the base structure now, we would
either add unused data to many definitions or later need compatibility assumptions saying that the enthalpy
agrees with pressure, density, internal energy, etc.

Let me know what you think!

jstoobysmith:
Yep - I think this sounds good. I think having `FluidState` as the core data structure is nice.

FloWsnr:
> Yep - I think this sounds good. I think having `FluidState` as the core data structure is nice.

Okay, I did just that. Do you want

1. to merge this in this general form to keep the PR small and I then add new PRs later?
2. or should I add more flesh, i.e. some of the stuff from the other PR to see how this all structure and abstraction behaves?

Physlib/FluidDynamics/NavierStokes/Basic.lean
```diff
@@ -0,0 +1,412 @@
...
+abbrev BodyForce (d : ℕ) := Time → Space d → EuclideanSpace ℝ (Fin d)
+
+/-!
+
+## B. Fluid and momentum-balance data
+
+-/
+
+/-- The density and velocity fields of a fluid on `d`-dimensional space. -/
+structure FluidState (d : ℕ) where
```
jstoobysmith: Can we move this to a file called `FluidState` or similar.

Physlib/FluidDynamics/NavierStokes/Basic.lean
```diff
@@ -0,0 +1,412 @@
...
+  /-- The mass density field. -/
+  rho : MassDensity d
+  /-- The velocity field. -/
+  velocity : VelocityField d
+
+/-- The data needed for a momentum balance: fluid state, stress, and body force. -/
+structure MomentumBalanceData (d : ℕ) extends FluidState d where
```
jstoobysmith: Not so sure about this name (although my knowledge of fluids is a decade old).

Physlib/FluidDynamics/NavierStokes/Basic.lean
```diff
@@ -0,0 +1,412 @@
...
+  bodyForce : BodyForce d
+
+/-!
+
+## C. Momentum fields
+
+-/
+
+/-- The momentum density `rho u`. -/
+def momentumDensity (d : ℕ) (fluid : FluidState d) : VelocityField d :=
```
jstoobysmith: `momentumDensity` should probably not have type `VelocityField`.

Physlib/SpaceAndTime/Space/Derivatives/MatrixDiv.lean
jstoobysmith, line 1: I think `MatrixDiv` might be a better name for this (`tensorDiv` would mean something else in Physlib). We have e.g. `tensorDeriv`.

Physlib/FluidDynamics/NavierStokes/Basic.lean
```diff
@@ -0,0 +1,412 @@
...
+
+/-!
+
+## D. Conservative equations
+
+-/
+
+/-- Conservation of mass in conservative form, `partial_t rho + div (rho u) = 0`. -/
+def ContinuityEquation (d : ℕ) (fluid : FluidState d) : Prop :=
+  ∀ time position,
```
jstoobysmith: Probably want a separate file for this.

jstoobysmith:
This length PR is ok - just need more segmentation between the different results in it.

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: done

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: I'm also struggling with this. Perhaps `MomentumBalanceFields` is better?

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: agreed! We now define `VectorField` and base the `MomentumDensityField` on that

Physlib/SpaceAndTime/Space/Derivatives/MatrixDiv.lean
FloWsnr: done

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: We now split:

- Continuity.lean: mass conservation only.
- Momentum.lean: momentum equations and their equivalence, assuming continuity.
- Basic.lean: combines continuity + momentum into full Navier-Stokes forms.

Physlib/FluidDynamics/NavierStokes/Basic.lean
jstoobysmith: Maybe `FluidInMomentumBalance`?

Physlib/FluidDynamics/NavierStokes/Continuity.lean
```diff
@@ -0,0 +1,63 @@
...
+
+/-!
+
+## A. Continuity equation
+
+-/
+
+/-- Conservation of mass in conservative form, `partial_t rho + div (rho u) = 0`. -/
+def ContinuityEquation (d : ℕ) (fluid : FluidState d) : Prop :=
+  ∀ time position,
```
jstoobysmith: Should there be some condition on differentiability here?

Physlib/FluidDynamics/NavierStokes/Momentum.lean
```diff
@@ -0,0 +1,306 @@
...
+
+/-- The left-hand side of the conservative momentum equation. -/
+noncomputable def conservativeMomentumLHS (d : ℕ) (fluid : FluidState d) : VectorField d :=
+  fun time position =>
+    ∂ₜ (fun time' => momentumDensity d fluid time' position) time +
+      matrixDiv d (momentumFlux d fluid time) position
+
+/-- The left-hand side of the convective momentum equation. -/
+noncomputable def convectiveMomentumLHS (d : ℕ) (fluid : FluidState d) : VectorField d :=
+  fun time position => fluid.rho time position • materialAcceleration d fluid.velocity time position
```
jstoobysmith: I think throughout it might be easier to use `t x` instead of `time position`.

FloWsnr: done

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: yes, that's also good. done

Physlib/FluidDynamics/NavierStokes/Continuity.lean
FloWsnr: Good question. I kept `ContinuityEquation` as just the pointwise balance-law equation. The differentiability assumptions are then supplied by the theorems that need them, such as the conservative/convective equivalence where we use product rules.
If I am not mistaken, some versions of Navier-Stokes might not need differentiability, e.g. shocks or free-boundary or interface problems.

If we later want a notion of classical fluid solution, I think it would be better to add a separate regularity predicate or solution structure rather than baking regularity into `ContinuityEquation` itself.

```lean
  def ClassicalContinuitySolution ... : Prop :=
    RegularityAssumptions fluid ∧ ContinuityEquation d fluid
```
or
```lean
  structure ClassicalFluidState extends FluidState d where
    hRhoTime : ...
    hMassFluxSpace : ....
```

Physlib/FluidDynamics/NavierStokes/Basic.lean
```diff
@@ -0,0 +1,85 @@
...
+namespace NavierStokes
+
+/-!
+
+## A. Full Navier-Stokes forms
+
+-/
+
+/-- The conservative Navier-Stokes balance-law form with an externally supplied stress tensor. -/
+def ConservativeForm (d : ℕ) (data : FluidInMomentumBalance d) : Prop :=
```
jstoobysmith: I wouldn't put these in the namespace `NavierStokes`, I would make one of them `NavierStokes` and show that it is equivalent to the other one in a theorem.

Physlib/FluidDynamics/NavierStokes/Continuity.lean
jstoobysmith: Ok, this makes sense to me. However, I think you might need to say that you only want the equations to hold where the function is actually differentiable (though don't assume it is everywhere) - otherwise you will have junk values floating around.

Physlib/FluidDynamics/NavierStokes/Continuity.lean
```diff
@@ -0,0 +1,63 @@
...
+/-!
+
+## A. Continuity equation
+
+-/
+
+/-- Conservation of mass in conservative form, `partial_t rho + div (rho u) = 0`. -/
+def ContinuityEquation (d : ℕ) (fluid : FluidState d) : Prop :=
+  ∀ t x,
+    ∂ₜ (fun t' => fluid.rho t' x) t +
```
jstoobysmith: Would put equations like this all on one line (if fit)

Physlib/FluidDynamics/NavierStokes/Momentum.lean
```diff
@@ -0,0 +1,306 @@
...
+/-- The nonlinear transport term `(u · ∇)u`. -/
+noncomputable def convectiveTerm (d : ℕ) (velocity : VelocityField d) : VectorField d :=
+  fun t x => ∑ j, velocity t x j • ∂[j] (velocity t) x
+
+/-- The material acceleration `∂ₜ u + (u · ∇)u`. -/
+noncomputable def materialAcceleration (d : ℕ) (velocity : VelocityField d) : VectorField d :=
+  fun t x =>
+    ∂ₜ (fun t' => velocity t' x) t +
+      convectiveTerm d velocity t x
+
```
jstoobysmith, line 99: I think these should be defined as function of `FluidState` rather then just a `velocity`. This might make the notation a lot easier later, as you will be able to use field notation.

Physlib/FluidDynamics/NavierStokes/Momentum.lean
```diff
@@ -0,0 +1,306 @@
...
+/-- Conservation of momentum in convective form.
+
+The equation is
+
+`rho (partial_t u + (u · ∇)u) = matrixDiv sigma + rho f`.
+
+Here `stress` is intentionally not yet specialized to a Newtonian stress law.
+-/
+def ConvectiveMomentumEquation (d : ℕ) (data : FluidInMomentumBalance d) : Prop :=
```
jstoobysmith, line 108: I would do similar to my Navier stokes suggestion here.

Physlib/FluidDynamics/NavierStokes/Momentum.lean
```diff
@@ -0,0 +1,306 @@
...
+The differentiability assumptions are exactly the product-rule assumptions used to rewrite
+`partial_t (rho u)` and `matrixDiv (rho u ⊗ u)`.
+-/
+theorem ConservativeMomentumEquation_iff_ConvectiveMomentumEquation
+    (d : ℕ) (data : FluidInMomentumBalance d)
+    (hContinuity : ContinuityEquation d data.toFluidState)
+    (hRhoTime : ∀ t x,
+      DifferentiableAt ℝ (fun t' => data.rho t' x) t)
+    (hVelocityTime : ∀ t x,
+      DifferentiableAt ℝ (fun t' => data.velocity t' x) t)
```
jstoobysmith: Try e.g. `DifferentiableAt ℝ (data.velocity · x) t`

jstoobysmith:
Would it be possible to rebase this PR with #949 or merge with that so the commits from that PR get included in the history. I think there might also be results there which are worth including and this might be the best way of doing that consistently

Physlib/FluidDynamics/NavierStokes/Basic.lean
FloWsnr: yes, that is a good point. I changed it so that the conservative form is `def NavierStokes` and the other one is called `ConvectiveNavierStokes`.
Both are now outside of the NavierStokes namespace...

Physlib/FluidDynamics/NavierStokes/Continuity.lean
FloWsnr: I now did the following:

- `ClassicalContinuityEquation` makes sure we use continuity only at points which are differentiable
- `SmoothContinuityEquation` extends this to overall differentiability as done for "normal" Navier-Stokes

Physlib/FluidDynamics/NavierStokes/Continuity.lean
FloWsnr: done

Physlib/FluidDynamics/NavierStokes/Momentum.lean
FloWsnr: done

Physlib/FluidDynamics/NavierStokes/Momentum.lean
FloWsnr: done, now the conservative form is just called `MomentumEquation` and we have an additional `ConvectiveMomentumEquation`. However, I still kept them inside the NS namespace.

Physlib/FluidDynamics/NavierStokes/Momentum.lean
FloWsnr: done

FloWsnr:
> Would it be possible to rebase this PR with #949 or merge with that so the commits from that PR get included in the history. I think there might also be results there which are worth including and this might be the best way of doing that consistently

Yes, I can do that. I will try to get a clean version where the commits are included.
My suggestion:

- we rebase / merge with the other PR, but keep the additional equations (thermodynamics etc) removed
- We merge the PR
- I create a new PR which brings back the content of the other PR

This way, this PR does not get super big. Of course I can also try to include everything in this PR if you dont mind the size and longer duration / wait time till merge.

jstoobysmith:
@FloWsnr Sounds like a good plan to me.

FloWsnr:
Just did that. The branch now contains history from #949 but no code yet. If you have time, you can merge the PR and I will start working on the next one which adds back the features from #949 based on the new structure

jstoobysmith (dismissed):
Adjusted some of the names - I've approved. Will merge once CI finishes (assuming it passes)

jstoobysmith (approved):
Approved

PR #1107, opened 2026-05-18
feat: Wirtinger calculus and N=1 SUSY scalar Wirtinger derivatives

pariandrea:
## Summary

  Adds a reusable **Wirtinger calculus** for `ℂ → ℂ` functions, and builds on it
  the **N=1 SUSY scalar Wirtinger derivatives** `∂/∂φ^I` and `∂/∂φ̄^I` on the
  chiral configuration space. The change is purely additive — 11 new files, no
  existing code modified (aside from registering the modules in `Physlib.lean`).

  ## What this adds

  **`Physlib/Mathematics/Calculus/Wirtinger/`** — a general 1-D Wirtinger calculus:
  - `dWirtinger` / `dWirtingerBar`, the operators `(1/2)(∂_x ∓ i∂_y)`;
  - their collapse to the complex derivative on holomorphic functions and the
    real-linear Wirtinger decomposition;
  - the differentiation rules — locality, real-linearity, the Leibniz rule, the
    two-term chain rule, conjugation lemmas;
  - an outer-function derivative catalog (`Complex.log`, …).

  **`Physlib/SUSY/N1/`** — the N=1 scalar sector:
  - `ChiralScalarValue n = Fin n → ℂ`, the physical configuration space, with the
    anti-chiral view as a derived conjugation;
  - `dScalar` / `dScalarBar`, the scalar Wirtinger operators, defined as the 1-D
    operators applied through a coordinate slice;
  - their calculus — coordinate rules `∂φ^J/∂φ^I = δ_IJ`, additivity, the Leibniz
    rule, the holomorphic/antiholomorphic collapse, and the outer chain rules.

  ## Worked example

  `SUSY/N1/Examples.lean` computes, in closed form, the chiral and anti-chiral
  Wirtinger derivatives of the multi-field upper-half-plane log Kähler potential
  `K = -∑_I log(i(z̄^I − z^I))`, and from them the Kähler metric
  `g_{IJ̄} = -δ_{IJ}/(z^I − z̄^I)²` — the Poincaré metric on `Hⁿ`.

PR #1113, opened 2026-05-20
feat(SUSY/N1): add chiral scalar configuration space and coordinate CLMs

pariandrea:
## Motivation

  This is the foundational layer of a formalization of the **N=1 chiral scalar sector**
   in Physlib. The broader goal is to build the Wirtinger differential calculus on
  complex field space — the `∂/∂φ^I` and `∂/∂φ̄^I` operators that underpin
  supersymmetric and Kähler geometry — and ultimately a worked Kähler-geometry
  application (the log Kähler potential on `H^n` and its metric).

  Following review feedback on the larger branch, that work is being split into
  reviewable pieces. This PR contributes **only the basic data layer**: the labels, the
   configuration space, and the coordinate maps that everything downstream is built on.
   It is fully self-contained over Mathlib, so it can be reviewed and merged
  independently of the calculus that will follow.

  ## What this adds

  `Physlib/SUSY/N1/Basic.lean`:

  - **`ChiralIndex`** — the label data of the chiral sector reduces to a single `n :
  ℕ`, the number of complex scalars. Chiral and anti-chiral indices share the one label
   set `Fin n`; in the scalar sector the bar is the identity, so no separate pairing is
   carried as data.
  - **`ChiralScalarValue n = Fin n → ℂ`** — the physical scalar configuration space,
  carrying exactly `2n` real degrees of freedom. Declared as an `abbrev` so Mathlib's
  `α → ℂ` calculus lemmas apply by unification.
  - **`ChiralScalarValue.antiChiral`** — the anti-chiral view as the *derived*
  pointwise complex conjugate, not an independent input. Every `φ̄` formula factors
  through this, keeping the configuration space physical and the `2n` count honest.
  - **`chiralCoordCLM I` / `antiChiralCoordCLM I`** — the per-coordinate projections
  `z^I = u I` and `z̄^I = star (u I)` as continuous ℝ-linear maps. Both are continuous
  and real-linear, but only `chiralCoordCLM` is ℂ-linear — `antiChiralCoordCLM` factors
   through conjugation and must live at the `→L[ℝ]` level.

### DIFF
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -164,6 +164,7 @@ public import Physlib.Particles.SuperSymmetry.MSSMNu.AnomalyCancellation.OrthogY
 public import Physlib.Particles.SuperSymmetry.MSSMNu.AnomalyCancellation.OrthogY3B3.ToSols
 public import Physlib.Particles.SuperSymmetry.MSSMNu.AnomalyCancellation.Permutations
 public import Physlib.Particles.SuperSymmetry.MSSMNu.AnomalyCancellation.Y3
+public import Physlib.Particles.SuperSymmetry.N1.Basic
 public import Physlib.Particles.SuperSymmetry.SU5.ChargeSpectrum.AllowsTerm
 public import Physlib.Particles.SuperSymmetry.SU5.ChargeSpectrum.Basic
 public import Physlib.Particles.SuperSymmetry.SU5.ChargeSpectrum.Completions
@@ -368,7 +369,6 @@ public import Physlib.Relativity.Tensors.RealTensor.Velocity.Basic
 public import Physlib.Relativity.Tensors.TensorSpecies.Basic
 public import Physlib.Relativity.Tensors.Tensorial
 public import Physlib.Relativity.Tensors.UnitTensor
-public import Physlib.SUSY.N1.Basic
 public import Physlib.SpaceAndTime.Space.Basic
 public import Physlib.SpaceAndTime.Space.ConstantSliceDist
 public import Physlib.SpaceAndTime.Space.CrossProduct
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -368,6 +368,7 @@ public import Physlib.Relativity.Tensors.RealTensor.Velocity.Basic
 public import Physlib.Relativity.Tensors.TensorSpecies.Basic
 public import Physlib.Relativity.Tensors.Tensorial
 public import Physlib.Relativity.Tensors.UnitTensor
+public import Physlib.SUSY.N1.Basic
 public import Physlib.SpaceAndTime.Space.Basic
 public import Physlib.SpaceAndTime.Space.ConstantSliceDist
 public import Physlib.SpaceAndTime.Space.CrossProduct
--- a/Physlib/Particles/SuperSymmetry/N1/Basic.lean
+++ b/Physlib/Particles/SuperSymmetry/N1/Basic.lean
@@ -37,7 +37,7 @@ This file carries only the index/configuration data; the chiral and anti-chiral
 derivatives `∂_I` / `∂_J̄` (the `Model` methods `M.dChiralScalar` /
 `M.dAntiChiralScalar`) are built on top of it in `Derivative.lean`.
 
-The files of this `SUSY/N1/` folder:
+The files of this `Particles/SuperSymmetry/N1/` folder:
 
 * `Basic.lean` (this file) — the `Model` indexing data and the chiral
   configuration type.
--- a/Physlib/SUSY/N1/Basic.lean
+++ b/Physlib/SUSY/N1/Basic.lean
@@ -1,0 +1,88 @@
+/-
+Copyright (c) 2026 Andrea Pari. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Andrea Pari
+-/
+module
+
+public import Mathlib.Data.Fintype.Defs
+public import Mathlib.Analysis.Complex.Basic
+public import Mathlib.Logic.Equiv.Basic
+
+/-!
+
+# SUSY N=1 chiral sector — basic indexing data
+
+## i. Overview
+
+The minimal label and configuration data for the N=1 chiral sector.
+
+A `Model` packages two finite index types — `ChiralIndexingType` (written `C`
+below) indexing the chiral scalars and `AntiChiralIndexingType` (written `A`)
+the anti-chiral (barred) slots — with an equivalence `equiv : C ≃ A`, kept
+distinct so the two slots of a contraction `g_{IJ̄}` cannot be confused. The
+physical configuration is `ChiralScalarConfiguration C = C → ℂ`, carrying
+`2 · Fintype.card C` real degrees of freedom — the only field data; the
+anti-chiral scalars are its complex conjugates, never an independent
+configuration.
+
+Doubling — adding an independent `A → ℂ` configuration — is the wrong choice:
+the physical identities would then hold only where anti-chiral is the conjugate
+of chiral, so every theorem carries that subspace as a hypothesis (e.g. the
+Kähler potential is real only there, making hermiticity of `g_{IJ̄}`
+conditional). Building the conjugate in keeps them unconditional — `K` is real
+by construction — and leaves `A` an index type only.
+
+This file carries only the index/configuration data; the chiral and anti-chiral
+derivatives `∂_I` / `∂_J̄` (the `Model` methods `M.dChiralScalar` /
+`M.dAntiChiralScalar`) are built on top of it in `Derivative.lean`.
+
+The files of this `SUSY/N1/` folder:
+
+* `Basic.lean` (this file) — the `Model` indexing data and the chiral
+  configuration type.
+* `Derivative.lean` — the chiral / anti-chiral derivative wrappers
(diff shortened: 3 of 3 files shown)


Physlib/SUSY/N1/Basic.lean
```diff
@@ -0,0 +1,179 @@
...
+The label data of the chiral sector reduces to a single natural number
+`n : ℕ`: the count of complex scalars. The chiral and anti-chiral indices
+both inhabit `Fin n`. There is no separate bar pairing as data, since in
+the scalar sector the bar is the identity.
+
+-/
+
+/-- The label data of an N=1 chiral sector: simply the number of complex
+scalars. The chiral and anti-chiral indices both range over `Fin n`. -/
+abbrev ChiralIndex : Type := ℕ
```
jstoobysmith: I have two questions about this definition: 

1. I think `ChiralIndex` might be the wrong name here, as an element of `ChiralIndex` is the number of chiral fields, whilst I think a chiral index is really an element of `Fin n`. 
2. what stops you generalizing this to an arbitary `Fintype`, rather than `Fin n`? 

I wonder if something like the following is better (I think discussed before): 

```lean
structure Model (ChiralIndexingType : Type) [Fintype ChiralIndexingType] 
    (AntiChiralIndexingType : Type) [Fintype AntiChiralIndexingType] where
   equiv : ChiralIndexingType ≃ AntiChiralIndexingType
```

Physlib/SUSY/N1/Basic.lean
```diff
@@ -0,0 +1,179 @@
...
+this carries `2n` real degrees of freedom, on which the Wirtinger
+operators of `SUSY/N1/Deriv.lean` are built as real differential
+operators.
+
+-/
+
+/-- The chiral scalar configuration: an assignment of complex values to each
+chiral label. Declared as `abbrev` so unification can apply Mathlib's
+calculus lemmas about `α → ℂ` directly. -/
+abbrev ChiralScalarValue (n : ChiralIndex) := Fin n → ℂ
```
jstoobysmith: I think `ChiralScalarConfiguration` would be a better name. Maybe add to the doc-string something like: 
The term of this type gives a value to each of the chiral scalar fields.

jstoobysmith: I think you likely also need a type `AntiChiralScalarConfiguration`

Physlib/SUSY/N1/Basic.lean
```diff
@@ -0,0 +1,179 @@
...
+-/
+
+namespace ChiralScalarValue
+
+variable {n : ChiralIndex}
+
+/-- The anti-chiral view of a chiral scalar configuration: pointwise complex
+conjugation. Realises the physical statement that in the scalar sector,
+barring is complex conjugation. -/
+def antiChiral (u : ChiralScalarValue n) : ChiralScalarValue n :=
```
jstoobysmith: Should be able to lift this to a CLM.

Physlib.lean
```diff
@@ -351,6 +347,7 @@ public import Physlib.Relativity.Tensors.RealTensor.Velocity.Basic
 public import Physlib.Relativity.Tensors.TensorSpecies.Basic
 public import Physlib.Relativity.Tensors.Tensorial
 public import Physlib.Relativity.Tensors.UnitTensor
+public import Physlib.SUSY.N1.Basic
```
doxtor6: We already have Physlib.Particles.SuperSymmetry, maybe we should put it there? Also if it is about Wirtinger calculus, maybe we should put this PR in Physlib.Mathematics

Physlib/SUSY/N1/Basic.lean
pariandrea: Adopted your suggestion: Model is now exactly that two-type structure.
The index types are arbitrary Type* with [Fintype …] now (not Fin n), and the
 ChiralIndex name is gone. In the client files I write the two types as the shorter C /
 A to keep signatures within the 100-col limit, with
 a one-line note in the overview bridging the descriptive names to C / A.

Physlib/SUSY/N1/Basic.lean
pariandrea: Changed the name to `ChiralScalarConfiguration` as suggested.

I did not add an independent `A → ℂ` configuration as I think it's the wrong choice. The physical identities would then hold only where the anti-chiral field is the conjugate of the chiral one, so every theorem would carry that subspace as a hypothesis (e.g. the Kähler potential is real only there, making hermiticity of
`g_{IJ̄}` conditional). Building the conjugate in keeps them unconditional: `K` is real
 by construction, and `A` is left as an index type only. We still keep distinct index
 types to enforce the correct contraction rules, but there is a single field
 configuration, avoiding a fictitious doubling of the physical DOF.

Physlib/SUSY/N1/Basic.lean
pariandrea: `CLM.lean` is gone. Those CLMs aren't SUSY-specific, they're general operations on a
  finite-index complex coordinate space `C → ℂ` — so rather than fold them into the SUSY
  `Basic.lean` I moved them down into the Wirtinger calculus, where anything on `C → ℂ`
  can reuse them:
  - `Basic.lean` — the ℝ/ℂ scalar-restriction bridge plus the coordinate-basis CLMs and
  conjugation.
  - `Multivariable.lean` — the directional Wirtinger derivatives and their full calculus
  (Leibniz, chain rule, conjugation, holomorphic collapse), capped by Schwarz's
  theorem.
  - `Coordinate.lean` — the coordinate operators `dWirtingerCoord` /
  `dWirtingerCoordBar`, the projection/conjugation CLMs (`coordProjCLM`, `conjCoordCLM`,
  `conjCLM`), and coordinate Schwarz.
  - `UpperHalfPlane.lean` — reusable upper-half-plane lemmas.

  The full Wirtinger calculus can be seen in the larger PR #1107.

Physlib.lean
pariandrea: Wirtinger calculus agreed, it's general complex analysis with no SUSY content, so in
the larger PR #1107 it already lives under `Physlib/Mathematics/Calculus/Wirtinger/`. (This PR is just the bite-sized `Basic.lean` split out of #1107.)

SUSY placement is a good point. The existing `Physlib/Particles/SuperSymmetry/` is MSSM
anomaly cancellation (`MSSMNu/`) and SU(5)+U(1) charge spectra (`SU5/`). The work in #1107 is the other side of SUSY: the continuous scalar-sector geometry (Kähler potential and metric, holomorphic superpotential). It could probably be moved in `N1/` subfolder.

Physlib/Particles/SuperSymmetry/N1/Basic.lean
```diff
@@ -0,0 +1,88 @@
...
+`equiv` between them. -/
+structure Model (ChiralIndexingType : Type*) [Fintype ChiralIndexingType]
+    (AntiChiralIndexingType : Type*) [Fintype AntiChiralIndexingType] where
+  /-- The correspondence between chiral and anti-chiral (barred) labels. -/
+  equiv : ChiralIndexingType ≃ AntiChiralIndexingType
+
+/-- The chiral scalar configuration: an assignment of complex values to each
+chiral label. An `abbrev` so unification applies Mathlib's `α → ℂ` calculus
+lemmas directly. -/
+abbrev ChiralScalarConfiguration (ChiralIndexingType : Type*) := ChiralIndexingType → ℂ
```
jstoobysmith, line 82: I would change this to the following. It means you end up carrying around 'unused' variables, but I think it will help with the overall consistency. 

```

namespace Model 

set_option linter.unusedVariables false in
/-- The chiral scalar configuration: an assignment of complex values to each
chiral label. An `abbrev` so unification applies Mathlib's `α → ℂ` calculus
lemmas directly. -/
@[nolint unusedArguments]
abbrev ChiralScalarConfiguration {ChiralIndexingType} (M : Model ChiralIndexingType) := ChiralIndexingType → ℂ

end Model
```

pariandrea: Thanks for the suggestion. I tried but it did not turn out easy to do.

ChiralScalarConfiguration is really just C → ℂ, so it only depends on the index type, not on a whole Model. Making it a Model method means the KahlerPotential/SuperPotential structures have to carry a model term they never use, and for concrete examples Lean then can't infer the index type on its own.

The concrete example is in the larger PR:

def K : KahlerPotential C            -- current 
def K : KahlerPotential model        -- C hidden in `model` → Lean can't find it

I think this would be resolved by manual type annotations all over the place to compensate. 
It is entirely possible that I am implementing the change naively and there is a cleaner way.

So I'd keep it as it is currently, and use the M. dot-notation for the operations that genuinely use the model (dChiralScalar, dAntiChiralScalar, metric). I am open to suggestions and happy to revisit this.

jstoobysmith:
This looks good to me now. I think placing it in ./Particles/Supersymmetry might be better though (or at least making sure we have only one folder for SUSY - whether that is moving the stuff there or here).

pariandrea:
I'll move the new code under Particles/SuperSymmetry/ rather than reorganize the existing SU5/MSSMNu files. Is Particles/SuperSymmetry/N1/ the correct path?

jstoobysmith:
Yes please! After this - I think we're all good to go.

doxtor6:
It looks good to me too.

jstoobysmith (approved):
Approved - will merge when linters pass. Many thanks.

PR #1114, opened 2026-05-21
feat(QM): Add `SpaceDQuantumSystem`

gloges:
Introduces a general structure for single-particle QM with Hilbert space `SpaceDhilbertSpace`, characterized by the number of dimensions, mass and potential function.
- Defines kinetic/potential/hamiltonian operators and introduces some basic API.
- Adds some lemmas for the potential operators when the potential _function_ satisfies some basic properties.

As a trial, `HydrogenAtom` is changed to extend this structure with a specific form for the potential.

### DIFF
--- a/Physlib/QuantumMechanics/DDimensions/Hydrogen/Basic.lean
+++ b/Physlib/QuantumMechanics/DDimensions/Hydrogen/Basic.lean
@@ -5,7 +5,7 @@ Authors: Gregory J. Loges
 -/
 module
 
-public import Physlib.QuantumMechanics.DDimensions.Operators.Momentum
+public import Physlib.QuantumMechanics.DDimensions.Basic
 public import Physlib.QuantumMechanics.DDimensions.Operators.Position
 /-!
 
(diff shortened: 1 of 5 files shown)


gloges:
t-quantum-mechanics

Physlib/QuantumMechanics/DDimensions/Hydrogen/Basic.lean
```diff
@@ -27,35 +27,49 @@ e.g. see https://doi.org/10.1103/PhysRevA.80.032507 and https://doi.org/10.1063/
...
-structure HydrogenAtom where
-  /-- Number of spatial dimensions -/
-  d : ℕ
-
-  /-- Mass (positive) -/
-  m : ℝ
-  hm : 0 < m
-
-  /-- Coefficient in potential (positive for attractive) -/
+structure HydrogenAtom extends SpaceDQuantumSystem where
```
jstoobysmith, line 35: Why this over e.g. 

```lean 
def hydrogenAtom (k : Real) : SpaceDQuantumSystem where 
   ... 
```

gloges: Would it not need to be
```lean
def hydrogenAtom (d : ℕ) {m : ℝ} (hm : 0 < m) (k : ℝ) : SpaceDQuantumSystem where
  d := d
  m := m
  hm := hm
  potential := fun x ↦ -k * ‖x‖⁻¹
```
and then one would not be able to have, for example, an associated LRL vector (which depends on both `m` and `k`)?

jstoobysmith: Yeah you are right, you would need to carry around the masses aswell. Maybe the way you have done it here is best then.

jstoobysmith (approved):
I've approved, but won't merge here until you've confirmed

Physlib/QuantumMechanics/DDimensions/Hydrogen/Basic.lean
gloges: I will experiment more with this - the way the form of the potential is specified seems clunky, but should work for now.

PR #1115, opened 2026-05-21
feat: Add scalarPotential of constructors

jstoobysmith:
Add lemmas giving the scalar potential of EMPotentials constructed using constructors. 
Also fix the ofScalarPotential constructor to include the speed of light.


zhikaip (approved):
lgtm, thanks

PR #1118, opened 2026-05-22
refactor: Move distributional EM to new file system

jstoobysmith:
The minimal changes needed to move the Distributional EM results to their own file system whilst keeping the documentation consistent, and everything building. 

I've done this in small steps which can be seen by the commits.


zhikaip (approved):
I think this is a good split, we can clean up the respective parts afterwards if necessary

PR #1119, opened 2026-05-22
feat: Constructive form of curl zero eq grad

jstoobysmith:
Add a constructive form of the statement that if the curl of a result is zero, then it is equal to a grad. This allows one not to rely on choice.

### DIFF
--- a/Physlib/SpaceAndTime/Space/Derivatives/Curl.lean
+++ b/Physlib/SpaceAndTime/Space/Derivatives/Curl.lean
@@ -646,6 +646,12 @@ lemma exists_grad_of_curl_zero (f : Space → EuclideanSpace ℝ (Fin 3)) (hf : 
   specialize hg x
   simpa using (HasGradientAt.unique (hg.differentiableAt.hasGradientAt_grad x) hg).symm
 
+/-- A constructive form of the statement that if the curl of a function is zero,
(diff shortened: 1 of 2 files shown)


Physlib/SpaceAndTime/Space/Derivatives/Curl.lean
```diff
@@ -646,6 +646,21 @@ lemma exists_grad_of_curl_zero (f : Space → EuclideanSpace ℝ (Fin 3)) (hf :
   specialize hg x
   simpa using (HasGradientAt.unique (hg.differentiableAt.hasGradientAt_grad x) hg).symm
 
+lemma eq_grad_integral_of_curl_zero (f : Space → EuclideanSpace ℝ (Fin 3)) (hf : ContDiff ℝ 1 f)
```
zhikaip, line 655: Maybe add a doc string about which this is constructing? My understanding is that practically it's the potential such that the value at origin is zero, and more sophisticatedly some sort of homotopy operator defined by integration along radial lines (not sure the exact wording here)?

jstoobysmith: Added a doc-string here, in relation to the work-done.

zhikaip (approved):
lgtm, thanks

PR #1120, opened 2026-05-23
feat(QM): Add `LinearPMap.sum`

gloges:
Adds to the functionality of unbounded operators:
- Defines `LinearPMap.sum` for finite sums of partial linear maps. While `sum f` and `∑ a, f a` are equal, it is not by definition and the latter is difficult to work with.
- Proves that `U` is densely defined iff `U.closure` is densely defined, allowing redundant hypotheses to be removed for essentially self-adjoint operators.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good - approved. will merge shortly.

PR #1125, opened 2026-05-26
feat(FluidDynamics): Adding more fluid dynamics - continuation of PR #949 and #1112 ,

FloWsnr:
## Summary

This PR extends the fluid dynamics API with reusable foundations for incompressibility, continuity, momentum balance, Euler flow, and Bernoulli flow based. It copies PR #949 ideas refactored based on the structure agreed upon in #1112 .

### Structure

I realized that we previously placed general equations into the Navier-Stokes namespace which instead should be general FluidDynamics, since they will be reused by systems other than Navier-Stokes. 

- Moves the continuity equation out of `NavierStokes/` into reusable `FluidDynamics/Continuity.lean`.
- Extracts generic momentum-density, momentum-flux, material-acceleration, and conservative/convective momentum LHS definitions into `FluidDynamics/Momentum.lean`.
- Updates Navier-Stokes files to reuse the shared continuity and momentum infrastructure.

### Additions

- Adds general incompressibility predicates in `FluidDynamics/Incompressible.lean`. This can be reused in Navier-Stokes, Euler and other fluid systems.
- Adds Euler momentum equations in conservative and convective form.
- Adds Bernoulli-flow data, steady-flow, material-derivative, isentropic-flow, Bernoulli-function, and Bernoulli-law predicates.

## Notes

This branch intentionally keeps the new fluid definitions dimension-generic and avoids recreating the older `IdealFluid` wrapper from PR #949.


Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,156 @@
...
+
+/-!
+
+## D. Equivalence of conservative and convective Euler forms
+
+-/
+
+/-- The conservative and convective Euler forms are equivalent when the fields are
+differentiable enough for the product rules. -/
+theorem Euler_iff_ConvectiveEuler
```
jstoobysmith: should be e.g. `euler_iff_convectiveEuler`

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,156 @@
...
+
+-/
+
+/-- The conservative Euler equations: classical continuity and conservative momentum balance. -/
+def Euler (d : ℕ) (data : FluidInEulerBalance d) : Prop :=
+  ClassicalContinuityEquation d data.toFluidState ∧
+    EulerMomentumEquation d data
+
+/-- The convective Euler equations: classical continuity and convective momentum balance. -/
+def ConvectiveEuler (d : ℕ) (data : FluidInEulerBalance d) : Prop :=
```
jstoobysmith: I would remove this, and just include this in the theorem below.

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,156 @@
...
+`partial_t (rho u) + matrixDiv (rho u ⊗ u) = -grad p + rho f`. -/
+def EulerMomentumEquation (d : ℕ) (data : FluidInEulerBalance d) : Prop :=
+  ∀ t x, conservativeMomentumLHS d data.toFluidState t x =
+    eulerMomentumRHS d data t x
+
+/-- Euler momentum balance in convective form,
+`rho (partial_t u + (u · grad)u) = -grad p + rho f`. -/
+def ConvectiveEulerMomentumEquation (d : ℕ) (data : FluidInEulerBalance d) : Prop :=
+  ∀ t x, convectiveMomentumLHS d data.toFluidState t x =
+    eulerMomentumRHS d data t x
```
jstoobysmith: How needed are these? Could they just be put in the propositions below direction?

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: done. Btw, is there a way to bake these rules into the linter since I keep making the same mistake?

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: done

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: did that, agreed

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,138 @@
...
+
+/-!
+
+## D. Equivalence of conservative and convective Euler forms
+
+-/
+
+/-- The conservative and convective Euler forms are equivalent when the fields are
+differentiable enough for the product rules. -/
+theorem euler_iff_convectiveEuler
```
jstoobysmith, line 65: This should now be `euler_iff_convective_euler`

Physlib/FluidDynamics/Euler/Bernoulli.lean
```diff
@@ -0,0 +1,131 @@
...
+  ∀ t x, data.bodyForce t x = -(∇ data.potential x)
+
+/-!
+
+## C. Flow-state predicates
+
+-/
+
+/-- A fluid state is steady when the velocity has zero time derivative everywhere. -/
+def isSteady (d : ℕ) (fluid : FluidState d) : Prop :=
```
jstoobysmith: Since this is a proposition, `IsSteady`

Physlib/FluidDynamics/Euler/Bernoulli.lean
```diff
@@ -0,0 +1,131 @@
...
+  ∀ t x, ∂ₜ (fluid.velocity · x) t = 0
+
+/-- The material derivative `D_t f = partial_t f + u · grad f` of a scalar field. -/
+noncomputable def materialDerivative (d : ℕ) (fluid : FluidState d)
+    (field : ScalarField d) : ScalarField d :=
+  fun t x => ∂ₜ (field · x) t + ⟪fluid.velocity t x, ∇ (field t) x⟫_ℝ
+
+/-- A Bernoulli flow is isentropic when the entropy is materially conserved. -/
+def isIsentropic (d : ℕ) (data : FluidInBernoulliFlow d) : Prop :=
```
jstoobysmith: Likewise `IsIsentropic`

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,138 @@
...
+namespace FluidDynamics
+
+/-!
+
+## A. Euler data
+
+-/
+
+/-- The fields needed for Euler momentum balance: fluid state, pressure, and body force. -/
+structure FluidInEulerBalance (d : ℕ) extends FluidState d where
```
jstoobysmith: This is ok for now - but I think we should have a think about how we name all of these things

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,138 @@
...
+  bodyForce : BodyForce d
+
+/-!
+
+## B. Euler momentum right-hand side
+
+-/
+
+/-- The right-hand side of Euler momentum balance, `-grad p + rho f`. -/
+noncomputable def eulerMomentumRHS (d : ℕ) (data : FluidInEulerBalance d) : VectorField d :=
```
jstoobysmith: Is there a better name for this which gives across the physical meaning? If not, maybe we should not define it explicitly. Similarly with other definitions like this.

Physlib/FluidDynamics/Euler/Basic.lean
jstoobysmith: The idea would be to have some key data structures `FluidState`, `FluidInEulerBalance`, etc. with which we organize the folders around.

Physlib/FluidDynamics/Euler/Basic.lean
jstoobysmith: This should be possible. If you make a comment here about this, someone might pick it up as something to do: https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/Infrastructure.20ideas/with/597860431

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: done

Physlib/FluidDynamics/Euler/Bernoulli.lean
FloWsnr: done

Physlib/FluidDynamics/Euler/Bernoulli.lean
FloWsnr: done

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: I now called it `eulerForceDensity` which is accurate but not really great. We can also inline twice, for both Euler equations. It really depends on whether future lemma will be using this. I don't know...

Physlib/FluidDynamics/Euler/Basic.lean
FloWsnr: Yes, I am not happy with the naming of all the structures, except `FluidState` which is fine.
I don't have a great idea tho. 
Perhaps it's better to define a few overarching structures like

- FluidState
- ThermodynamicFluidState

which provide "all" parameters we might need. The functions (NavierStokes, Euler, Bernoulli) would then all use these even though some fields are not needed for this.
But again, not super convinced yet

Physlib/FluidDynamics/Momentum.lean
```diff
@@ -0,0 +1,193 @@
...
+namespace FluidDynamics
+
+/-!
+
+## A. Momentum fields
+
+-/
+
+/-- The momentum density `rho u`. -/
+def momentumDensity (d : ℕ) (fluid : FluidState d) : MomentumDensityField d :=
```
jstoobysmith: All these should be in the FluidState namespac

Physlib/FluidDynamics/Euler/Basic.lean
jstoobysmith: Would splitting based on kinematics vs dynamics work? Which of these quantities are actually derived and which define the system?

jstoobysmith:
I think it might be best to sort this `Fluid` structure out first before merging this, as I think it will help with the organization of things.

FloWsnr:
I spent some time thinking about this and this is my current favorite:

Two main fluid data structures:

The generic kinematic fluid data
```lean
structure FluidFlow (d : ℕ) where
    massDensity : MassDensity d
    velocity : VelocityField d
```
and the generic dynamic fluid data, based on the generic [Cauchy momentum equation.](https://en.wikipedia.org/wiki/Cauchy_momentum_equation)

```lean
structure CauchyFlow (d : ℕ) extends FluidFlow d where
    stress : CauchyStress d
    bodyForce : SpecificBodyForce d
```

We can then use definitions/predicates for the physical cases.
```lean
def IsUnforced (d : ℕ) (flow : CauchyFlow d) : Prop :=
    ∀ t x, flow.bodyForce t x = 0

def HasConservativeBodyForce
      (d : ℕ) (flow : CauchyFlow d) (potential : Space d → ℝ) : Prop :=
    ∀ t x, flow.bodyForce t x = -(∇ potential x)

def IsInviscid
      (d : ℕ) (flow : CauchyFlow d) (pressure : ScalarField d) : Prop :=
    ∀ t x, flow.stress t x = -pressure t x • 1

def IsNewtonian
      (d : ℕ) (flow : CauchyFlow d)
      (pressure shearViscosity bulkViscosity : ScalarField d) : Prop :=
```

And define force-density helpers:

```lean
noncomputable def bodyForceDensity (d : ℕ) (flow : CauchyFlow d) : VectorField d :=
    fun t x => flow.massDensity t x • flow.bodyForce t x

noncomputable def stressForceDensity (d : ℕ) (flow : CauchyFlow d) : VectorField d :=
    fun t x => matrixDiv d (flow.stress t) x

noncomputable def totalForceDensity (d : ℕ) (flow : CauchyFlow d) : VectorField d :=
    fun t x => stressForceDensity d flow t x + bodyForceDensity d flow t x
```
  
Then the aforementioned Cauchy momentum equation is the generic:

```lean
def CauchyMomentumEquation (d : ℕ) (flow : CauchyFlow d) : Prop :=
    ∀ t x, conservativeMomentumLHS d flow.toFluidFlow t x = totalForceDensity d flow t x
```

Euler becomes:
```lean
def Euler (d : ℕ) (flow : CauchyFlow d) (pressure : ScalarField d) : Prop :=
    ClassicalContinuityEquation d flow.toFluidFlow ∧
      CauchyMomentumEquation d flow ∧
        IsInviscid d flow pressure
```
  Navier-Stokes becomes:
```lean
def NavierStokes
      (d : ℕ) (flow : CauchyFlow d)
      (pressure shearViscosity bulkViscosity : ScalarField d) : Prop :=
    ClassicalContinuityEquation d flow.toFluidFlow ∧
      CauchyMomentumEquation d flow ∧
        IsNewtonian d flow pressure shearViscosity bulkViscosity
```

If we want to work with thermodynamic equations (Bernoulli, energy balances etc), we should introduce a single new structure with combines flow with thermodynamics.

```lean
structure ThermodynamicCauchyFlow (d : ℕ) extends CauchyFlow d where
    entropy : ScalarField d
    enthalpy : ScalarField d
-- perhaps also
    temperature : ScalarField d
    heatFlux : VectorField d
```

With this, we can basically describe all fluid flow situations.
What do you think of this?

jstoobysmith:
It is unclear why something like `pressure` is not in `FluidFlow` here? Maybe because you can get it from the stress?

jstoobysmith:
I also think something like `HasConservativeBodyForce` should be an exists statement, not include a potential.

FloWsnr:
yes exactly, you get it from the stress:

In my idea, FluidFlow is meant to be the minimal kinematic/mass-transport object, not the full dynamic fluid model. Pressure is not needed for continuity, incompressibility, material derivatives, or kinetic quantities.

Once we move to momentum balance, I think the more invariant field is the Cauchy stress tensor; pressure then appears through stress laws, e.g. inviscid stress σ = -p I or Newtonian stress σ = -p I + τ.

jstoobysmith:
How does viscosity come into this all?

FloWsnr:
Viscosity is just another form of **stress** in the Cauchy version.
As above for the Newtonian stress:

```lean
σ = -p I + viscousStress
```

The total stress is `pressure (p)` plus `viscosity`.

jstoobysmith:
Ok great! This looks good then. I wonder if FluidKinematics and FluidDynamics would be better names, but I’m also happy with what you have callled them. Would you be able to restructure the API around these?

FloWsnr:
Great!
`FluidKinematics` could be okay, but `FluidDynamics` would be ugly inside the namespace which is already called `FluidDynamics` as well, right? 

I will get right onto it and refactor...

jstoobysmith:
Yeah - ok let’s keep names as they are

FloWsnr:
okay, check out the new structure and let me know what you think:

## Main changes

- We renamed the minimal carrier from `FluidState` to `FluidFlow`. It still contains only `rho` and `velocity`,

- We replaced the old usage-shaped structures with 2 physically meaningful layers:
  - `CauchyFlow extends FluidFlow` with `stress` and `bodyForce`
  - `ThermodynamicCauchyFlow extends CauchyFlow` with `entropy` and `enthalpy`.

- We added a shared Cauchy momentum layer. This defines the generic balance
  `conservativeMomentumLHS = matrixDiv stress + rho • bodyForce`
  plus the convective version and their equivalence theorem. This replaces the old Navier-Stokes-specific momentum module.

- Euler now uses `CauchyFlow` plus an `IsInviscid` predicate. Pressure appears only through the stress law `stress = -p I`, not as a field of the flow.

- Navier-Stokes now uses `CauchyFlow` plus an `IsNewtonian` predicate. Pressure and viscosities are parameters to `newtonianStressTensor`, so viscosity enters through stress, exactly as discussed in the PR.

- Bernoulli no longer has a separate `FluidInBernoulliFlow` structure. It uses `ThermodynamicCauchyFlow` and takes the external potential explicitly. `HasConservativeBodyForce` was changed to existential form: there exists a potential.

## Side changes

- Reorganized namespaces: FluidFlow.* now contains flow-derived continuity, incompressibility, momentum, acceleration, and time-independence APIs. CauchyFlow.* now contains Cauchy momentum equations and Cauchy stress-law helpers/predicates. Equation systems like `Euler`, `ConvectiveEuler`, `NavierStokes`, and `ConvectiveNavierStokes` stayed top-level in FluidDynamics.

- Added the inviscid bridge theorem: `CauchyFlow.matrixDiv_stress_eq_neg_grad_pressure_of_is_inviscid`, showing that inviscid stress -p I gives the usual -grad p force term.


- Moved flow time-independence predicates into FluidFlow: `FluidFlow.DensityTimeIndependent` and `FluidFlow.VelocityTimeIndependent`.

- Removed Bernoulli’s local `IsSteady`, since it only described velocity time-independence and now lives more accurately as `FluidFlow.VelocityTimeIndependent`.

Physlib/FluidDynamics/Basic.lean
jstoobysmith, line 1: Maybe make this file `Basic.lean`

Physlib/FluidDynamics/FluidFlow.lean
```diff
@@ -0,0 +1,121 @@
...
+public import Physlib.SpaceAndTime.Time.Derivatives
+/-!
+
+# Fluid flows
+
+## i. Overview
+
+This module defines the basic fields used to describe a fluid on `d`-dimensional space.
+The core structure `FluidFlow` contains only the density and velocity fields. Additional
+fields used by momentum balance and thermodynamic laws are provided by extension structures.
```
jstoobysmith: Probably write a doc string about the implementation choice here, and why we have chosen the three different structures.

Physlib/FluidDynamics/CauchyMomentum.lean
```diff
@@ -0,0 +1,139 @@
...
+
+-/
+
+/-- The conservative and convective Cauchy momentum equations are equivalent when the classical
+continuity equation holds.
+
+The differentiability assumptions are exactly the product-rule assumptions used to rewrite
+`partial_t (rho u)` and `matrixDiv (rho u ⊗ u)`.
+-/
+theorem cauchy_momentum_iff_convective_cauchy_momentum
```
jstoobysmith: This can have the names of the definitions in it, but would be 

`cauchyMomentumEquation_iff_convectiveCauchyMomentumEquation`

Physlib/FluidDynamics/Euler/Bernoulli.lean
```diff
@@ -0,0 +1,112 @@
...
+  ∀ t x, materialDerivative d flow.toFluidFlow flow.entropy t x = 0
+
+/-!
+
+## C. Bernoulli function
+
+-/
+
+/-- The specific kinetic energy `|u|^2 / 2` of a fluid flow. -/
+noncomputable def specificKineticEnergy (d : ℕ) (fluid : FluidFlow d) : ScalarField d :=
```
jstoobysmith: I think something like this, and `materialDerivative` should go in the `Basic.lean` file

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,130 @@
...
+
+/-!
+
+## C. Equivalence of conservative and convective Euler forms
+
+-/
+
+/-- The conservative and convective Euler forms are equivalent when the fields are
+differentiable enough for the product rules. -/
+theorem euler_iff_convective_euler
```
jstoobysmith: Same with nameing as mentioned below.

Physlib/FluidDynamics/Euler/Basic.lean
```diff
@@ -0,0 +1,130 @@
...
+namespace CauchyFlow
+
+/-!
+
+## A. Inviscid stress law
+
+-/
+
+/-- A Cauchy flow is inviscid with pressure `p` when its stress is `-p I`. -/
+def IsInviscid (d : ℕ) (flow : CauchyFlow d) (pressure : ScalarField d) : Prop :=
```
jstoobysmith: Maybe include in the docstring here what Inviscid means here

Physlib/FluidDynamics/Euler/Basic.lean
jstoobysmith, line 1: I would think about wether it makes sense to have folders 

```
./CaucyFlow
./FluidState
./ThermodynamicFluid
```

jstoobysmith (commented):
Sorry for my delay here - have been travelling.

jstoobysmith:
@FloWsnr Regarding this, if you don't think you have time to address the reviews, we could put it in the new PhyslibAlpha directory for now. See: 

https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/PhyslibAlpha/with/601570535

FloWsnr:
Hey, I'm sorry, had some travel & new job. Still plan to do the final fixes asap. I'll try this weekend!

FloWsnr:
I think I addressed all comments now except for the larger rename / restructure into subdirs
> I would think about whether it makes sense to have folders

Not sure if this split up the code too much, but if you want I can quickly do that

jstoobysmith:
@FloWsnr Sorry I think I missed this comment. I think the larger restructure would be good here, as it will make it easier to see what is related to what structure.

FloWsnr:
@jstoobysmith No problem! I just attempted the restructure. Let me know what you think!

Physlib/FluidDynamics/FluidFlow/Continuity.lean
```diff
@@ -67,16 +70,17 @@ This version records the first-order regularity needed by the classical continui
...
 -/
-def SmoothContinuityEquation (d : ℕ) (fluid : FluidState d) : Prop :=
+def SmoothContinuityEquation (d : ℕ) (fluid : FluidFlow d) : Prop :=
   (∀ x, Differentiable ℝ (fluid.rho · x)) ∧
     (∀ t, Differentiable ℝ (fun x => fluid.rho t x • fluid.velocity t x)) ∧
       ∀ t x, continuityResidual d fluid t x = 0
 
 /-- A smooth continuity equation satisfies the guarded classical continuity equation. -/
-lemma SmoothContinuityEquation.toClassical (d : ℕ) (fluid : FluidState d) :
+lemma SmoothContinuityEquation.toClassical (d : ℕ) (fluid : FluidFlow d) :
```
jstoobysmith: this should be called something along the lines `classicalContinuityEquation_of_smoothContinuityEquation`.

Physlib/FluidDynamics/FluidFlow/Incompressible.lean
```diff
@@ -0,0 +1,75 @@
...
+    incompressibilityResidual d fluid t x = 0
+
+/-- A smooth incompressible flow has globally differentiable velocity and vanishing
+incompressibility residual everywhere. -/
+def SmoothIncompressible (d : ℕ) (fluid : FluidFlow d) : Prop :=
+  (∀ t, Differentiable ℝ (fluid.velocity t)) ∧
+    ∀ t x, incompressibilityResidual d fluid t x = 0
+
+/-- A smooth incompressible flow is classically incompressible. -/
+lemma SmoothIncompressible.toClassical (d : ℕ) (fluid : FluidFlow d) :
```
jstoobysmith: Similar here with the naming.

Physlib/FluidDynamics/FluidFlow/Kinematics.lean
```diff
@@ -0,0 +1,78 @@
...
+
+/-- A fluid flow has time-independent density when the density has zero time derivative at
+each spatial point. -/
+def DensityTimeIndependent (d : ℕ) (fluid : FluidFlow d) : Prop :=
+  ∀ t x, ∂ₜ (fluid.rho · x) t = 0
+
+/-- A fluid flow has time-independent velocity when the velocity has zero time derivative at
+each spatial point. -/
+def VelocityTimeIndependent (d : ℕ) (fluid : FluidFlow d) : Prop :=
+  ∀ t x, ∂ₜ (fluid.velocity · x) t = 0
```
jstoobysmith: Not sure how needed these will be, maybe we could just use e.g. `∀ t x, ∂ₜ (fluid.velocity · x) t = 0` explicitly.

Physlib/FluidDynamics/FluidFlow/Momentum.lean
```diff
@@ -54,76 +54,43 @@ namespace NavierStokes
...
 /-!
 
-## D. Equivalence of conservative and convective momentum
+## B. Conservative and convective left-hand sides
 
 -/
 
 /-- The left-hand side of the conservative momentum equation. -/
-noncomputable def conservativeMomentumLHS (d : ℕ) (fluid : FluidState d) : VectorField d :=
+noncomputable def conservativeMomentumLHS (d : ℕ) (fluid : FluidFlow d) : VectorField d :=
```
jstoobysmith: Still not sure about these definitions, if they only appear in a given lemma, then we should write it out explicitly in that lemma or prop.

Physlib/FluidDynamics/CauchyFlow/NavierStokes.lean
jstoobysmith, line 1: Would move to the `CauchyFlow` file.

Physlib/FluidDynamics/CauchyFlow/Newtonian.lean
```diff
@@ -0,0 +1,76 @@
...
+namespace CauchyFlow
+
+/-!
+
+## A. Newtonian stress law
+
+-/
+
+/-- The spatial velocity-gradient matrix, with entries `partial_j u_i`. -/
+noncomputable def velocityGradient (d : ℕ) (flow : FluidFlow d) :
```
jstoobysmith: This should be in `FluidFlow`, no need to redefine it here.

Physlib/FluidDynamics/CauchyFlow/Newtonian.lean
```diff
@@ -0,0 +1,76 @@
...
+    Time → Space d → Matrix (Fin d) (Fin d) ℝ :=
+  fun t x i j => ∂[j] (fun x' => flow.velocity t x' i) x
+
+/-- The Newtonian stress tensor
+`-p I + mu (grad u + grad u^T) + lambda (div u) I`.
+
+The scalar fields are pressure, shear viscosity, and second viscosity.
+-/
```
jstoobysmith, line 40: Likewise, should be in `FluidFlow`

Physlib/FluidDynamics/FluidFlow/Momentum.lean
FloWsnr: They are actually used in Cauchy Flow as well, but we can remove them, I am not super attached to them :D

FloWsnr:
@jstoobysmith Sorry for the long cycles. I applied your comments.

jstoobysmith:
Would it be possible to include the motivation behind the different data structures within the files: 

e.g. for `FluidFlow` etc. why they have the fields that they do, why is that enough, why are non redundent etc. As well as motivation for why we have split things up into these data structures.

Physlib/FluidDynamics/FluidFlow/Momentum.lean
jstoobysmith: I think we should remove them. 

`LHS` isn't really a good name for something, as it is non-physical and depends on how someone writes down an equation. It is much better to package things in definitions which actually have physical meaning, and then to prove things about those.

FloWsnr: Okay, done!

FloWsnr:
@jstoobysmith Added documentation and removed the LHS stuff

jstoobysmith:
I think we are getting there :). Would you mind fixing the merge-conflicts.

jstoobysmith (approved):
Approved - will merge shortly. Thanks for iterating with me on this.

FloWsnr:
Awesome, thanks for all the help!

PR #1127, opened 2026-05-26
feat(QuantumMechanics): Add `LinearPMap.compRestricted`

gloges:
The composition of two partial linear maps, `g.comp f`, requires a proof that the range of `f` is contained in the domain of `g`. This PR defines `g.compRestricted f` (`g ∘ᵣ f`) which automatically restricts the domain of `f` to those `x` for which `g (f x)` is defined.
Also includes:
- Some lemmas on the domain of `g ∘ᵣ f` and the relationship between `g ∘ᵣ f` and other compositions of `g` with `f`.
- The inequality `U† ∘ᵣ V† ≤ (V ∘ᵣ U)†`. (The same inequality but with `∘ᵣ` replaced by any composition on either side follows immediately from this, `comp_le_compRestricted` and `adjoint_antitone`.)
- If `T` is symmetric then so too is `T ∘ᵣ T`.
- For multiplication operators, `𝓜 f ∘ᵣ 𝓜 g ≤ 𝓜 (f • g)` with equality when `𝓜 g` is bounded.

### DIFF
--- a/Physlib/QuantumMechanics/DDimensions/Operators/Unbounded.lean
+++ b/Physlib/QuantumMechanics/DDimensions/Operators/Unbounded.lean
@@ -510,6 +510,14 @@ lemma add_adjoint_isSymmetric [CompleteSpace H] (h : T.HasDenseDomain) :
   simp only [add_apply, inner_add_left, inner_add_right, h₁, h₂]
   exact add_comm _ _
 
+lemma IsSymmetric.comp_self (h : T.IsSymmetric) : (T ∘ᵣ T).IsSymmetric := by
+  intro x y
+  have hTx := mem_domain_of_mem_compRestricted_domain x
+  have hTy := mem_domain_of_mem_compRestricted_domain y
+  trans ⟪T ⟨x, x.2.2⟩, T ⟨y, y.2.2⟩⟫_ℂ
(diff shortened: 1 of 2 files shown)


gloges:
t-quantum-mechanics

Physlib/QuantumMechanics/DDimensions/Operators/Unbounded.lean
```diff
@@ -366,6 +431,20 @@ lemma adjoint_add_le_add_adjoint [CompleteSpace H]
       adjoint_isFormalAdjoint h₁ ⟨u, u.2.1⟩ ⟨w, w.2.1⟩,
       adjoint_isFormalAdjoint h₂ ⟨u, u.2.2⟩ ⟨w, w.2.2⟩]
 
+lemma adjoint_comp_le_comp_adjoint [CompleteSpace H] [CompleteSpace H']
```
jstoobysmith: should this be `compRestricted` not `comp` in the name?

gloges: Thanks, I've fixed this and three other lemma names.

jstoobysmith (approved):
Looks good to me. Will merge once linters finish

PR #1128, opened 2026-05-27
feat: EM Potential from E and B field

jstoobysmith:
The construction of an EM potential from an E and B field. The proofs that the potential gives back the corresponding E and B field.

Needed to build up API a bit for this to work.

### DIFF
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -364,7 +364,7 @@ lemma contDiff_ofElectricMagneticField {n : ℕ} (c : SpeedOfLight)
   have h1 : ContDiff ℝ (n + 1) ↿A := by
     simp only [WithLp.equiv_apply, A]
     apply ContDiff.neg
-    apply contDiff_intervalIntegral_of_contDiff
+    apply contDiff_parametric_intervalIntegral_of_contDiff
     refine contDiff_euclidean.mpr ?_
     intro i
     let C : (Time × Space) × ℝ → EuclideanSpace ℝ (Fin 3) := fun p =>
@@ -390,7 +390,7 @@ lemma contDiff_ofElectricMagneticField {n : ℕ} (c : SpeedOfLight)
     apply ContDiff.div _ (by fun_prop) (by simp)
     apply timeSlice_symm_contDiff
     apply ContDiff.neg
-    apply contDiff_intervalIntegral_of_contDiff
+    apply contDiff_parametric_intervalIntegral_of_contDiff
     suffices h : ContDiff ℝ n (fun (x : (Time × Space) × ℝ) => ⟪E x.1.1 (x.2 • x.1.2) + ∂ₜ (A · (x.2 • x.1.2)) x.1.1, Space.basis.repr x.1.2⟫_ℝ) by
       convert h using 1
       ext x
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -198,8 +198,9 @@ noncomputable def ofElectricMagneticField (c : SpeedOfLight)
     - ∫ u in (0 : ℝ)..1, ⟪E t (u • x) + ∂ₜ (A · (u • x)) t, basis.repr x⟫_ℝ ∂(volume)
   ofPotentials c φ A
 
-TODO "Prove differentiability conditions with respect to the constructors of
-  the electromagnetic potential."
+TODO "The scalar potential of `ofElectricMagneticField` can be defined as
+  `- ∫ u in (0 : ℝ)..1, ⟪E t (u • x), basis.repr x⟫_ℝ ∂(volume)`. This change should be made.
+  it works because the vector potential is zero when dotted with the position vector."
 
 TODO "Write lemmas for the various properties (e.g. the electric field) of
   the electromagnetic potential from the various constructors."
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -294,6 +294,68 @@ TODO "Add results related to the differentiability of the
 
 /-!
 
+### A.5. Differentiablity in terms of constructors
+
+-/
+
+lemma differentiable_ofScalarPotential {d} (c : SpeedOfLight) (φ : Time → Space d → ℝ)
+    (hϕ : Differentiable ℝ ↿φ) : Differentiable ℝ (ofScalarPotential c φ) := by
+  simp [ofScalarPotential]
+  rw [← SpaceTime.differentiable_vector]
+  intro μ
+  match μ with
+  | Sum.inl 0 => fun_prop
+  | Sum.inr _ => fun_prop
+
+lemma contDiff_ofScalarPotential {n} {d} (c : SpeedOfLight) (φ : Time → Space d → ℝ)
+    (hϕ : ContDiff ℝ n ↿φ) : ContDiff ℝ n (ofScalarPotential c φ) := by
+  simp [ofScalarPotential]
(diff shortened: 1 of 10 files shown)


wock9000 (approved):
Tested locally — full build passes (8,771 jobs), sorry_lint clean, runPhyslibLinters clean, codespell clean on all new files.

Checked the mathematical structure: `ElectromagneticPotential` as a function from spacetime to contravariant Lorentz vectors, decomposition into scalar and vector potentials (`ofPotentials_eq_add`), static potential specialization, Lorentz group action with `MulAction` instance, and differentiability of components. The E and B field constructions in `ElectricField.lean` and `MagneticField.lean` look correct — E from time derivative of A and gradient of φ, B from curl of A.

The `ParametricIntegration.lean` helper for differentiating under the integral sign is a useful utility.

Module documentation is thorough with clear overview, key results, and TODO items.

Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -186,21 +186,16 @@ lemma ofStaticPotentials_eq_ofPotentials {d} (c : SpeedOfLight) (ϕ : Space d 
...
-/-- The electromagnetic potential from a static electric and a static magnetic field.
-  There is no canonical choice here, so this depends on choice. -/
-noncomputable def ofStaticElectricMagneticField (c : SpeedOfLight)
-    (E : Space 3 → EuclideanSpace ℝ (Fin 3))
-    (B : Space 3 → EuclideanSpace ℝ (Fin 3))
-    (hE : Differentiable ℝ E) (hB : ContDiff ℝ 1 B)
-    (E_curl : Space.curl E = 0) (B_div : Space.div B = 0) :
+open MeasureTheory Matrix Space InnerProductSpace Time in
+/-- The electromagnetic potential from an electric and a magnetic field. -/
+noncomputable def ofElectricMagneticField (c : SpeedOfLight)
```
zhikaip: just a comment here, I understand that this is constructing from separate `E` and `B`, just wondering is `ElectromagneticField` better here and throughout the file to avoid names like `ofElectricMagneticField_magneticField ` which is somewhat confusing, but I can see where the current name is coming from so either is fine

Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -186,21 +186,16 @@ lemma ofStaticPotentials_eq_ofPotentials {d} (c : SpeedOfLight) (ϕ : Space d 
...
-  have A : Space 3 → EuclideanSpace ℝ (Fin 3) :=
-    Classical.choose (Space.exists_curl_of_div_zero B hB B_div)
-  ofStaticPotentials c φ A
-
-TODO "Add a constructor of the electromagnetic potential from non-static electric and
-  magnetic fields."
+  let A := fun t (x : Space) => - ∫ u in 0..(1 : ℝ), (u • basis.repr x) ⨯ₑ₃ B t (u • x) ∂(volume)
+  let φ := fun t (x : Space) =>
+    - ∫ u in (0 : ℝ)..1, ⟪E t (u • x) + ∂ₜ (A · (u • x)) t, basis.repr x⟫_ℝ ∂(volume)
+  ofPotentials c φ A
```
zhikaip, line 198: Apparently this specific construction is known as `Poincare Gauge`, it's probably the natural one to construct given E and B, but maybe worth pointing out in the docstring that this is choosing a gauge.

see e.g. https://kirkmcd.princeton.edu/examples/gauge.pdf

Physlib/Electromagnetism/Kinematics/MagneticField.lean
```diff
@@ -6,6 +6,7 @@ Authors: Joseph Tooby-Smith
 module
 
 public import Physlib.Electromagnetism.Kinematics.ElectricField
+import all Physlib.SpaceAndTime.Space.Derivatives.Curl
```
zhikaip: is `import all` necessary? does this indicate some lemmas shouldn't be private or they're in the wrong file?

Physlib/Electromagnetism/Kinematics/ElectricField.lean
```diff
@@ -75,6 +75,70 @@ lemma electricField_eq {c : SpeedOfLight} (A : ElectromagneticPotential d) :
...
+  ext1 t
+  ext1 x
+  suffices h : E t x + ∂ₜ (fun t => (ofElectricMagneticField c E B).vectorPotential c t x) t =
+      - ∇ ((ofElectricMagneticField c E B).scalarPotential c t) x by
+    simp only [electricField]
+    grind
+  convert congrFun (eq_grad_integral_of_curl_zero (fun x => E t x +
+      ∂ₜ (fun t => (ofElectricMagneticField c E B).vectorPotential c t x) t) ?_ ?_) x
+  · simp [ofElectricMagneticField_scalarPotential]
+    erw [grad_neg]
```
zhikaip: does tagging `grad_neg` with `@[to_fun]` work? same for all other `erw`s

zhikaip (commented):
left some comments

Physlib/Electromagnetism/Kinematics/EMPotential.lean
jstoobysmith: Done

Physlib/Electromagnetism/Kinematics/MagneticField.lean
jstoobysmith: Fixed - changed the proof.

Physlib/Electromagnetism/Kinematics/ElectricField.lean
jstoobysmith: Nice - didn't know about this. Fixed

Physlib/Electromagnetism/Kinematics/EMPotential.lean
```diff
@@ -186,24 +186,21 @@ lemma ofStaticPotentials_eq_ofPotentials {d} (c : SpeedOfLight) (ϕ : Space d 
...
 
-TODO "Prove differentiability conditions with respect to the constructors of
-  the electromagnetic potential."
+TODO "The scalar potential of `ofElectromagneticField` can be defined as
```
zhikaip: Sorry I don't quite get what this refers to, is it not already in this form? Is it that removing the time derivative actually makes it more difficult or something else?

jstoobysmith: Yeah - it makes one of the proofs more difficult. I'll see if I've got time this afternoon to try and do this within this PR, and can then eliminate this TODO item.

jstoobysmith: Done this - now the definition matches the notes you gave above.

Physlib/Electromagnetism/Kinematics/VectorPotential.lean
```diff
@@ -146,6 +197,39 @@ lemma contDiff_vectorPotential_ofElectromagneticField {n : ℕ} (c : SpeedOfLigh
     simp [ofElectromagneticField_vectorPotential, A]
   fun_prop
 
+open InnerProductSpace
+lemma vectorPotential_inner_radial_eq_zero_ofElectromagneticField
```
zhikaip, line 201: I'm somewhat skeptical that this should be placed here, this is only a property of this specific construction (gauge). I would think VectorPotential should contain statements that are true of all VectorPotentials, but maybe it's fine for now and we can reorganise after we have more gauge stuff.

zhikaip (approved):
Thanks for the changes. I think some of the stuff is not generally true and should be moved into its own folder, but we can restructure later.

(btw Kirk M. McDonald's notes reference above seems like a good resource, maybe in the future we can formalise some of the content there.)

Physlib/Electromagnetism/Kinematics/VectorPotential.lean
jstoobysmith: Yeah sounds like a good plan

PR #1130, opened 2026-05-27
chore: bump to v4.30.0

zhikaip:
@jstoobysmith there's some category stuff in `Physlib.Relativity.Tensors.Color.Basic`, do you mind having a look?

I can continue tomorrow on the rest if you're busy, but also happy if you want to take over from here


jstoobysmith:
I think the PR which is breaking the category theory stuff is [#36613](https://github.com/leanprover-community/mathlib4/pull/36613)

jstoobysmith:
Maybe easier to remove category theory from tensors all together. Started a PR here #1131.

zhikaip:
worked through more changes, apart from usual deprecations, many of them are related to `NNReal.toReal` or other subtype casting issues.

Tried doing some import cleaning on quantuminfo and removing some `set_option backward.isDefEq.respectTransparency false in` along the way, but probably should do a more systematic cleanup later on

jstoobysmith:
I'm going to try and get #1131 tomorrow as have got a long flight - I've done most of the technical work there I think, but needs some more fixes and clean up.

zhikaip:
@jstoobysmith this should be all, could you please review the final commit: according to claude,


>  Lean v4.30.0 tightened a runtime rule. In InitAttr.lean, importModules with loadExts := true now calls runInitAttrs, which begins:
> 
> ```
> if !(← isInitializerExecutionEnabled) then
>   throw <| IO.userError "`enableInitializerExecution` must be run before calling `importModules (loadExts := true)`"
>```
> So any executable that imports Physlib's environment at runtime must first call enableInitializersExecution to opt in. The Physlib scripts were written before this requirement, so they crash. (Mathlib doesn't hit this because its runtime-linter scripts 

claude was also used to handle some linters, I've reviewed the changes but hopefully I haven't missed anything

jstoobysmith (approved):
This looks good to me - if the linters all work correctly, then it seems like your fix is good. I've approved, but there are a couple of linters to fix still (new ones it seems).

jstoobysmith:
Alternatively we can just turn the linter `defsWithUnderscore` off for now.

zhikaip:
I think it's worth fixing rather than just letting the technical debt to accumulate, conforming better to Mathlib style is good standardisation IMO

zhikaip:
sorry I changed my mind after investigating, probably requires a significant of metaprogramming to fix properly:

.lake\packages\mathlib\Mathlib\Tactic\Linter\Style.lean introduces a new `defsWithUnderscore` linter, 

The most problematic issue is `informal_lemma` which elaborates to a `def`, probably need to include `@[nolint defsWithUnderscore]`

_physlib-suffixed instances are also flagged, the linter identifies `_mathlib` ones but not physlib

there's also some meta declarations e.g. `semiformal_result`, `Sorryful_attr`, `Pseudo_attr `

one more funny one is the linter exempts `_1` and `_2` but not `_3` and `_4` in Physlib.Mathematics.DataStructures.FourTree.Basic

maybe worth turning off for now and have a dedicated PR to fix this linter issue separately

zhikaip:
@jstoobysmith please have a final check and feel free to merge if you think this is fine

zhikaip:
closes #1129

PR #1131, opened 2026-05-28
refactor: Remove category theory from Tensors

jstoobysmith:
Removing category theory from Tensors.


jstoobysmith:
There is one proof that I have turned into a sorry proof because I think it is worth building an API to prove it properly. Otherwise, everything now builds. Need to do clean up

jstoobysmith:
This should hopefully now build, and pass linters.

zhikaip (approved):
As long as nothing breaks I think we should just go ahead and merge for now, any potential issues will surface eventually and this should hopefully be an improvement anyway.

jstoobysmith:
Agreed - I think because of the need of the bump on this we should bump this now. Removing the cat theory is no doubt and improvement to the overall maintaiblity

PR #1132, opened 2026-05-28
feat(QuantumMechanics): Add operator monoid

gloges:
Introduces an instance of `Monoid` for `LinearPMap` using the restricted composition defined in #1127. This allows one to take powers of operators, a prerequisite for defining the exponential map (e.g. for QHO coherent states).

Also adds a few basic lemmas about `compRestricted` and operator powers.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good - approved. Will merge shortly.

PR #1135, opened 2026-05-31
refactor: fix dimension of distDiv_inv_pow_eq_dim and remove several transparency options

zhikaip:
`distDiv_inv_pow_eq_dim` carries the wrong spatial dimensions which confuses me when I was looking at the point charge examples, hopefully this change makes sense.


jstoobysmith (approved):
Great! Approved .

PR #1134, opened 2026-05-28
feat(QuantumInfo): Cauchy-Schwarz norm bound for Bargmann invariant

wock9000:
## Summary

- Add `Braket.norm_dot_le_one` to `Braket.lean`: the bra-ket product of normalized states satisfies `‖⟨ψ₁|ψ₂⟩‖ ≤ 1`, proved via Cauchy-Schwarz
- Add `norm_bargmannInvariantThree_le_one` to `BargmannInvariant.lean`: the Bargmann invariant satisfies `‖Δ₃‖ ≤ 1`, as a product of three unit-bounded factors

## Approach

The proof connects `Braket.dot` to the `EuclideanSpace ℂ d` inner product via private helpers in `Braket.lean`, then applies Mathlib's `norm_inner_le_norm` (Cauchy-Schwarz on unit vectors). The Bargmann bound follows by factoring `‖a · b · c‖ = ‖a‖ · ‖b‖ · ‖c‖` and bounding each factor.

Per review feedback, `norm_dot_le_one` and the `EuclideanSpace` bridge live in `Braket.lean` (where the bra-ket product is defined), not in `BargmannInvariant.lean`. The `PiL2` import is already available transitively via `ForMathlib.LinearEquiv`.

## Motivation

The norm bound is needed for connecting the Bargmann invariant to the solid angle on the Bloch sphere (Pancharatnam's theorem), where `|Δ₃| = cos(Ω/4)` requires `|Δ₃| ≤ 1`.

## Build

Rebased onto master (includes #1133 and #1124 reorganization). Zero warnings, zero errors on `lake build`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1139, opened 2026-06-02
feat(QuantumInfo): angle-parameterized qubit ket for Pancharatnam connection

wock9000:
## Summary

Define `qubitKet α θ` representing `cos(α/2)|0⟩ + e^{iθ}sin(α/2)|1⟩` — the standard angle-parameterized qubit state mapping to `blochPoint α θ` on the Bloch sphere.

Prove `dot_qubitKet`: the inner product of two such states in closed half-angle form,

`〈ψ₁‖ψ₂〉 = cos(α₁/2)cos(α₂/2) + e^{i(θ₂-θ₁)}sin(α₁/2)sin(α₂/2)`

which is the computational core for connecting the Bargmann invariant to the solid angle (Pancharatnam's theorem).

## New definitions and results

| Name | Kind | Description |
|------|------|-------------|
| `qubitKet` | `def` | Qubit ket from Bloch sphere angles (α, θ) |
| `qubitKet_apply_zero` | `@[simp] lemma` | Component 0: `↑(cos(α/2))` |
| `qubitKet_apply_one` | `@[simp] lemma` | Component 1: `exp(iθ) · ↑(sin(α/2))` |
| `dot_qubitKet` | `lemma` | Inner product in half-angle form |

## Design notes

- Uses `Ket (Fin 2)` directly rather than importing `Qubit` (which pulls in the full Channels stack). The parameterization is identical to `blochPoint` from `BlochSphere.lean`.
- The raw vector is factored into a `private def qubitKetVec` to prevent `simp [qubitKet]` from unfolding through the normalization proof term.
- Single import: `BlochSphere`, which transitively provides `Braket`, `BargmannInvariant`, and Mathlib's complex analysis.

## Build

74 lines, zero warnings from this file, zero errors. Tested on `v4.29.1` after rebasing onto current master (includes #1134).

## References

S. Pancharatnam, *Generalized theory of interference, and its applications*, Proc. Indian Acad. Sci. A **44**, 247–262 (1956)

### DIFF
--- a/QuantumInfo/States/Pure/Pancharatnam.lean
+++ b/QuantumInfo/States/Pure/Pancharatnam.lean
@@ -1,0 +1,74 @@
+/-
+Copyright (c) 2026 Anand Nambakam. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Anand Nambakam
+-/
+module
+
+public import QuantumInfo.States.Pure.BlochSphere
+
+/-!
+# Pancharatnam Connection
+
+The angle-parameterized qubit ket `qubitKet α θ` represents the quantum
+state `cos(α/2)|0⟩ + e^{iθ}sin(α/2)|1⟩`, mapping to the Bloch sphere
+at polar angle `α` and azimuthal angle `θ`.
+
+## Important definitions
+ * `qubitKet`: qubit state from Bloch sphere angles
+
+## Important results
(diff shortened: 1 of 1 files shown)


QuantumInfo/States/Pure/Pancharatnam.lean
jstoobysmith, line 1: I think the stuff in this file is maybe more appropriate for the Braket or Qubit file

PR #1140, opened 2026-06-02
Add coordinate-level jet points

juanjfndz:
## Summary

This PR is the next step in a planned local Classical Field Theory development for PhysLib, aimed
at a formalization of the local Euler--Lagrange criterion appearing in Cortes--Haupt, Chapter 5.

The previous PRs introduced:
- reusable multi-index infrastructure,
- iterated spatial derivatives,
- and bounded derivative indices.

This PR begins the jet-coordinate layer by introducing the core coordinate-level jet point object.

## What This PR Adds

This PR adds `Physlib/ClassicalFieldTheory/Local/JetPoint.lean`, together with its import in
`Physlib.lean`.

Concretely, it provides:
- `JetCoordinates d m k`,
- `JetPoint d m k`,
- coordinate and value projections,
- basic coordinate extensionality and base-coordinate conversions,
- `jetCoordinatesAt`,
- and `jetAt`.

## Interpretation

The object introduced here is the coordinate-level point of the locally trivialized `k`-jet bundle
for fields `Space d -> EuclideanSpace ℝ (Fin m)`.

It formalizes only the local coordinate model
`Jet^k(Omega, R^m) ~= Omega x V`; it does not attempt to define global jet bundles.

The zero-th order field value is not stored separately in `JetPoint`; it is recovered as the zero
derivative coordinate.

The jet-coordinate data is kept in scalar component form,
`DerivativeIndex d k -> Fin m -> ℝ`, because the local Euler--Lagrange formulas are indexed by
both the derivative index `I` and the field component `a`.

## Reference

The reference for this local development is:

- Vicente Cortes and Alexander S. Haupt, *Lecture Notes on Mathematical Methods of Classical
  Physics*, arXiv:1612.03100v2, Chapter 5.

arXiv link: https://arxiv.org/abs/1612.03100

## Scope Of This PR

This PR is intentionally limited to the core jet-point object and the basic `jetAt` evaluation map.

It does not yet add:
- jet fiber-direction data,
- affine-line operations in the jet fiber,
- smoothness or support lemmas for jet-coordinate maps,
- total derivatives,
- or Lagrangians.

Those pieces are intended for follow-up PRs so that the core coordinate object can be reviewed on
its own.

## Build

Locally checked with:
- `python3 scripts/lint-style.py Physlib/ClassicalFieldTheory/Local/JetPoint.lean`
- `lake exe runPhyslibLinters Physlib/ClassicalFieldTheory/Local/JetPoint.lean`
- `lake build Physlib.ClassicalFieldTheory.Local.JetPoint`
- `lake build Physlib`

### DIFF
--- a/Physlib/ClassicalFieldTheory/Local/JetPoint.lean
+++ b/Physlib/ClassicalFieldTheory/Local/JetPoint.lean
@@ -13,7 +13,7 @@ public import Physlib.SpaceAndTime.Space.Derivatives.Iterated
 ## i. Overview
 
 This module introduces the coordinate-level point of the locally trivialized `k`-jet bundle for
-fields on `Space d` with values in `Space m`.
+fields on `Space d` with values in `EuclideanSpace ℝ (Fin m)`.
 
 This module only formalizes the local coordinate model `Jet^k(Ω, ℝ^m) ≃ Ω × V`; it does not
 attempt to define global jet bundles.
@@ -42,7 +42,7 @@ coordinate.
 ## iv. References
 
 - J. Cortés and A. Haupt, *Lecture Notes on Mathematical Methods of Classical Physics*,
-  Chapter 5, Section 5.1.
+  arXiv:1612.03100v2, Chapter 5, Section 5.1.
 
 -/
 
@@ -58,7 +58,11 @@ namespace Local
 
 -/
 
-/-- Coordinate data `u^a_I` for `k`-th order jet points. -/
+/-- Coordinate data `u^a_I` for `k`-th order jet points.
+
+For each derivative index `I`, these are the scalar components of the corresponding
+`EuclideanSpace ℝ (Fin m)` fiber coordinate. We keep this componentwise form because the
+local Euler-Lagrange formulas are indexed by both `I` and `a`. -/
 abbrev JetCoordinates (d m k : ℕ) := DerivativeIndex d k → Fin m → ℝ
 
 /-!
@@ -67,7 +71,7 @@ abbrev JetCoordinates (d m k : ℕ) := DerivativeIndex d k → Fin m → ℝ
 -/
 
 /-- The coordinate-level points of the locally trivialized `k`-jet bundle for fields
-`Space d → Space m`. -/
+`Space d → EuclideanSpace ℝ (Fin m)`. -/
 structure JetPoint (d m k : ℕ) where
   /-- The base point of the jet point. -/
   base : Space d
@@ -82,11 +86,13 @@ variable {d m k : ℕ}
 def coord (J : JetPoint d m k) (I : DerivativeIndex d k) (a : Fin m) : ℝ := J.fiber I a
 
 /-- The zero-th order value encoded by a jet point. -/
-def value (J : JetPoint d m k) : Space m := ⟨fun a => J.coord 0 a⟩
+def value (J : JetPoint d m k) : EuclideanSpace ℝ (Fin m) :=
+  WithLp.toLp 2 fun a => J.coord 0 a
 
 @[simp]
 lemma value_apply (J : JetPoint d m k) (a : Fin m) :
-    J.value a = J.coord 0 a := rfl
+    J.value a = J.coord 0 a := by
+  simp [value]
 
 @[ext]
 lemma ext (J K : JetPoint d m k) (hbase : J.base = K.base)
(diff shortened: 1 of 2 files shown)


Physlib/ClassicalFieldTheory/Local/JetPoint.lean
```diff
@@ -0,0 +1,191 @@
...
+abbrev JetCoordinates (d m k : ℕ) := DerivativeIndex d k → Fin m → ℝ
+
+/-!
+## B. Jet points
+
+-/
+
+/-- The coordinate-level points of the locally trivialized `k`-jet bundle for fields
+`Space d → Space m`. -/
+structure JetPoint (d m k : ℕ) where
```
jstoobysmith, line 75: Would change to to be for fields `Space d -> EuclideanSpace ℝ (Fin n)` or similar

Physlib/ClassicalFieldTheory/Local/JetPoint.lean
```diff
@@ -0,0 +1,191 @@
...
+
+-/
+
+/-- The jet-coordinate function determined by a field `g`. -/
+noncomputable def jetCoordinatesAt (k : ℕ) (g : Space d → Space m) (x : Space d) :
+    JetCoordinates d m k :=
+  fun I a => ∂^[I.1] (fun y => (g y) a) x
+
+/-- The coordinate-level `k`-jet point of a field `f` at the point `x`. -/
+noncomputable def jetAt (k : ℕ) (f : Space d → Space m) (x : Space d) : JetPoint d m k where
```
jstoobysmith: Would make this a map to `EuclideanSpace`

jstoobysmith: Likewise with the above.

Physlib/ClassicalFieldTheory/Local/JetPoint.lean
```diff
@@ -0,0 +1,191 @@
...
+
+## iii. Table of contents
+
+- A. Jet coordinates
+- B. Jet points
+- C. Jet points of fields
+
+## iv. References
+
+- J. Cortés and A. Haupt, *Lecture Notes on Mathematical Methods of Classical Physics*,
```
jstoobysmith, line 44: If this is AI generated, please can you make sure the reference is correct.

juanjfndz:
Thanks, I have pushed a correction addressing these points.

The field target in this PR is now `Space d → EuclideanSpace ℝ (Fin m)` rather than
`Space d → Space m`. This has been updated in the module overview, the `JetPoint` docstring,
`JetPoint.value`, `jetCoordinatesAt`, `jetAt`, and the zero-order value lemmas.

I kept `JetCoordinates d m k` in scalar component form,
`DerivativeIndex d k → Fin m → ℝ`, since this file is explicitly the local coordinate model with
coordinates `u^a_I`, and the later Euler--Lagrange coordinate formulas are indexed componentwise
by both `I` and `a`. I added a docstring clarification to make that interpretation explicit.

I also checked the reference. The source is Vicente Cortes and Alexander S. Haupt,
*Lecture Notes on Mathematical Methods of Classical Physics*, arXiv:1612.03100v2:
https://arxiv.org/abs/1612.03100. This is the paper/notes we are using as the reference for this
local development.

The scope of the PR is unchanged: it still only contains the core `JetPoint` object and `jetAt`;
fiber directions, affine line maps, regularity/support lemmas, total derivatives, Lagrangians, and
Euler--Lagrange material remain deferred to later PRs.

jstoobysmith:
I've asked @or4nge19 to also have a look at this PR - since he knows a bit about this area.

or4nge19:
I think it would be nice to know what is the plan (if any) for this set of PRs, so that they could be placed where they best fit, within the longer jet/variational ideal pipeline in physlib (currently in advanced stage of WIP).

Anyhow, the object here (and in the previous PRs) seems not yet a jet bundle or an intrinsic finite jet (mathlib's target, see [here](https://github.com/leanprover-community/sphere-eversion/blob/master/SphereEversion/Global/OneJetSec.lean) and [here](https://github.com/leanprover-community/sphere-eversion/blob/master/SphereEversion/Global/OneJetBundle.lean) and [here](https://github.com/leanprover-community/sphere-eversion/blob/master/SphereEversion/Global/TwistOneJetSec.lean) for early partial implementations) but the coordinate table of a flat trivial local model, obtained by applying the existing coordinate iterated-derivative operator. That is useful layer, for later Euler-Lagrange coordinate formulas among other things, as long as we do not treat `JetPoint d m k` as the public ontology of jets and have the idea, perhaps a TODO, to eventually derive or discharge most of the definitions (and lemmas/theorems) here from a 'deeper' (Fréchet/Taylor/contact) substrate API.

In particular, I would suggest adding to the module/docstring something along these lines:

- `JetPoint` is a coordinate readout for the trivial bundle `Space d × EuclideanSpace ℝ (Fin m)`.
- Equality of these objects is equality of chosen coordinate data, not a germ/contact equivalence.
- Coordinate Euler-Lagrange formulas should eventually be derived from Fréchet/Taylor/contact data, not used as the foundational definition.

Also, since the target is now `EuclideanSpace`, you may want to consider adding `coordVector` so higher-order coordinates have the same bundled readout as value. Something like:

```
def JetPoint.coordVector (J : JetPoint d m k) (I : DerivativeIndex d k) :
    EuclideanSpace ℝ (Fin m)
```

Physlib/ClassicalFieldTheory/Local/JetPoint.lean
```diff
@@ -0,0 +1,202 @@
...
+namespace ClassicalFieldTheory
+namespace Local
+
+/-!
+## A. Jet coordinates
+
+-/
+
+/-- Coordinate data `u^a_I` for `k`-th order jet points.
+
```
or4nge19, line 62: Re-iterating on the message I'd add two comments/future tasks along these lines, or as TODOs
- Equality of these objects is equality of chosen coordinate data, not a germ/contact equivalence.
- Coordinate Euler-Lagrange formulas should eventually be derived from Fréchet/Taylor/contact data, not used as the foundational definition.

juanjfndz:
Thanks @jstoobysmith and @or4nge19 for the comments and context.

For the longer plan, I had outlined the local CFT stack in this Physlib Zulip thread:
https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/ClassicalFieldTheory/with/582986380

I am keeping that roadmap message updated as the project adapts, and I am also adjusting the
planned PR labels so that they do not suggest that this stack is defining the foundational ontology
of jets.

The reference I am using for this local development is Cortes--Haupt, *Lecture Notes on
Mathematical Methods of Classical Physics*, arXiv:1612.03100v2:
https://arxiv.org/abs/1612.03100

More specifically, the current stack is aimed at Chapter 5, Section 5.1:

- bounded derivative indices correspond to the multi-indices `I` with `|I| ≤ k` used in the local
  coordinates `u^a_I`;
- this PR is the coordinate readout of the local trivial model `Jet^k(Ω, R^m) ≃ Ω × V`;
- the next coordinate fiber/affine-line layer corresponds to the `j^k(f + εh)` calculation in the
  proof of Theorem 5.2;
- the regularity/support layer supplies the smoothness and compact-support infrastructure used in
  that variational calculation;
- total derivatives correspond to the operators `D_i`, `D_I`, and `(-D)^I`;
- the later Lagrangian/action and Euler--Lagrange layers correspond to equation (5.1), equations
  (5.2)/(5.3), and the proof of Theorem 5.2.

So I will update the module documentation/TODOs to make explicit that this is a coordinate readout
layer for the flat/trivial local model, not a jet bundle or intrinsic finite-jet construction. In
particular, equality here is equality of chosen coordinate data, not germ/contact equivalence, and
the coordinate Euler--Lagrange formulas should eventually be related to or derived from a deeper
Fréchet/Taylor/contact API.

I can also add the bundled `coordVector` projection you suggested.

If there is a better location or naming convention for this coordinate-readout layer within the
longer jet/variational pipeline, I am happy to adjust the namespace/file organization before
continuing with the follow-up PRs.

jstoobysmith:
@juanjfndz If it is easier, you are welcome to develop this in the new PhyslibAlpha - for which the review process will be quicker. (see more details here: https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/PhyslibAlpha/with/602490950)

juanjfndz:
Thanks, that sounds like a good fit for this stack.

Given the length of the local CFT development and the fact that this is currently a coordinate-first
formalization of the Cortes--Haupt setup, I am happy to continue it in `PhyslibAlpha` so that the
whole pipeline can be developed and reviewed there more quickly. Then, if some parts mature into
reusable maintained API, we can later move them back into `Physlib` through the stricter process.

juanjfndz:
I have now opened the continued development in `PhyslibAlpha` as #1189, following the discussion here. I am closing this core `Physlib` PR to avoid keeping two parallel reviews of the same CFT stack; if parts of the Alpha development later mature into reusable maintained API, we can move those back into `Physlib` through the stricter process.

PR #1141, opened 2026-06-02
feat(QuantumMechanics): Begin spectral theory of closed operators

gloges:
Begins developing the spectral theory for closed unbounded operators, following Ch 2 of the book by Schmüdgen ([DOI](https://doi.org/10.1007/978-94-007-4753-1)).

This PR defines the regularity domain (set of regular points) of a `LinearPMap` and proves some of its key properties, namely that it is an open subset of ℂ and its relationship to the invertibility and range of `T - z • 1`.
These will eventually feed into the definitions and properties of the resolvent and spectrum.


gloges:
t-quantum-mechanics

jstoobysmith (approved):

PR #1142, opened 2026-06-02
refactor: redefine distLaplacian as a LinearMap

zhikaip:
simple fix


jstoobysmith (approved):
Approved - will merge now.

PR #1144, opened 2026-06-02
refactor:  curl definition and golfing

zhikaip:
change def of curl such that it is not explicitly matching on coordinates. this relies on the fact that ` (i + 3 : Fin 3)` and `(i : Fin 3)` are rfl.

Golfed some proofs accordingly, there might be further room for golfing especially in the magneticField file with how the fieldstrengthmatrix is setup, but I think this is good enough for now.

(wanted to get this merged before version bump to avoid merge conflicts afterwards)


jstoobysmith (approved):
Nice - looks good to me! Thanks for this. Approved.

PR #1146, opened 2026-06-04
feat(SpaceAndTime/Space): Give Space a proper IsManifold instance

Bergschaf:


jstoobysmith:
@Bergschaf Let me know if you want this reviewed properly. If you want to use your manifold stuff in Mathlib now, the best way to do that is to include it here with the same name as it is in Mathlib, then when we bump, it will result in an error, to tell us to remove it added here.

PR #1150, opened 2026-06-06
chore: remove unnecessary transparency options

zhikaip:
ran `scripts/rm_set_option.py` which checks and removes unncessary `set_option backward.isDefEq.respectTransparency false in`


jstoobysmith (approved):
Great! Approved

PR #1159, opened 2026-06-08
feat: Add PhyslibAlpha

jstoobysmith:
Add the basic linters and file structure for PhyslibAlpha. Happy to change the name if there is a different consensus. 

See: 

https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/PhyslibOL/with/601169454


Timeroot (approved):
With the import linter I think this makes sense. And if the experiment goes poorly (too much slop, or detracts from the main library's contributions) then we can always get rid of it. 👍

PR #1162, opened 2026-06-09
feat(PhyslibAlpha): Surfaces

jstoobysmith:
Add some surfaces which were part of #942 to PhyslibAlpha. 

This includes the surface associated with a spherical shell and the surface associated with a ring.

### DIFF
--- a/PhyslibAlpha.lean
+++ b/PhyslibAlpha.lean
@@ -1,3 +1,5 @@
 module
 
 public import PhyslibAlpha.Basic
+public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Ring
+public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SphericalShell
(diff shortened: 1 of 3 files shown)


zhikaip (approved):
is the plan to approve as long as it builds? I do wonder if there's a way to encourage cleaning up and "upstreaming" to physlib

jstoobysmith:
@zhikaip No - I think the idea is that you look at it, and decide if 
1) Has a chance of being will be useful 
2) Is mainstream physics 
3) Looks vaguely reasonable in other respects. 

If it looks reasonable w.r.t. the above then I think it should just be approved. 

In terms of up-streaming, we can think of incentives; part of which I think will be related to the fact that upstreamed code will be maintained, and the other part been some way to associate credit with such activities. These are things that (given they are part of the same repo) we can implement more easily I think.

PR #1173, opened 2026-06-12
refactor: Add invariants theorem for Higgs potential

jstoobysmith:
Adding the theorem which says that the Higgs potential is the most general potential up to order 4 which  is invariant under the SM action.


morrison-daniel:
Can't give a full review yet, but I think some of this should be separated into other lemmas. For example, define `V'` and `p'` in general and then pretty much each of your have statements can be its own lemma.

jstoobysmith:
@morrison-daniel I did something similar to this. I have defined a more general notion of an `EffectivePotential` and then define properties of that, the proof is slightly different from what I had before, but hopefully it is now slightly more generalizable.

Physlib/Particles/StandardModel/HiggsBoson/EffectivePotential.lean
```diff
@@ -0,0 +1,249 @@
...
+
+lemma termOfMassDim_isInvariant {V : EffectivePotential} {n : ℕ} (h : HasMaxMassDimLE V n)
+    (m : ℕ) (hV : IsInvariant V) : IsInvariant (termOfMassDim V h m) := by
+  intro g φ
+  have hV (t : ℝ) := hV g (t • φ)
+  have h1 (t : ℝ) : ∑  m ∈ Finset.range (n + 1), t ^ m * (termOfMassDim V h m (g • φ) -
+      termOfMassDim V h m φ) = 0 := by
+    simp [mul_sub, ← apply_smul_eq_sum_termOfMassDim]
+    rw [smul_comm, hV, sub_eq_zero]
+  -- Proof produced using Claude from here.
```
morrison-daniel: I'd avoid having comments like this. If you want to come back to clean up later maybe change to a TODO

morrison-daniel (approved):
Looks good!

Physlib/Particles/StandardModel/HiggsBoson/EffectivePotential.lean
jstoobysmith: Remove it now

PR #1178, opened 2026-06-12
chore: update contributing links

jfindlay:
This could be fixed in multiple ways:
- This pull request is offered as an easy way to just fix the current state especially if these links are unlikely to change again
- physlib.io could dynamically reroute old paths (probably not the best option)
- Docs update dev/CI tooling that works out link graph implications and maintains correctness (in addition to other linters).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 # Contributing
 
 See:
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - many thanks. Will wait for linters to finish and then will merge.

PR #1182, opened 2026-06-13
feat: quantum harmonic oscillator

bjoernkjoshanssen:
I put this in "Physlib Alpha" since I haven't studied the overall fit with Physlib yet.


jstoobysmith:
@bjoernkjoshanssen This looks good - would need to pass the linters before we could merge though. Would also recommend putting it in ./PhyslibAlpha/QuantumMechanics/HarmonicOscillator.lean

bjoernkjoshanssen:
> @bjoernkjoshanssen This looks good - would need to pass the linters before we could merge though. Would also recommend putting it in ./PhyslibAlpha/QuantumMechanics/HarmonicOscillator.lean

Thanks! Does each linter run need approval to be started? Would be nice if that was automatic.

jstoobysmith:
@bjoernkjoshanssen It does for first-time contributors. In future PRs it will run automatically. It is also possible to run them locally (which is usually easier) details here: 

https://github.com/leanprover-community/physlib/blob/master/scripts/PhyslibAlpha/README.md

bjoernkjoshanssen:
> @bjoernkjoshanssen It does for first-time contributors. In future PRs it will run automatically. It is also possible to run them locally (which is usually easier) details here:
> 
> https://github.com/leanprover-community/physlib/blob/master/scripts/PhyslibAlpha/README.md

Ah, makes sense.

bjoernkjoshanssen:
(I got a linter error that `/.lake/build/lib/lean/PhyslibAlpha.olean` does not exist, but fixed it by running `lake build PhyslibAlpha`.)

PhyslibAlpha/QuantumMechanics/QuantumHarmonicOscillator.lean
```diff
@@ -0,0 +1,405 @@
+module
+/- Copyright 2026 Bjørn Kjos-Hanssen. -/
```
jstoobysmith: To help with the linters a bit: take a look at other PhysLib files for an example copyright statement

jstoobysmith (approved):
Approved - many thanks! I will merge shortly.

PR #1187, opened 2026-06-15
Ai policy and agent instructions

jfindlay:
- `AI-POLICY.md`: human-facing contract for AI-assisted contributions. Covers author obligations, review process rules, and enforcement consequences.
- `AGENTS.md`: instructions for AI agents working on Physlib. Covers content rules, proof structure guidance, the pre-finish checklist, PR scope, commit conventions, and the format for PR descriptions.

The two files cross-reference each other and both point to `docs/ReviewGuidelines.md`.

**Reviewer map**: There is no Lean code in this PR. Read `AI-POLICY.md` first (human obligations), then `AGENTS.md` (agent-facing rules). Check that the two files are mutually consistent and that together they cover the governance surface you expect.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,90 @@
+# AGENTS.md
+
+Instructions for AI agents contributing to Physlib. The human-facing policy and disclosure
+obligations are in [AI-POLICY.md](AI-POLICY.md); read it too. You produce the code, but the human
+author certifies it: a clean build proves you proved what was written, never that it is what they
+meant.
+
+Also obey [docs/ReviewGuidelines.md](docs/ReviewGuidelines.md) and the Mathlib [style
+guide](https://leanprover-community.github.io/contribute/style.html).
+
+## Content
+
+- Use `lemma`, not `theorem`, unless the result is well known in the physics literature.
+- No `axiom` declarations. No `sorry` in submitted work.
+- Do not add lemmas that are trivial rewrites of existing Mathlib or Physlib results, unless they
+  add genuine physics context.
+- Place results in the appropriate existing file; do not create new files without good reason.
+- Number sections `# A. ...`, `## A.1. ...`. See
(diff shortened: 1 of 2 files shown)


AI-POLICY.md
```diff
@@ -0,0 +1,66 @@
...
+
+2.1. You must verify any bibliographic references for correctness yourself: that the work exists,
+that the cited statement is accurate, and that page numbers are right. This must not be delegated to
+an AI.
+
+2.2. You must confirm your contribution satisfies the content and structure rules in
+[AGENTS.md](AGENTS.md) before opening the PR. The agent's compliance is your responsibility, not the
+reviewer's.
+
+## 3. PR description
```
jstoobysmith: I put these conditions in the Agents.md file, so maybe we should remove them here? I think the agent should be doing this, I don't think we should put this as an obligation on the human (as in general it would be a lot of work).

AI-POLICY.md
```diff
@@ -0,0 +1,31 @@
...
+
+1.7. The human author(s) are fully responsible for every line of the PR, however it was produced. A human must vouch that each definition, theorem statement, and proof step means what it claims. A clean Lean build proves what was written, but only the human can certify what was written is what they meant.
+
+## 2. Author obligations
+
+2.1. You must verify any bibliographic references for correctness yourself: that the work exists, that the cited statement is accurate, and that page numbers are right. This must not be delegated to an AI.
+
+2.2. You must confirm your contribution satisfies the content and structure rules in [AGENTS.md](AGENTS.md) before opening the PR. The agent's compliance is your responsibility, not the reviewer's.
+
+## 4. Review process
```
jstoobysmith: Needs changing to `3. ...`

AGENTS.md
doxtor6, line 13: Better to add following too:

Never use structure fields of type True, or theorems returning True.
Never use an existential statement ∃ x, ..., True (any number of quantifiers ending in True).
Never use a scope-level variable hypothesis that assumes the conclusion.

jstoobysmith (approved):
I've approved this and will merge it shortly

PR #1191, opened 2026-06-16
feat: Stinespring dilation

bjoernkjoshanssen:
Stinespring dilation theorem, finite-dimensional version.
One of the Top 100 Quantum Theorems from https://marcodavid.net/top100/

### DIFF
--- a/PhyslibAlpha.lean
+++ b/PhyslibAlpha.lean
@@ -20,3 +20,4 @@ public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Line
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Ring
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SphericalShell
 public import PhyslibAlpha.QuantumMechanics.QuantumHarmonicOscillator
+public import PhyslibAlpha.QuantumMechanics.StinespringDilation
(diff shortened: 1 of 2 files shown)


jstoobysmith:
This is my fault, but one of the linters for PhyslibAlpha is broken - meaning it is not properly checking whether or not the file is correctly imported into ./PhyslibAlpha.lean. I've prepared a fix at #1193. But in the meantime - please can add 
```
public import  PhyslibAlpha.QuantumMechanics.StinespringDilation
```
to `./PhyslibAlpha.lean`. This will end up triggering a bunch of other linters; primarily that a lot of your definitions don't have doc-strings.

bjoernkjoshanssen:
> This is my fault, but one of the linters for PhyslibAlpha is broken - meaning it is not properly checking whether or not the file is correctly imported into ./PhyslibAlpha.lean. I've prepared a fix at #1193. But in the meantime - please can add
> 
> ```
> public import  PhyslibAlpha.QuantumMechanics.StinespringDilation
> ```
> 
> to `./PhyslibAlpha.lean`. This will end up triggering a bunch of other linters; primarily that a lot of your definitions don't have doc-strings.

Sounds good, trying that! But I can't find a linter that picks up on the missing doc-strings, `runPhyslibAlphaLinters` has no comment about it.

jstoobysmith:
Ok, so what you have is technically not wrong, but your file here is missing a `@[expose] public section`. This means that every result in your file is 'private' and that your definitions don't need doc-strings. Adding: 

```
@[expose] public section
```
underneath your module doc-string will result in these been picked up in your linters. 

You can: 
1. either leave things as they are for now and I can merge,
2. or, add this line and fix the linters. 

Which would you rather do?

bjoernkjoshanssen:
(2) sounds best since it'll improve the document in a straightforward way.

jstoobysmith (approved):
Looks good to me now. Approved. Will merge shortly

PR #1194, opened 2026-06-17
feat(QuantumMechanics): Begin spectral theory of symmetric operators

gloges:
Begins the spectral theory for symmetric operators with some results on the regularity domain and numerical range.
- Adds some general results for behavior under negation / scalar multiplication.
- Defines `realNumericalRange` which is the numerical range of a symmetric operator as a subset of ℝ rather than of ℂ, allowing for the notion of upper/lower bounds.
- Proves that the regularity domain contains all complex numbers with nonzero imaginary part as well as `(-∞,m)` or `(m,∞)` when `m` is a lower/upper bound on the numerical range.
- Adds a characterization of when the regularity domain is connected.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good to me. Approved - will merge shortly.

Issue #854, opened 2025-12-29
API: Space


jstoobysmith:
# Space

## Key data structure

The key data structure for the Space Api is a structure with a single value corresponding to a map from Fin n to Real. This corresponds to flat space. It acts as part of the standard world volume for many problems in physics.

- [x] Is defined

## Need

This API is one of the most heavily used in PhysLean and acts as the backbone to many other problems.

## Requirements

The following are requirements of the API:

- [x] the instance of a vector space
- [x] basis vectors
- [x] derivatives of functions from space to normed spaces 
- [ ] derivatives of functions from space to manifolds
- [ ] the action of the group of rotations
- [x] translations
- [x] relation between Spaces of different dimensions
- [x] derivatives of distributions from Space to normed spaces
- [ ] derivatives of distributions from Space to manifolds 
- [x] The API shall contain a `VSub` instance to `EuclideanSpace`.
- [ ] The API shall contain an `NormAddTorsor` instance from `EuclideanSpace`, see the conversation [here](https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Space.20vs.20EuclideanSpace/with/575938877).

## Corresponding File System 

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/SpaceAndTime/Space

## Contributing to this API

This API is open to contributions from anyone. If you believe that the requirements outlined above need modification, feel free to comment here. If you make a PR which is related to this API, please link it to this issue.

## References 

See 
- https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Space.20vs.20EuclideanSpace/with/575938877
## Parent APIs 

#911

LibertasSpZ:
Status update after the EuclideanGroup/torsor additions landed on master:

- Items 10 & 11 — already implemented, suggest ticking. `VSub` instance at `Space/Basic.lean:177`, `NormedAddTorsor instance at Space/Basic.lean:225` (with `AddTorsor` at `:190`) in current `master` branch.
- Item 5 (action of the group of rotations) — addressed by PR #1196  (adds `MulAction (RotationGroup d) (Space d)` by restriction of the Euclidean action; in progress).
- Item 4 (derivatives of functions Space → manifolds) — in progress next.
[...]

PR #1196, opened 2026-06-17
The Euclidean Action and the Action of the Group of Rotations

LibertasSpZ:
"Addresses item 5 of Issue #854 (the action of the group of rotations).
- Adds `Physlib/SpaceAndTime/Space/Origin.lean` (new file):
-- `instance : Zero (Space d), Space.zero_val, Space.zero_apply` — the origin (the vector-space zero) and its projection lemmas, moved here from `Module.lean`.
-- `Space.chartEuclidean, Space.chartEuclidean_apply` — the standard chart `Space d ≃ᵃⁱ[ℝ] EuclideanSpace ℝ (Fin d), p ↦ p -ᵥ 0`.

- Adds `Physlib/SpaceAndTime/Space/EuclideanGroup/Action.lean` (new file):
-- `Space.origin`, `Space.origin_apply` — coordinate basepoint and its projection.
-- `instance : MulAction (EuclideanGroup d) (Space d)` — the Euclidean action.
-- `EuclideanGroup.smul_apply` — coordinate formula.
-- `EuclideanGroup.smul_vsub_smul` — displacements transform by the linear part.
-- `EuclideanGroup.dist_smul` — the action is by isometries.
-- `EuclideanGroup.rotation_smul_eq` — restriction to RotationGroup action (`r • p = ↑r • p`).
-- `EuclideanGroup.rotation_smul_origin` — rotations fix the `0`.
-- `EuclideanGroup.rotation_smul_vsub_origin` — rotations act about the origin `0` by their orthogonal part.
-- `EuclideanGroup.rotation_dist_smul` — rotations preserve distance.
-- `EuclideanGroup.chartEuclidean_smul` — the chart and the bridge to `toAffineIsometryMulEquiv` (optional; rest of file doesn't depend on it).

- `Physlib/SpaceAndTime/Space/Module.lean` (modified):
-- Removed `instance : Zero (Space d), Space.zero_val, Space.zero_apply` — moved to `Origin.lean`; Module now sources them via its `import …Origin`.

Reading order: Part 1 (Euclidean action) → Part 2 (rotation specialization) → Part 3 (optional affine-isometry bridge), top to bottom.

Substantive AI usage (document setup, some statements, filling in some proofs; Claude Opus 4.8). All AI-generated content is human-reviewed and believed correct. No bibliographic references.


Physlib/SpaceAndTime/Space/EuclideanGroup/Action.lean
```diff
@@ -0,0 +1,163 @@
...
+@[expose] public section
+
+variable {d : ℕ}
+
+namespace Space
+
+/-- The coordinate origin of `Space d`, used as the basepoint for the Euclidean action. -/
+def origin (d : ℕ) : Space d := ⟨0⟩
+
+@[simp] lemma origin_apply (d : ℕ) (i : Fin d) : (origin d) i = 0 := rfl
```
jstoobysmith: Maybe it is worth doing the following here: 

In `Space.Module` we have an instance of `Zero` on `Space`. This essentially defines the origin in the same way you have it defined here. However, I understand that we don't want to import the whole module structure on `Space` here. So should we make a file `.../Space/Origin.lean` which has that instance of a `zero` so we don't need to define it both here and there? 

In it we could also include `Space.chartEuclidean` etc.

LibertasSpZ: Thanks, Joseph.

- Factored `origin` and `chartEuclidean` into `Space/Origin.lean` (imported by `Space/EuclideanGroup/Action.lean`). 
- Kept the `Module`'s Zero instance and added `origin_eq_zero : Space.origin d = 0 (by rfl)` in `Module.lean`, so the two are pinned definitionally equal. 
- Updated the inventory above.

Happy to further edit if you'd prefer. Thanks!

jstoobysmith: I think I would get rid of `origin` completely. and in `./Space/Origin` move the instance of the `zero` and just used `zero` in the `Action.lean` file. The idea of the `zero` is that it is a choice of origin, so the two concepts here are the same so it doesn't make sense (to me) to have two separate definitions for it.

LibertasSpZ: That makes sense. Done — moved the `Zero` instance (and `zero_val/zero_apply`) into `Space/Origin.lean`, dropped the separate `origin`, and `Action.lean` now uses `0` directly. Module sources its zero from the `import Origin`."

Physlib/SpaceAndTime/Space/EuclideanGroup/Action.lean
```diff
@@ -0,0 +1,147 @@
...
+/-! ## Part 3: relation to the affine isometry action (optional bridge)
+
+`chartEuclidean_smul` records that, under the chart `Space.chartEuclidean` (`Space/Origin.lean`),
+`p ↦ p -ᵥ origin`, the Part 1 action is the transport of `toAffineIsometryMulEquiv`
+(`AffineGroup.lean`) from `EuclideanSpace` to `Space`. Nothing in Parts 1–2 depends on it. -/
+
+/-- **The unification bridge.** Under the standard chart, the Euclidean action on `Space d` is the
+transport of `toAffineIsometryMulEquiv` acting on `EuclideanSpace`:
+`chart (g • p) = (toAffineIsometryMulEquiv g) (chart p)`. -/
+lemma chartEuclidean_smul (g : EuclideanGroup d) (p : Space d) :
```
jstoobysmith, line 136: Would remove the **The unification bridge** from this, but otherwise this looks good.

jstoobysmith (approved):
Approved - this now looks good to me.

Issue #943, opened 2026-02-13
API: Spatial Surfaces


jstoobysmith:
## Key data structure

This API is not directed around a single key data structure, instead it is centred around the concept of a surface in space. For each type of surface is defined around an embedding into `Space d`.

## Need

This API is needed for a number of reasons, common examples involve any system where distributions in space are important for example in Electromagnetism. 

## Requirements

For each of the following surface we require that the API has the following requirements:
- The API shall contain a map corresponding to the embedding of the surface into `Space d`. 
- The API shall contain a measure on `Space d` corresponding to integration over the surface. 
- The API shall contain a proof that the above measure has tempered growth.
- The API shall contain a distribution corresponding to the surface. 
- The API shall where applicable contain a proof that the surface is of measure zero. 

The surfaces these apply to are: 

- [ ] The spherical shell 
- [ ] The line 
- [ ] The solid sphere 
- [ ] The spherical cylinder 
- [ ] The solid cylinder 
- [ ] The thin ring 
- [ ] The half-plane 

## Example PRs 

Example PRs in this direction include #942. 

## Corresponding file system

There is as of yet no file system for this API, but we see it sitting in `./SpaceAndTime/Space/Surfaces/`

## Parent APIs 

#854

Robby955:
Is this still needed?

jstoobysmith:
@Robby955 Yes - there a lot of this still needs doing. There are some parts of this in 

#942 (never merged) 
and #1162 (merged). But a more complete coverage would be great.

PR #1198, opened 2026-06-17
feat: add solid sphere surface

Robby955:
Adds the solid sphere (closed unit ball) to the spatial surfaces API in PhyslibAlpha, following the merged `Line.lean`: the embedding into `Space d.succ`, the associated measure (ambient volume restricted to the closed unit ball), the temperate-growth instance, and the distribution. A solid region has positive ambient volume, so the measure-zero requirement does not apply and is replaced by a positive-volume statement. Part of #943.


jstoobysmith (approved):
Looks good - approved. Will merge shortly.

jstoobysmith (approved):
Looks good to me.

PR #1199, opened 2026-06-17
feat(space): add spherical cylinder surface

Robby955:
## Summary
Adds the spherical cylinder surface API under `PhyslibAlpha.SpaceAndTime.Space.Surfaces`.

## Changes
- Defines the spherical cylinder embedding into `Space 3` as the unit circular shell in `Space 2` transported through `slice 2`, extruded along `ℝ`.
- Adds injectivity, continuity, and measurable embedding lemmas.
- Adds the induced surface measure, tempered-growth instance, and `SFinite` instance.
- Adds the corresponding distribution and integral application lemmas.

## Notes
- This intentionally does not add `IsFiniteMeasure`: the cylinder is unbounded along the axis.
- No measure-zero theorem: issue #943 asks for that only where applicable, and the sibling surface files do not expose analogous null-range theorems.

### DIFF
--- a/PhyslibAlpha.lean
+++ b/PhyslibAlpha.lean
@@ -3,5 +3,6 @@ module
 public import PhyslibAlpha.Basic
 public import PhyslibAlpha.ClassicalFieldTheory.Local.FirstVariation
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Ring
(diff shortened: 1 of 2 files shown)


jstoobysmith (approved):
Looks good.

PR #1202, opened 2026-06-18
refactor: Split LinearPMap from Unbounded

gloges:
Splits off general results about `LinearPMap` from `Operators/Unbounded.lean` (sections which do not rely on an inner product structure) into the mathematics folder.
The individual commits walk through the split and subsequent clean-up in a way that I hope is easy to follow.


jstoobysmith (approved):
This seems a sensible separation - approved.

PR #1204, opened 2026-06-18
feat: Add PhyslibAlpha to TODO items

jstoobysmith:
Made using Claude Opus 4.8.  Though the changes here are minor. 

Adds PhyslibAlpha to the TODO item collection, making sure that they appear in the TODO list.

### DIFF
--- a/scripts/MetaPrograms/TODO_to_yml.lean
+++ b/scripts/MetaPrograms/TODO_to_yml.lean
@@ -111,33 +111,37 @@ instance : ToString PhyslibCategory where
   toString := PhyslibCategory.string
 
 def PhyslibCategory.ofFileName (n : Name) : PhyslibCategory :=
-  if n.toString.startsWith "Physlib.ClassicalMechanics"  then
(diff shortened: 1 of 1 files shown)


morrison-daniel:
I think it would be good to separate PhyslibAlpha TODOs from the rest. Probably making one category for all of PhyslibAlpha is sufficient since the idea of PhyslibAlpha is to be low impact on maintenance and I foresee fewer TODOs as stuff will be added "as is".

jstoobysmith:
@morrison-daniel Sounds good, will make this change now.

zhikaip (approved):
LGTM. but honestly before I looked at this I thought this would be saying upstreaming physlibalpha to physlib as a todo, maybe there should be a mechanism that we should also think about for a process to do so

jstoobysmith:
@zhikaip Yes! I agree. I think the right mechanism for doing this is to look for PhyslibAlpha files which do not depend on any other PhyslibAlpha files (so in a sense are 'next' to be upstreamed), then generate a TODO item on these.

PR #1206, opened 2026-06-18
feat: gamma anticommutator and slash of Lorentz vector

wdconinc:
This PR adds the gamma matrix anticommutation relation (γ^μ γ^ν + γ^ν γ^μ = 2 g^μν I_4) in the Dirac representation, and the slash operator on Lorentz vectors: /a = γ^μ a_μ. Introduce a combining character for slash notation (not used yet).

Rebased onto master after #1627, which derives `γ0` to `γ3` from the gamma endomorphisms as `endEquivMatrix (Fermion.Dirac.gamma …)` and proves the Clifford relations at the endomorphism level. That subsumes the re-indexing of `γ` to `Fin 1 ⊕ Fin 3` that this PR used to carry, and lets both remaining results be obtained by transporting the endomorphism API through `endEquivMatrix`. The PR is now purely additive.

# Reviewer map

Changes in `Physlib/Relativity/CliffordAlgebra.lean`, with two entries added to `Physlib/Relativity/Fermions/Dirac/API-map.yaml`.

## `γ_anticomm`

The anticommutator for the gamma matrices, indexed by `Fin 1 ⊕ Fin 3` against `minkowskiMatrix`. Since the gamma matrices are now the images of the gamma endomorphisms, this is the image of `Fermion.Dirac.gamma_anticomm` under the algebra equivalence `endEquivMatrix`, and no longer needs an explicit evaluation on matrix entries. It is named `γ_anticomm` rather than `gamma_anticomm` to avoid ambiguity with `Fermion.Dirac.gamma_anticomm`, which this file opens.

## `slash`

The definition of /a = γ^μ a_μ, as `endEquivMatrix (gammaMatrix k)`: the Dirac representation of Clifford multiplication by `k`. Note that `Fermion.Dirac.gammaMatrix` contracts the *lowered* vector coordinates, so this is the standard slash with signature (+,-,-,-); the earlier version of this PR summed `k^μ γ^μ` without the metric, which disagreed in the spatial components with the definition stated above. `slash_eq_sum` records the component form, and `slash_zero`, `slash_add`, `slash_smul` and `slashLinear` now follow from linearity of `gammaMatrix` rather than from case evaluation.

The slash API lives in `spaceTime.Slash`, since the `spaceTime.γ` namespace now holds the Clifford algebra material.

## `slashProd`

Unchanged: to assist with the common occurrence of multiple (typ. 4) slash momentum products, the `slashProd` definition takes a list of Lorentz vectors, slashes them, and gives the product (left to right). Also implemented via LinearMap.

# Future work

The next step here is to implement a theorem that extends `γ_anticomm` to `slash_anticomm`, which is now a short corollary of `Fermion.Dirac.gammaMatrix_anticomm`. The step after that is to add the trace identities on products of slash momenta.

The addition of the generalized Kronecker delta will allow for a Levi-Civita module, which can then be included here for the trace identities that have a gamma5.

# AI disclosure

The initial version was written with help of AI, but it was reworked a few times (e.g. making the `gamma_anticomm` theorem the basis instead of including a `slash_anticomm` theorem from scratch). The rebase onto #1627 and the rewrite of the proofs in terms of the gamma endomorphisms were done with Claude Opus 5.


Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +202,52 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
   exact ofCliffordAlgebra_range_eq_top
 
+/-! ### Dirac Slash Operators -/
+
+namespace Slash
+
+/-- Components of a Lorentz vector in the `γ0,γ1,γ2,γ3` ordering.
+
+Required cast since Lorentz.Vector is Fin 1 ⊕ Fin d → ℝ, not Fin 4 → ℂ. -/
+def coord (k : Lorentz.Vector 3) : Fin 4 → ℂ :=
```
jstoobysmith: You shouldn't actually need the `3` in `Lorentz.Vector 3` as it should default to that value.

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +202,52 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
+  ![(k (Sum.inl 0) : ℂ), (k (Sum.inr 0) : ℂ), (k (Sum.inr 1) : ℂ), (k (Sum.inr 2) : ℂ)]
```
jstoobysmith: I also think the solution to this is to change the definition of gamma so it is based on ` Fin 1 ⊕ Fin 3` rather then `Fin 4`, likewise with all the occurances of `Fin 4` below.

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +202,52 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
+
+/-- The Dirac slash of a Lorentz vector. -/
+def slash (k : Lorentz.Vector 3) : Matrix (Fin 4) (Fin 4) ℂ :=
```
jstoobysmith: The three below lemmas can be consolidated with lifting this to a linear map.

Physlib/Relativity/CliffordAlgebra.lean
wdconinc: I checked and no other internal module depends on this module or the gamma set in particular, so the risk in changing the definition is limited to external dependents. What is the policy for incompatible changes to public definitions? Should we also retain a Fin 4 version and mark it as deprecated?

jstoobysmith: I think for a change like this there would be no need to deprecate it (even for Mathlib I think), since we are updating the definition rather than replacing it. In this particular case I also think it will be pretty safe to just change it.

Physlib/Relativity/CliffordAlgebra.lean
wdconinc: With my C++ background, I misunderstood the default value here. I was thinking of it as a default template argument, when it is stronger than that. All 3s are gone now.

Physlib/Relativity/CliffordAlgebra.lean
wdconinc: Changed gamma set to Fin 1 ⊕ Fin 3, and propagated.

Physlib/Relativity/CliffordAlgebra.lean
wdconinc: There is now a linear map.

wdconinc:
-awaiting-author

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +232,67 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
+  fin_cases i <;> fin_cases j <;>
+    simp [complex_vector, Fintype.sum_sum_type, Fin.sum_univ_three] <;>
+    ring_nf
+
+/-- The Dirac slash as an `ℝ`-linear map. -/
+def slashLinear :
+    LinearMap (RingHom.id ℝ) Lorentz.Vector (Matrix (Fin 4) (Fin 4) ℂ) where
+  toFun := slash
+  map_add' := slash_add
+  map_smul' := slash_smul
```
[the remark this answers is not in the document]
wdconinc: Hmm, yeah, but searching for 'linear' or 'map' naturally did not bring up anything in the relativity code base.

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +232,68 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
+namespace Slash
+
+/-- The Dirac slash of a Lorentz vector. -/
+def slash (k : Lorentz.Vector) : Matrix (Fin 4) (Fin 4) ℂ :=
+  ∑ μ : Fin 1 ⊕ Fin 3, (k μ : ℂ) • γ μ
+/-- The notation for the Dirac slash operator, `k̸`. -/
+notation k "̸" => slash k
+
+/-- The componentwise complexification of a Lorentz vector in 1+3 dimensions. -/
+private def complex_vector (k : Lorentz.Vector) : Fin 1 ⊕ Fin 3 → ℂ :=
```
jstoobysmith: Should be in the Lorentz.Vector file.

jstoobysmith: Also e.g.  `toComplexVector`

Physlib/Relativity/CliffordAlgebra.lean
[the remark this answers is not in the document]
jstoobysmith: I would follow Copilots suggestion here, `→ₗ[ℝ]` is common for Linear maps

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +232,67 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
...
+def slashLinear :
+    LinearMap (RingHom.id ℝ) Lorentz.Vector (Matrix (Fin 4) (Fin 4) ℂ) where
+  toFun := slash
+  map_add' := slash_add
+  map_smul' := slash_smul
+
+/-- Left multiplication by a slash matrix as a linear endomorphism. -/
+def slashMulLeft (k : Lorentz.Vector) :
+    LinearMap (RingHom.id ℂ) (Matrix (Fin 4) (Fin 4) ℂ) (Matrix (Fin 4) (Fin 4) ℂ) :=
+  LinearMap.mulLeft ℂ (slash k)
```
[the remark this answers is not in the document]
jstoobysmith: Likewise here

jstoobysmith:
@wdconinc let me know when you want a review of this again :).

wdconinc:
> @wdconinc let me know when you want a review of this again :).

I'm at a conference this week (including talking to theorists in my field about Lean and Physlib), but I will take a final look when I'm back and mark as ready for review.

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -191,6 +232,63 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
   rw [← AlgHom.range_eq_top]
   exact ofCliffordAlgebra_range_eq_top
 
+/-! ### Dirac Slash Operators -/
+
+namespace Slash
+
+/-- The Dirac slash of a Lorentz vector. -/
+def slash (k : Lorentz.Vector) : Matrix (Fin 4) (Fin 4) ℂ :=
```
jstoobysmith, line 240: This may be difficult, but I think it might be possible to lift this to an intertwining map between representations.

wdconinc: Is your thinking here that this will then have an easier way to reuse the same structures for various spinor representations (Dirac and Weyl) without having this built up on the foundation of only a single gamma matrix definition? I think that would likely require some more work on generalizing the gamma matrix definition because the current form only works for Dirac spinors.

jstoobysmith: This wasn't my thinking, but this may be a benefit here. I was just thinking that `slash` really does represent a map between representations, and it might be nice to encode that somewhere. But this can wait till a future date. I think generalizing the gamma matrices would also be nice (making explicit that they carry e.g. `complexLorentzTensor` indices or similar).

jstoobysmith: (I think one of the things we would need to do before any of this is define and make a good API around `Fermion.dirac`, which we currently do not have.

jstoobysmith: As it stands, I'm happy to approve this PR now for merger.

wdconinc:
Rebased after #1627, which was a largely mechanical change in notation, not a substantial change in proofs.

wdconinc:
-awaiting-author

Physlib/Relativity/CliffordAlgebra.lean
```diff
@@ -208,5 +220,63 @@ theorem ofCliffordAlgebra_surjective : Function.Surjective ofCliffordAlgebra :=
 
 end γ
 
+/-! ## C. The Dirac slash operator -/
+
+namespace Slash
```
jstoobysmith: I think we should give this its own file, ./Dirac/Slash.lean? Otherwise this looks good to me.

jstoobysmith:
awaiting-author

for comment above

Physlib/Relativity/CliffordAlgebra.lean
wdconinc: You mean from Physlib/Relativity/CliffordAlgebra.lean to ./Physlib/Relativity/Fermions/Dirac/Slash.lean?

jstoobysmith: This is what I had in mind (but I got confused with where it was currently living), but maybe `Physlib/Relativity/DiracSlash.lean` would be more suited?

wdconinc: Ok, I'll move it there. That's still all inside the same relativity and not specific to fermions. I think Fermions/Dirac/Slash.lean is actually fine. I'll move it there.

wdconinc:
-awaiting-author

jstoobysmith (approved):
This looks good to me. I've approved. I will merge shortly.

PR #1209, opened 2026-06-19
feat(meta): add agent-review rubric and example GitHub Action

jfindlay:
* `AGENTS-REVIEW.md`: A vendor-neutral, rubric for a single review pass on a Physlib PR.
* `.github/workflows/agent-review.yml`: An example implementation triggered by the `agent-review` label.

### DIFF
--- a/.github/workflows/AgentReview.yml
+++ b/.github/workflows/AgentReview.yml
@@ -1,0 +1,157 @@
+name: Agent PR Review
+
+on:
+  pull_request_target:
+    types: [labeled]
+
+# One concurrent run per PR; cancel any in-flight run for the same PR when a new label fires.
+concurrency:
+  group: agent-review-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  agent-review:
+    name: Agent review pass
+    runs-on: ubuntu-latest
+
+    # Only run when the triggering label is `agent-review`.  Skip drafts and PRs marked
+    # `awaiting-author`.
+    if: |
+      github.event.label.name == 'agent-review' &&
+      github.event.pull_request.draft == false &&
+      !contains(github.event.pull_request.labels.*.name, 'awaiting-author')
+
+    steps:
+      - name: Checkout PR base
+        uses: actions/checkout@v4
+        with:
+          # Check out the base branch so the agent can search the repository.
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 0
+
+      - name: Fetch PR head
+        run: |
+          git fetch origin pull/${{ github.event.pull_request.number }}/head:pr-head
+
+      - name: Generate unified diff
+        run: |
+          git diff ${{ github.event.pull_request.base.sha }}...pr-head \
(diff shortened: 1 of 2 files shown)


.github/workflows/AgentReview.yml
```diff
@@ -0,0 +1,157 @@
...
+          cp AI-POLICY.md /tmp/ai-policy.md
+          cp docs/ReviewGuidelines.md /tmp/review-guidelines.md
+
+      - name: Run agent review
+        id: review
+        env:
+          AGENT_REVIEW_API_KEY: ${{ secrets.AGENT_REVIEW_API_KEY }}
+          # Configure the model here. Default: a mid-tier model.
+          # Escalate to a top-tier model only for large or complex PRs by exception.
+          AGENT_MODEL: "claude-sonnet-4-5"
```
doxtor6, line 59: sonnet 4.5 is not good enough. I think we should use the top tier models here like opus-4.8 or gpt-5.5

jstoobysmith: Do we know who actually provides the agents for this? I assume we would need some AI key. 

This might be a use case of #1211

jfindlay: > sonnet 4.5 is not good enough. I think we should use the top tier models here like opus-4.8 or gpt-5.5

Sonnet 4.5 is a compromise on token cost.  There are some other token optimizations in `AGENTS-REVIEW.md` as it's currently written, like only running the review on explicit (re)label.  I agree that some aspects of the review would be better at a higher tier agent, but more mechanical tasks like assessing docs coverage could be given to a lower tier agent to save cost.

> Do we know who actually provides the agents for this? I assume we would need some AI key.

Proprietary frontier agent models are less generous with their free tier than for example GitHub itself.  I do not know much about this area, but since this is a high profile, important, public project, there could be a  way to get a sponsorship or a discount or perhaps GitHub or some libre software consortium has a program for such.

doxtor6: Considering the current PR traffic of PhysLib, I think there is no need to compromise on token cost. I also generally do not trust models weaker than Opus-4.6; they sometimes produce serious hallucinations and require a lot of engineering effort to make them work properly.

I am still reading the GitHub agentic workflows documentation. It seems like it supports AI model subscription tiers, in which case a 100-200usd/month subscription should definitely be enough.

jstoobysmith: @jfindlay and @doxtor6 See the sort of AI-review on https://github.com/FormalFrontier/TauCeti/pull/283 (btw this is a project which has not been publicly announced yet, but which is meant to be an AI version of Mathlib so lots we could probably take from there). I think this sort of format would be nice. Would also be nice if people could volunteer to run the review process locally - which I think would help with costs.

doxtor6: I see. I agree that we can follow their procedure and format if it is working well. Perhaps we can ask Kim for help?

jstoobysmith: Not entirely sure about how they are going about doing the reviewing - if that is in house or whether they are using volunteer runs. But I think the format in which the review is presented is nice here.

doxtor6: Maybe we should have a quick chat with Kim about these.

PR #1213, opened 2026-06-19
feat: Representation on Lorentz Vectors

jstoobysmith:
The representation of the Lorentz group on Lorentz vectors. This is needed as part of a larger move to clean up Tensors, although it should be independently useful.

### DIFF
--- a/Physlib/Relativity/Tensors/RealTensor/CoVector/Basic.lean
+++ b/Physlib/Relativity/Tensors/RealTensor/CoVector/Basic.lean
@@ -50,6 +50,12 @@ def equivEuclid (d : ℕ) :
     CoVector d ≃ₗ[ℝ] EuclideanSpace ℝ (Fin 1 ⊕ Fin d) :=
   (WithLp.linearEquiv _ _ _).symm
 
+@[ext]
+lemma eq_of_apply_eq {d : ℕ} {v w : CoVector d} (h : ∀ i, v i = w i) : v = w := by
+  apply (equivEuclid d).injective
+  ext i
+  simpa using h i
+
 instance (d : ℕ) : Norm (CoVector d) where
   norm := fun v => ‖equivEuclid d v‖
 
@@ -115,6 +121,10 @@ lemma apply_add {d : ℕ} (v w : CoVector d) (i : Fin 1 ⊕ Fin d) :
 @[simp]
 lemma apply_sub {d : ℕ} (v w : CoVector d) (i : Fin 1 ⊕ Fin d) :
     (v - w) i = v i - w i := by rfl
+
+@[simp]
+lemma apply_sum {d : ℕ} {ι : Type} [Fintype ι] (f : ι → CoVector d) (i : Fin 1 ⊕ Fin d) :
+    (∑ j, f j) i = ∑ j, f j i := Finset.sum_apply i Finset.univ f
 
 @[simp]
 lemma neg_apply {d : ℕ} (v : CoVector d) (i : Fin 1 ⊕ Fin d) :
--- a/Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
+++ b/Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
@@ -1,0 +1,89 @@
+/-
+Copyright (c) 2026 Joseph Tooby-Smith. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joseph Tooby-Smith
+-/
+module
+
+public import Mathlib.RepresentationTheory.Basic
+public import Physlib.Relativity.LorentzGroup.Basic
+public import Physlib.Relativity.Tensors.RealTensor.CoVector.Basic
+/-!
+
+# Representation of the Lorentz group on Lorentz vectors
+
+In this module we define the representation of the Lorentz group on Lorentz covectors.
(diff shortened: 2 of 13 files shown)


Physlib/Relativity/Tensors/RealTensor/CoVector/Basic.lean
```diff
@@ -116,6 +122,10 @@ lemma apply_add {d : ℕ} (v w : CoVector d) (i : Fin 1 ⊕ Fin d) :
 lemma apply_sub {d : ℕ} (v w : CoVector d) (i : Fin 1 ⊕ Fin d) :
     (v - w) i = v i - w i := by rfl
 
+@[simp]
+lemma apply_sum {d : ℕ} {ι : Type} [Fintype ι] (f : ι → CoVector d) (i : Fin 1 ⊕ Fin d) :
+    (∑ j, f j) i = ∑ j, f j i := Finset.sum_apply i Finset.univ f
```
morrison-daniel, line 127: I feel this is redundant as it's just an application of `Finset.sum_apply`

Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
```diff
@@ -0,0 +1,89 @@
...
+-/
+
+@[expose] public section
+
+
+open Module
+open Matrix
+open MatrixGroups
+open Complex
+open TensorProduct
```
morrison-daniel:
```suggestion
open Module Matrix MatrixGroups Complex TensorProduct
```

Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
```diff
@@ -0,0 +1,89 @@
...
+
+noncomputable section
+
+namespace Lorentz
+
+namespace CoVector
+attribute [-simp] Fintype.sum_sum_type
+
+/-- The representation of the Lorentz group on Lorentz vectors. -/
+def rep (d : ℕ) : Representation ℝ (LorentzGroup d) (CoVector d) where
```
morrison-daniel:
```suggestion
def rep {d : ℕ} : Representation ℝ (LorentzGroup d) (CoVector d) where
```
Should be able to make this implicit

Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
```diff
@@ -0,0 +1,89 @@
...
+  toFun g := Matrix.toLinAlgEquiv basis (LorentzGroup.transpose g⁻¹)
```
morrison-daniel:
```suggestion
  toFun Λ := Matrix.toLinAlgEquiv basis (LorentzGroup.transpose Λ⁻¹)
```
I think it would be nice to make the notation here consistent with the rest of the section, same with map_mul' below.

Physlib/Relativity/Tensors/RealTensor/CoVector/Representation.lean
```diff
@@ -0,0 +1,89 @@
...
+  exact (LinearEquiv.eq_symm_apply (LinearMap.toMatrix basis basis)).mp rfl
+
+lemma rep_injective (d : ℕ) (Λ : LorentzGroup d) : Function.Injective (rep d Λ) := by
+  intro v1 v2 h
+  rw [rep_apply_eq_mulVec, rep_apply_eq_mulVec] at h
+  exact Matrix.mulVec_injective_of_isUnit (isUnit_of_invertible _) h
+
+lemma rep_surjective (d : ℕ) (Λ : LorentzGroup d) : Function.Surjective (rep d Λ) := by
+  intro v
+  use (LorentzGroup.transpose Λ⁻¹)⁻¹ *ᵥ v
```
morrison-daniel:
```suggestion
  use (LorentzGroup.transpose Λ) *ᵥ v
```
I see what you wrote is easier to prove by simp, but I think it's nicer to use the simpler form even if it takes an extra step to prove (likely could be fixed by adding appropriate simp lemmas). But this technically doesn't matter

Physlib/Relativity/Tensors/RealTensor/Vector/Basic.lean
```diff
@@ -54,6 +54,12 @@ def equivEuclid (d : ℕ) :
 lemma equivEuclid_apply (d : ℕ) (v : Vector d) (i : Fin 1 ⊕ Fin d) :
     equivEuclid d v i = v i := rfl
 
+@[ext]
```
morrison-daniel, line 57: I won't duplicate all my comments but same things go for both versions

morrison-daniel (changes requested):
One general comment that maybe we don't fix here because it's out of scope is I think it would help clean this up is to define `LorentzGroup` as a subtype (probably `Subgroup`) of `Matrix` because I think that would fix some casting problems and allow us to use dot notation. But that's maybe for another PR

Physlib/Relativity/Tensors/RealTensor/CoVector/Basic.lean
jstoobysmith: I think this is needed for the @[simp] to work.

morrison-daniel: Okay sounds good

morrison-daniel (approved):

PR #1214, opened 2026-06-20
Replace erw with simp only + rw in kineticTerm_eq_sum

Robby955:
Small cleanup in kineticTerm_eq_sum: the two erw calls are replaced by simp only [Tensorial.self_toTensor_apply] then a plain rw [coMetric_repr_apply_eq_minkowskiMatrix]


jstoobysmith (approved):
Approved - many thanks! This looks great

PR #1216, opened 2026-06-21
suggested change of def IsTotalTimeDerivative + proof of equivalence with the old one

abudjum:
Dear community,

I suggest to change the definition of IsTotalTimeDerivative in Physlib/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean

**Motivation**
 Invariance of the theory with respect to adding a total derviative  to the Lagrangian is quite a fundamental, eventhough rather trivial concept. In the current state, the file is incompatible with Physlib\ClassicalMechanics\EulerLagrange.lean (different order of arguments, \R instead of Time, explicitely written euclidean space instead of X), so first of all, I wanted to change that. Also, the usage of the uncurry function F in the definition of the derivative, but curry function for delta L looks odd, so I suggest to use the curry version both times (as in EulerLagrange.lean). Finally, in the current version it is only stated, but never formally proved that  IsTotalTimeDerivative \delta L indeed holds iff \delta L is a total derivative along a trjectory. Moreover, it looks much more natural to use this fact as a definition, and prove the explicit formula in the current version as a lemma. I had doubts if it would be better to separate this two changes (mechanical corrections for compatibility EulerLagrange.lean, and the new definition with the equivalence proof), but it seems odd to change the same definition twice in a row.  

**Technical issues**
I am aware of the linter's warning on the unused [CompleteSpace X] variable. For some reason, omit [CompleteSpace X]  does not work for me in front of a lemma that follows just after a definition (if I add some junk lemma in between, it works just as is should). I suspect a bug here. 

I did not change anything essential in lemma isTotalTimeDerivativeVelocity. Mostly I swap of the t and q variables, inl->inr, and \R->Time, and made sure that it works. So, actual contribution is about 120 lines

I changed r to q in the docstring, and v to d\_t q, again, to make it more consistent with EulerLagrange. However, later I've seen that v is actually used in other places, so I do not insist on that change.

**PS**
If you decide that the explicit formula, as in the current version, should be kept as the definition, I still think that the proof that it is really equivalent to the total derivative formula is worth adding, and the order of the variables should be changed


Physlib/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
```diff
@@ -56,31 +58,137 @@ This is because:
...
+/--
+    Explicit reformulation (by the chain rule):
+    δL(t, q, dₜ q) = ∂F/∂t(t, q) + ⟨∇ᵣF(t, q), dₜ q⟩
+
+    or
+
+    δL(t, q, dₜ q) = fderiv ℝ F (t, q) (1, dₜ q)
+-/
+-- omit [CompleteSpace X] in
+lemma IsTotalTimeDerivativeExplicit {δL : Time → X → X → ℝ} :
```
jstoobysmith: would rename this as `isTotalTimeDerivative_explicit` or more descriptive name, otherwise this looks good.

Physlib/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
```diff
@@ -56,31 +58,137 @@ This is because:
...
+      ∂ₜ qv t = fderiv ℝ (fun t' => (q₀ - t.val • v) + t'.val • v) t 1 := by rfl
+      _ = v := by
+        rw [fderiv_const_add,fderiv_smul_const]
+        simp only [ContinuousLinearMap.smulRight_apply, fderiv_val, one_smul]
+        fun_prop
+  rw [← h_qv_t, ← h_qv_der, hFder, ← h_F_tq_der]
+  rfl
+  exact hFdif
+  exact h_qv_contDiff
+  exact h_qv_contDiff
```
jstoobysmith: here I think you have multiple goals open at once. It is probably best to use the dot notation `\dot` to split them

abudjum:
Thank you for the feedback. I will think of better name (or use what yous
suggested) and add the dot notation where applicable. Also, I probably
should remove [CompleteSpace X] from the variable declaration, and add it
to the last lemma's arguments to avoid the warning.

Il Lun 22 Giu 2026, 06:21 Joseph Tooby-Smith ***@***.***> ha
scritto:

abudjum:
Sorry for the delay. Did as promissed in the comment above (did not come up with a better name though)

abudjum:
-awaiting-author

Physlib/ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean
```diff
@@ -56,31 +58,137 @@ This is because:
...
+
+/--
+    Explicit reformulation (by the chain rule):
+    δL(t, q, dₜ q) = ∂F/∂t(t, q) + ⟨∇ᵣF(t, q), dₜ q⟩
+
+    or
+
+    δL(t, q, dₜ q) = fderiv ℝ F (t, q) (1, dₜ q)
+-/
+lemma IsTotalTimeDerivative_explicit {δL : Time → X → X → ℝ} :
```
jstoobysmith: This should start with a lower case `i`. i.e. `isTotalTimeDerivative_explicit`. This is a convention in Lean, and just helps with consistency.

jstoobysmith (approved):
Great - Approved. Will merge shortly.

PR #1218, opened 2026-06-22
feat(QuantumMechanics): Begin spectral theory of self-adjoint operators

gloges:
Begins the spectral theory for self-adjoint unbounded operators with a few facts about the resolvent set and spectrum:
- Two lemmas which phrase that `z ∈ ρ T`, `z ∈ T.regularityDomain` and `(T - z • 1).range = ⊤` are all equivalent.
- `spectrum_real` : The full spectrum is real (a strengthening of the familiar fact that the _point_ spectrum is real for symmetric operators).
- `residualSpectrum_eq_empty` : The residual spectrum is trivial.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good - approved.

Issue #385, opened 2025-03-10
Remove `erw`s


jstoobysmith:
There are lots of `erw [...]` in PhysLean. These are slow, and thus it is beneficial for them to be removed. 

To remove an `erw [...]` usually means changing the corresponding lemma to simp-normal form.

jstoobysmith:
I think one of the worst files for this is:

- PhysLean.Relativity.Tensors.Color.Lift

Using `erw?` on some of the `erw`'s indicates that a lot have a similar origin.

PR #1227, opened 2026-06-23
refactor(QuantumMechanics): remove erw in momentumOperator_smul

Robby955:
Replace erw [deriv_smul] with rw [deriv_const_smul] in momentumOperator_smul. Since the scalar is a constant, deriv_const_smul applies directly and rewrites to simp-normal form, dropping the spurious zero term the old proof had to clean up. Addresses #385.


jstoobysmith (approved):
Great! Many thanks for this! I've approved. Will merge shortly

PR #1230, opened 2026-06-23
feat(space): add solid cylinder surface

### LINKED ISSUE #943: API: Spatial Surfaces

Robby955:
Adds the solid cylinder to the spatial surfaces API in PhyslibAlpha, the last remaining surface in #943. Following the merged `SphericalCylinder.lean` and `SolidSphere.lean`: the embedding of a cross-sectional disk extruded along the axis into `Space 3`, the associated measure (the solid-sphere measure on `Space 2` extruded along the line), its temperate-growth and s-finite instances, and the distribution. As the solid cylinder is a region of positive ambient volume the measure-zero requirement does not apply and is replaced by a positive-volume statement, mirroring the solid sphere. Closes #943.

### DIFF
--- a/PhyslibAlpha.lean
+++ b/PhyslibAlpha.lean
@@ -20,6 +20,7 @@ public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.HalfPlane
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Line
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.Ring
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SphericalCylinder
+public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SolidCylinder
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SolidSphere
 public import PhyslibAlpha.SpaceAndTime.Space.Surfaces.SphericalShell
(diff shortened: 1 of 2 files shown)


jstoobysmith (approved):
Looks good to me! Approved.

PR #1238, opened 2026-06-23
feat: Using WithDim

NicolaBernini:
# Overview

feat: Using WithDim


jstoobysmith (approved):
Great! Looks good - approved.

PR #1239, opened 2026-06-23
feat(QuantumMechanics): Add spectral measure

gloges:
Defines `SpectralMeasure` on a measurable space `α` as an extension of `VectorMeasure α (H →L[ℂ] H)` with the constraints that `μ A` is an orthogonal projection for each `A : Set α` and `μ Set.univ` is the identity.
Also includes a proof that `μ A ∘ μ B = μ (A ∩ B)` for all measurable sets `A` and `B` which follows from σ-additivity and properties of orthogonal projections.


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Looks good. Approved, will merge now.

PR #1240, opened 2026-06-23
refactor(IsMaximalSelfAdjoint): enable linters for QuantumInfo/ForMathlib/IsMaximalSelfAdjoint.lean

yidiq7:
Enables the Physlib linters for `QuantumInfo/ForMathlib/IsMaximalSelfAdjoint.lean` and removes it from `scripts/LinterExemption.txt`.

The file was previously exempt from linting. This PR updates the file to satisfy the two required linters (`lake exe runPhyslibLinters` and `./scripts/lint-style.sh`) and drops its exemption line so the rules are now enforced on it.

Verified locally: the build succeeds and both required linters pass on the file.

### DIFF
--- a/QuantumInfo/ForMathlib/IsMaximalSelfAdjoint.lean
+++ b/QuantumInfo/ForMathlib/IsMaximalSelfAdjoint.lean
@@ -7,18 +7,35 @@ module
 
 public import Mathlib.Analysis.Matrix.Normed
 
+/-!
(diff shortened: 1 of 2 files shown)


jstoobysmith (approved):
I've approved this. Will merge when it passes the linters. Many thanks :). There is also another script which you might be interested in `Scripts/auto-golf.sh` which helps golf proofs.

PR #1224, opened 2026-06-22
chore(QuantumInfo): document LinearEquiv and remove from linter exemption

tracyphasespace:
## What

Removes `QuantumInfo/ForMathlib/LinearEquiv.lean` from `scripts/LinterExemption.txt` by adding the documentation the linters require:

- a **module docstring** in the house format (`# Title` / `## i. Overview` / `## ii. Key results` / `## iii. Table of contents` / `## iv. References`), fixing the `ERR_MOD` from `scripts/lint-style.py`;
- **doc-strings** on the two flagged definitions, `LinearEquiv.of_relabel` and `LinearEquiv.euclidean_of_relabel`, fixing the `docBlame` errors from `runPhyslibLinters`.

No proof or definition bodies were changed — documentation only, plus the one-line exemption removal.

## Verification

With the file removed from the exemption list, it passes locally (Lean v4.30.0):

- `lake exe runPhyslibLinters` → `-- Linting passed for QuantumInfo.ForMathlib.LinearEquiv.`
- `scripts/lint-style.py QuantumInfo/ForMathlib/LinearEquiv.lean` → clean (exit 0)
- `lake exe module_doc_lint` → no longer flags this file

This is the first `QuantumInfo` file removed from the exemption list.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1244, opened 2026-06-24
chore(QuantumInfo): clear 56 files from the linter exemption list

tracyphasespace:
## What

Removes **59 of the 84** `QuantumInfo/` files from `scripts/LinterExemption.txt` by making each
pass `lake exe runPhyslibLinters` **and** `scripts/lint-style.py`. Changes are docstrings
(module + declaration/notation), line-wraps to ≤100 chars, resolved non-terminal `simp`, and a
small number of **safe** code-level fixes: removing `@[simp]` from lemmas the linter reports as
simp-provable, and `def`→`theorem` where flagged. No proof bodies were weakened.

Verified by a full `QuantumInfo` build (Lean v4.30.0): `runPhyslibLinters` reports
`-- Linting passed for QuantumInfo`, and `lint-style.py` is clean on every de-exempted file.

**Integrity:** the 59 de-exempted files add **no `sorry`, no `axiom`, no `admit`, no
`set_option … false`**, and contain no pre-existing `sorry` (files that still had `sorry` were
intentionally left exempted — see below). One disclosed mechanical choice: `Channels/Bundled.lean`
uses `attribute [nolint docBlame]` on 4 auto-generated multi-parent projections that cannot take
doc-strings without breaking projection inference.

This supersedes #1224 (that single-file change is included here).

## The 25 files left exempted (with suggested paths)

Documented so the cleanup isn't silently blocked:

- **6 exceed the 1500-line cap** (`Relative`, `LownerHeinzCore`, `SteinsLemma`, `Matrix`, `CFC`,
  `MState`) — wrapping only grows them past `ERR_NUM_LIN`; they need a **file split** or a
  `scripts/style-exceptions.txt` watermark.
- **2 don't build on baseline** (`Entropy/Axiomatized/Defs`, `ResourceTheory/ResourceTheory`) —
  pre-existing compile errors (look rename-induced); need a fix before linting.
- **8 have `unusedArguments` with multiple unused instance binders** — iterative signature
  surgery best done by someone who knows the intended API.
- **5 contain pre-existing `sorry`** (`Renyi` incl. data-processing inequality / Klein, `Capacity`,
  `ClassicalInfo/Capacity`, `Fidelity`, `Misc`) — genuine proof obligations.
- **4 are simply unfinished** (`Inner`, `FreeState`, `Qubit`, `Braket`) — tractable follow-up.


tracyphasespace:
Update: now **60** files de-exempted (added `ForMathlib/Misc.lean` — its only `sorry`s are inside a `/- -/` comment showing example future code; all real declarations are proved, confirmed via `#print axioms` = no `sorryAx`). **24** files remain exempted; the "contains `sorry`" group is now 4 (`Renyi`, `Capacity/Capacity`, `ClassicalInfo/Capacity`, `Fidelity`).

jstoobysmith:
Many thanks for doing this. 

Through no fault of yours, I'm struggling a bit with what to do with this PR, for following reason:
1. For me (or anyone else) to review this properly I would need a fair amount of time, in which other changes will no doubt happen and make some of the changes here redundant.  
2. I also don't think opening around 50 PRs for individual files in one go is a viable solution either. 

I think one of the ideas behind the ai-tool is that it produces a PR which is easy to verify/review/check in a very short amount of time, but when they are combined like this it makes it harder to verify. 

Maybe starting with 3-4 files would be a good start (either combined, or as separate PRs). 

Also a material comment: 

I think I know why it is doing this, but we should stop it putting the comment "This can be filled in later" for the table of content.

tracyphasespace:
Thanks — that's fair, and sorry for the slow reply.

Point taken on both counts: a ~56-file PR isn't reviewable in a window where it stays current, and
~50 single-file PRs would just move the cost rather than remove it. I've opened #1422 as the small
start you suggested:

- **4 stale exemptions** — those files already pass both linters with *no source changes*, so the
  diff is four deleted lines. Nothing to review beyond CI.
- **The `This can be filled in later` ToC placeholder** — `ForMathlib/LinearEquiv.lean` is the only
  file in `QuantumInfo/` still carrying it, and it got there via my #1224. #1422 replaces it with a
  real ToC in the `Capacity/Capacity.lean` house style, and the generator is fixed going forward.

Closing this one. The branch still has finished, CI-green fixes for ~50 files that remain exempted —
if it's useful to your `auto-task` tooling as a source to pull from at your own pace, help yourself;
otherwise ignore it.

One caveat if anyone does cherry-pick from it: this branch drops `@[simp]` from four lemmas. Two are
correct — `Prob.zero_le` and `CPTPMap.IsTracePreserving` are simp-provable, so the tag was redundant.
**Two are wrong**: `HermitianMat.ker_one` and `HermitianMat.support_one` are proved by `simp [ker]` /
`simp [support]`, so plain `simp` can't re-derive them and they should keep their `@[simp]`. I
verified each by hand. Neither is carried into #1422.

Happy to keep sending batches the size of #1422 if they're useful.

PR #1246, opened 2026-06-24
refactor(Thermodynamics): golf adiabatic_relation_log proof

yidiq7:
## Summary

Golfs the proof of `adiabatic_relation_log` in
`Physlib/Thermodynamics/IdealGas/Basic.lean`, replacing a 109-line, six-step
tactic block with an 11-line proof. The theorem statement (name, binders, type)
is unchanged — only the proof term was rewritten.

## What the proof shows

If two ideal-gas states have equal entropy at fixed `N`, then
`c * log (Ua / Ub) + log (Va / Vb) = 0`.

## How it was golfed

The original proof took six labelled steps, each introducing intermediate
`have`s: cancel `N * s0` via `add_left_cancel`, cancel `N * R` via
`mul_left_cancel₀`, cancel the shared `(c + 1) * log (N / N0)` term, rearrange
into a "difference = 0" form with several heavy `simpa [...]`/`field_simp`
calls, and finally re-fold the log differences back into `log (Ua / Ub)` and
`log (Va / Vb)`.

The key observation is that the difference of the two entropies is exactly
`N * R` times the goal. So the golfed proof:

1. **Length** — unfolds `entropy` in `hS` and uses `Real.log_div` to expand
   every `log (x / y)` into `log x - log y` (in both `hS` and the goal). Both
   sides then become linear in the atomic logarithms.
2. **Speed** — closes the algebra with a single `linear_combination hS` (a
   targeted `ring`-backed step) instead of a chain of `simpa`/`field_simp`
   rewrites with large simp-set arguments, then finishes with the pure term
   `(mul_eq_zero.mp key).resolve_left (mul_ne_zero hN.ne' hR.ne')` to extract
   the nonzero-factor cancellation — avoiding `mul_left_cancel₀`,
   `add_left_cancel`, `add_right_cancel`, and `sub_eq_zero` plumbing.
3. **Structure** — the proof now reads as its mathematical content: expand the
   logs, observe the entropy difference factors as `N * R * (goal)`, conclude.
   Uses idiomatic `hUa.ne'`/`mul_ne_zero` instead of `ne_of_gt`/`mul_ne_zero
   (ne_of_gt …)`.

## Verification

- `lake build Physlib.Thermodynamics.IdealGas.Basic` succeeds (2.9s),
  no errors or warnings.
- `#print axioms`-equivalent check shows only `propext`, `Classical.choice`,
  `Quot.sound` (the standard Mathlib axioms) — no `sorry`, no new axioms.

### DIFF
--- a/Physlib/Thermodynamics/IdealGas/Basic.lean
+++ b/Physlib/Thermodynamics/IdealGas/Basic.lean
@@ -55,115 +55,17 @@ theorem adiabatic_relation_log
       entropy c R s0 U0 V0 N0 Ua Va N =
       entropy c R s0 U0 V0 N0 Ub Vb N) :
     c * log (Ua / Ub) + log (Va / Vb) = 0 := by
-  -- Step 1: cancel `N * s0` and isolate the `N * R * (...)` pieces.
-  have h1 :
-      N * R *
-          (c * log (Ua / U0) +
-            log (Va / V0) -
-            (c + 1) * log (N / N0)) =
-        N * R *
-          (c * log (Ub / U0) +
-            log (Vb / V0) -
-            (c + 1) * log (N / N0)) := by
-    -- unfold entropy and use `add_left_cancel` to cancel `N * s0`
-    unfold entropy at hS
-    -- now `hS` is: N*s0 + N*R*(...) = N*s0 + N*R*(...)`
-    -- cancel `N*s0` from both sides
-    exact add_left_cancel hS
-
-  -- Step 2: cancel the common factor `N * R`.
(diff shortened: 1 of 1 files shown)


jstoobysmith:
Interestingly this is almost identical to #1234. Will close then and leave this one open. We should probably add something to the prompt of that script to randomise this a bit more.

yidiq7:
I also found the same issue #1247. I am adding some prompts asking Claude to check the current PR list first.

jstoobysmith (approved):
Approved. will merge now.

PR #1249, opened 2026-06-24
refactor(QuantumMechanics): golf orthogonal_exp_of_mem_orthogonal proof

yidiq7:
# Golf `orthogonal_exp_of_mem_orthogonal`

Golfs the proof of `orthogonal_exp_of_mem_orthogonal` in
`Physlib/QuantumMechanics/OneDimension/HarmonicOscillator/Completeness.lean`.

The lemma statement (name, binders, type) is unchanged — only the tactic block
is rewritten. The proof shrinks from **154 to ~115 lines** (`+18 / −68` in the
diff).

## What changed

**1. Pointwise-limit step (`h1`).**
Removed two helper rewrites: `h11` (manual `funext … simp [Finset.sum_mul]`) is
replaced by a single `simp_rw [← Finset.sum_mul]`, and `h12` (which rewrote an
expression to a syntactically identical one — a no-op) is deleted. Also dropped a
redundant `haveI : CauSeq.IsComplete ℂ norm` that is already found by instance
resolution.

**2. Measurability obligation of dominated convergence.**
The original spent ~28 lines reshaping the summand into a real polynomial and
invoking `guassian_integrable_polynomial` just to get
`AEStronglyMeasurable`. Since the only non-continuous factor is `f` (measurable
via `aeStronglyMeasurable_of_memHS hf`), this collapses to a one-liner: the
product of two `Continuous.aestronglyMeasurable (by fun_prop)` factors with the
measurable `f`.

**3. Vanishing-integral step (`h3b`).**
The two near-identical `funext … simp only [Complex.ofReal_*] … ring` rewrites
(one in `*` form, one in `•` form) are unified into a single reusable
`key (r : ℕ)` equation. The integrability side goal now uses
`(Q.mul_power_integrable …).const_mul _` instead of an extra `•`-shaped rewrite
plus `Integrable.smul`.

## Improvements

- **Length:** −50 net lines; eliminates a no-op rewrite and a duplicated
  `simp only [Complex.ofReal_*]` block.
- **Speed:** the measurability goal is closed by lightweight continuity /
  `fun_prop` reasoning rather than constructing a `Polynomial ℤ` and routing
  through `guassian_integrable_polynomial`.
- **Structure:** the repeated integrand reshaping is factored into one named
  `key` lemma; redundant intermediate `have`s are gone, making the
  dominated-convergence argument read top-to-bottom.

## Verification

- `lake build Physlib.QuantumMechanics.OneDimension.HarmonicOscillator.Completeness`
  succeeds with no errors or warnings.
- No `sorry`; axiom set is unchanged (`propext`, `Classical.choice`,
  `Quot.sound`).

### DIFF
--- a/Physlib/QuantumMechanics/OneDimension/HarmonicOscillator/Completeness.lean
+++ b/Physlib/QuantumMechanics/OneDimension/HarmonicOscillator/Completeness.lean
@@ -274,23 +274,9 @@ lemma orthogonal_exp_of_mem_orthogonal (f : ℝ → ℂ) (hf : MemHS f)
         (Complex.I * ↑c * ↑y) ^ r / r ! * (f y * Real.exp (- y^2 / (2 * Q.ξ^2))))
       Filter.atTop (nhds (Complex.exp (Complex.I * c * y) *
       (f y * Real.exp (- y^2 / (2 * Q.ξ^2))))) := by
-    have h11 : (fun n => ∑ r ∈ range n,
-        (Complex.I * ↑c * ↑y) ^ r / r !
-        * (f y * Real.exp (- y^2 / (2 * Q.ξ^2)))) =
-        fun n => (∑ r ∈ range n,
-        (Complex.I * ↑c * ↑y) ^ r / r !)
-        * ((f y * Real.exp (- y^2 / (2 * Q.ξ^2)))) := by
-      funext s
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - loots good.

PR #1217, opened 2026-06-21
feat(ClassicalMechanics): initial conditions from two positions

nateabr:
Resolves the "two positions at different times" item of the initial-conditions TODO in `HarmonicOscillator/Solution.lean` (the two-velocity item is left for a follow-up).

Adds:
- `InitialConditionsFromTwoPositions` (raw two-position data, S-independent)
- a total `toInitialConditions` conversion
- correctness lemmas `..._trajectory_at_t₁` / `_at_t₂`, under the nondegeneracy hypothesis `sin (S.ω * (t₂ - t₁)) ≠ 0`

Also completes the table of contents (previously missing A.2.1 / A.2.1.1 / D.1 entries) and corrects an incorrect section number A.2.1.2 → A.2.1.1. A one-word docstring typo fix is included as a separate commit.

Question for reviewers: correctness is stated under the nondegeneracy condition `sin (S.ω * (t₂ - t₁)) ≠ 0`. In the degenerate case the two measurements are dependent and do not determine the initial conditions (the classical result `x(t₂) = (-1)ⁿ · x(t₁)`). Does this need to be formalized as a lemma, or is the conditional statement sufficient?

Developed with AI assistance; all mathematical content and proofs were derived and verified by me.

PR #1250, opened 2026-06-24
feat(ClassicalMechanics): initial conditions from two velocities

nateabr:
## Summary

This adds another way to specify initial conditions for the harmonic oscillator:
from two velocity measurements taken at different times. It's the velocity
counterpart to the two-position version added in #1217, and follows the same
pattern.

The new `InitialConditionsFromTwoVelocities` structure just holds the raw data
(`t₁, v_t₁, t₂, v_t₂`), with a `toInitialConditions` conversion to the standard
form and correctness lemmas in a new section D.3. As in #1217, the conversion is
total and the nondegeneracy condition lives on the lemmas rather than in the
structure, so you can build the structure from any data but only recover the
initial conditions when `sin(ω(t₂ - t₁)) ≠ 0`.

## The math

It's a straightforward inversion of the velocity formula
`v(t) = -ω·sin(ωt)·x₀ + cos(ωt)·v₀` at the two times, which gives a 2×2 system
with determinant `ω·sin(ω(t₂ - t₁))`:

```
x₀ = (cos(ωt₂)·v_t₁ - cos(ωt₁)·v_t₂) / (ω·sin(ω(t₂ - t₁)))
v₀ = (sin(ωt₂)·v_t₁ - sin(ωt₁)·v_t₂) / sin(ω(t₂ - t₁))
```

## Docs cleanup

I also tidied up two documentation things in the same file while I was here:

- The uniqueness section (C.1) still described `trajectories_unique` as a TODO /
  "semiformal implementation", even though it's been fully proved. I rewrote the
  prose and docstring to actually describe the result (it goes through energy
  conservation). The theorem and proof are untouched.
- The arbitrary-time section (A.2.1) had its conversion broken out into its own
  A.2.1.1 subsection, which looked odd next to the two-position section where the
  conversion sits inline. I moved that derivation up into the A.2.1 preamble so all
  three sections are laid out the same way. Happy to pull this out into its own PR
  if you'd rather keep it separate, since it touches the arbitrary-time section.

## Housekeeping

The TOC, the A.2 implemented/future-work lists, and the in-file TODO are updated to
match. Amplitude-phase is still listed as future work.

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
@@ -457,15 +457,16 @@ lemma trajectory_equationOfMotion (IC : InitialConditions) :
 ### C.1. Uniqueness of the solutions
 
 We show that the trajectories are the unique solutions to the equation of motion
-for the given initial conditions. This is currently a TODO.
+for the given initial conditions.
 
 -/
 /-- The trajectories to the equation of motion for a given set of initial conditions
   are unique.
 
-  Semiformal implementation:
-  - One may needed the added condition of smoothness on `x` here.
-  - `EquationOfMotion` needs defining before this can be proved. -/
+  Given any smooth `x` satisfying the equation of motion with the same initial
+  position and velocity, the difference `y = x - IC.trajectory S` also solves the
+  equation of motion with zero initial conditions; energy conservation then forces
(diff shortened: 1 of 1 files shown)


nateabr:
t-classical-mechanics

jstoobysmith (approved):
Great. This looks good. I've approved and will merge shortly.

PR #1252, opened 2026-06-24
refactor(StatisticalMechanics): golf helmholtzFreeEnergy_eq_meanEnergy_sub_temp_mul_thermodynamicEntropy proof

yidiq7:
## Summary

Golfs the proof of `helmholtzFreeEnergy_eq_meanEnergy_sub_temp_mul_thermodynamicEntropy`
in `Physlib/StatisticalMechanics/CanonicalEnsemble/Lemmas.lean`.

The theorem statement (name, binders, type) is unchanged — only the tactic block was
rewritten. The identity proved is the fundamental thermodynamic relation `F = U − T·S`.

## What changed

The previous proof was ~70 lines built from three nested `calc` blocks, an auxiliary
`hS_form` rewrite of the entropy, and a final algebraic chain alternating `ring` with
several heavy `simp [...]` calls (each carrying a long list of `mul_comm`/`mul_assoc`/
`add_comm` lemmas).

All of that algebra is in fact a single linear identity once the three available facts
are substituted:

* `helmholtzFreeEnergy_def` / `partitionFunction_def` unfold `F = −kB·T·log(Z/hⁿ)`;
* `Real.log_div` + `Real.log_pow` split `log(Z/hⁿ) = log Z − n·log h`;
* `thermodynamicEntropy_eq_differentialEntropy_sub_correction` and
  `differentialEntropy_eq_kB_beta_meanEnergy_add_kB_log_mathZ` expand `S`.

After those rewrites the goal is a polynomial identity in the atoms `kB, T.val, T.β,
meanEnergy, log Z, dof, log h` modulo the single relation `T.val · (kB · T.β) = 1`
(from `kB_mul_beta`). It is discharged in one step with
`linear_combination 𝓒.meanEnergy T * hkβT`.

## How it improves the proof

* **Length** — 70 lines → 9 lines; the two intermediate lemmas (`hS_form`, `h_rhs`) and
  the `hF_rewrite` block are eliminated entirely.
* **Speed** — the chain of broad `simp [...]` calls (with large lemma lists) is replaced
  by targeted `rw`s plus one `ring`-backed `linear_combination`, which is cheaper to
  elaborate.
* **Structure** — the proof now reads as "unfold the definitions, expand the entropy,
  cancel `kB·β·T = 1`", matching the Mathlib idiom of `rw` to normalise followed by a
  single `linear_combination` for the residual algebra, instead of a hand-rolled
  multi-step `calc`.

## Verification

* `lake build Physlib.StatisticalMechanics.CanonicalEnsemble.Lemmas` succeeds with no
  errors or warnings.
* `lake build Physlib` succeeds (4234 jobs).
* `#print axioms` reports only `propext`, `Classical.choice`, `Quot.sound` — no `sorryAx`
  and no new axioms.


jstoobysmith:
I updated this script so that it can be run completely in auto mode, either with 
``` 
./auto-golf.sh --auto
```
other instructions in the Readme.

jstoobysmith (approved):
Looks good - I've approved.

jstoobysmith (approved):
Approved - thanks, will merge now.

PR #1257, opened 2026-06-24
refactor(TimeAndSpace): golf time_integral_iteratedFDeriv_apply proof

yidiq7:
## Summary

Golfs the proof of `Space.time_integral_iteratedFDeriv_apply` in
`Physlib/SpaceAndTime/TimeAndSpace/ConstantTimeDist.lean`.

Only the tactic block of this single lemma is changed; its statement
(name, binders, type) is byte-for-byte identical.

## What was wrong

The original proof closed five differentiability side-goals (across the
two nested inductions) with five near-identical copies of the same
8–10 line boilerplate block:

```lean
apply Differentiable.differentiableAt
apply ContDiff.differentiable_iteratedFDeriv (n := (n + 1 : ℕ))
refine Nat.cast_lt.mpr ?_
simp only [lt_add_iff_pos_right, zero_lt_one]
have hη := η.smooth'
apply ContDiff.comp
· exact hη.of_le (by exact ENat.LEInfty.out)
· fun_prop
```

All five goals are instances of just two facts:

* `Differentiable ℝ (iteratedFDeriv ℝ m ⇑η)`, and
* `Differentiable ℝ (iteratedFDeriv ℝ m (fun x => η (t, x)))`,

with the third/fifth occurrences only differing by post-composition
with `fun x => (t, x)`.

## How it was golfed

* **Length / structure:** the two facts are factored out once at the top
  of the proof as `∀ m`-quantified helper `have`s (`hη_diff`,
  `hη_diff'`). Because they are quantified over the iterated-derivative
  order `m` and depend only on the fixed parameter `η`, they survive both
  `induction n` blocks and can be reused in every occurrence. Each of the
  five repeated blocks collapses to a single `exact … .differentiableAt`
  (using `Differentiable.fun_comp` for the two composed goals). Net change:
  **16 insertions, 44 deletions (−28 lines)**, and the deduplication makes
  the recursion structure of the proof much easier to read.
* **Speed:** each occurrence no longer re-runs `apply ContDiff.comp`
  (which relied on higher-order unification of `⇑η` against `?g ∘ ?f`) nor
  the per-site `simp only`. The `ContDiff.differentiable_iteratedFDeriv`
  premise is discharged directly via `η.smooth'.of_le …` (and one
  `.comp (by fun_prop)` for the composed case), avoiding the redundant
  unification work.

## Verification

* `lake build Physlib.SpaceAndTime.TimeAndSpace.ConstantTimeDist` succeeds
  with no errors or warnings.
* No `sorry`; `#print axioms` for the lemma reports only the standard
  `propext`, `Classical.choice`, `Quot.sound` (no new axioms).


jstoobysmith (approved):
Approved - thanks. Will merge now. 

ready-to-merge

PR #1258, opened 2026-06-24
refactor(Relativity): golf prodT_toComplex proof

yidiq7:
## Summary

Golfs the proof of `prodT_toComplex` in
`Physlib/Relativity/Tensors/RealTensor/ToComplex.lean`.

The lemma states that `toComplex` of a tensor product equals the (complexified)
product of the `toComplex`s, and is proved by a double induction over the tensor
basis (outer over `t`, inner over `t1`).

## What changed

The statement (name, binders, type) is **unchanged** — only the tactic block is
rewritten.

The original proof set up the induction motives by hand: it `let`-bound a
predicate `P` (with an explicit `∀ t1, …`), proved `have hP : P t := by apply
induction_on_basis …`, did the same again for an inner predicate `P1`, and
finished with `exact hP t1`. Each non-basis case was wrapped in
`dsimp [P] … ; refine (by simp [...])`.

`induction_on_basis` is already a well-formed eliminator (implicit motive `P`,
major premise `t` last, conclusion `P t`), so the whole scaffold collapses into
the idiomatic `induction t using induction_on_basis with` form, nested for the
second tensor. Fixing `t1` instead of quantifying over it lets the inductive
hypotheses be used directly, removing every `let P` / `have hP` / `dsimp [P]` /
`refine (by …)` step.

## How it's golfed

- **Length**: ~84 lines → 16 lines. The two hand-rolled motive definitions, the
  two `apply induction_on_basis (c := …) (P := …) (t := …)` invocations, the
  `exact hP …` plumbing, and four `refine (by …)` wrappers are all gone.
- **Speed**: drops the per-case `dsimp [P]` / `dsimp [P1]` unfolding steps and
  the `refine` indirection; the zero/scalar/additive cases are now single `simp`
  calls. The (unavoidable) basis-case `simp` is kept verbatim.
- **Structure**: replaces the manual predicate/`apply`/`exact` pattern with
  named-case `induction … using … with`, matching Mathlib conventions and making
  the four basis-induction cases (`h`, `hzero`, `hsmul`, `hadd`) explicit and
  readable.

## Verification

- `lake build Physlib` succeeds (full build green).
- No `sorry`, no errors or warnings introduced.
- Axioms unchanged: `propext`, `Classical.choice`, `Quot.sound`.


jstoobysmith (approved):
Approved - will merge now.

PR #1259, opened 2026-06-24
auto-task(trajectories_unique): golf harmonic oscillator trajectory uniqueness proof

yidiq7:
## Summary

Golfs the proof of `ClassicalMechanics.HarmonicOscillator.trajectories_unique`
in `Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean`, reducing the
tactic body from **169 lines to 51 lines** with no change to the lemma
statement.

## What changed

Only the proof term of `trajectories_unique` was rewritten. The strategy is
unchanged: the difference `y := x - traj` is shown to be a smooth solution of
the equation of motion with vanishing initial data, and energy conservation
forces `y ≡ 0`. The cleanups:

- **Factored out the repeated derivative-of-a-difference computation.** The
  original proved `∂ₜ y` and `∂ₜ (∂ₜ y)` separately with ~30 lines of
  near-identical `DifferentiableAt`/`fderiv_fun_sub` boilerplate. These are now
  a single local `dsub` helper applied twice, with differentiability supplied
  directly by `ContDiff.differentiable` / `deriv_differentiable_of_contDiff`
  instead of being rederived step by step.

- **Collapsed Newton's-law / linearity bookkeeping.** The `calc` chain folding
  `force S (x t) - force S (traj t)` back into `force S (y t)` is now a single
  `simp [smul_sub, hNewt_x, hNewt_traj, hydef, force_eq_linear]`, and the
  equation of motion for `y` is obtained directly from
  `equationOfMotion_iff_newtons_2nd_law`.

- **Streamlined the energy/positivity endgame.** The ~50-line block deriving
  `y t = 0` from `energy y t = 0` is replaced by `add_eq_zero_iff_of_nonneg`
  (kinetic + potential are each nonnegative, so both vanish) followed by
  `mul_eq_zero` / `inner_self_eq_zero` to extract `⟪y t, y t⟫_ℝ = 0` and hence
  `y t = 0`.

The nonnegativity facts now use targeted `mul_nonneg` / `real_inner_self_nonneg`
terms rather than `unfold` + ad-hoc lemmas.

## Verification

- `lake build Physlib.ClassicalMechanics.HarmonicOscillator.Solution` succeeds.
- `lake build Physlib` succeeds (4234 jobs).
- No errors or warnings reported for the file; no `sorry` and no new axioms.


jstoobysmith (approved):
Looks good - approved.

PR #1260, opened 2026-06-24
auto-task(quantaWaveNumber_subset_brillouinZone): golf Brillouin-zone membership proof

yidiq7:
## Summary

Golfs the proof of `quantaWaveNumber_subset_brillouinZone` in
`Physlib/CondensedMatter/TightBindingChain/Basic.lean` from 74 lines down to 26,
with no change to the statement.

## What changed

The lemma shows that each quantized wavenumber
`2π/(a·N)·(n − ⌊N/2⌋)` lies in the Brillouin zone `[−π/a, π/a)`.

The original proof duplicated its whole setup (`generalize T.N`, `generalize T.a`,
`NeZero` reasoning) across the two `Set.Ico` bounds, and inside each bound it ran a
full even/odd case split on `N` (`Nat.even_or_odd'`) with bespoke `field_simp` /
`div_lt_iff₀` manipulations.

The key observation is that both bounds reduce, after factoring out the common
positive scalar `π/a`, to the linear facts

* `-1 ≤ 2·(n − ⌊N/2⌋)/N`  and
* `2·(n − ⌊N/2⌋)/N < 1`,

which in turn follow from the elementary nat bounds `2·⌊N/2⌋ ≤ N` and
`N ≤ 2·⌊N/2⌋ + 1` (both discharged by `omega`) together with `0 ≤ n < N`. No
even/odd case analysis is needed.

Concretely:

* The shared `generalize`/positivity setup is hoisted out of the two bullets and
  done once.
* The two `Nat.even_or_odd'` case-bashes are replaced by clearing the denominator
  (`le_div_iff₀` / `div_lt_one`) followed by `linarith` against the cast `omega`
  bounds.
* The high-level factoring scaffold (`le_of_eq_of_le` / `le_of_le_of_eq` /
  `mul_le_mul_of_nonneg_left`, and the `lt_…` analogues) is kept, since it is
  already idiomatic.

## Why

Shorter, faster to elaborate, and more readable: the proof now makes explicit that
the result is a single linear inequality about `⌊N/2⌋`, rather than two parallel
even/odd computations.

## Verification

* `lake build` completes successfully with the golfed proof in place.
* No `sorry`, no new axioms (only `propext`, `Classical.choice`, `Quot.sound`), and
  no new errors or warnings.
* The lemma statement (name, binders, type) is byte-for-byte unchanged.


jstoobysmith (approved):
lgtm - approved.

PR #1261, opened 2026-06-24
auto-task(pauliBasis_minkowskiMetric_pauliBasis'): golf Pauli basis Minkowski metric proof

doxtor6:
## Summary

Golfs the proof of `pauliBasis_minkowskiMetric_pauliBasis'` in
`Physlib/Relativity/PauliMatrices/SelfAdjoint.lean`.

The original proof did an explicit four-way `match` on `i : Fin 1 ⊕ Fin 3`.
Each of the three `Sum.inr` branches repeated a byte-for-byte identical 8-line
block: an unnecessary inner `cases i` (on a value already fixed by the outer
match) followed by `ext`/`simp_all`. Only the `Sum.inl 0` branch was different.

The entire case analysis collapses to a single `fin_cases i <;> simp [...]`,
unioning the simp lemma sets that were used across the branches
(`pauliSelfAdjoint'`, `pauliSelfAdjoint`, `pauliBasis`, `pauliBasis'`,
`minkowskiMatrix.inr_i_inr_i`, plus `Subtype.ext_iff`, `NegMemClass.coe_neg`,
`neg_neg` to discharge the subtype/negation equalities).

## Changes

- `Physlib/Relativity/PauliMatrices/SelfAdjoint.lean`: proof body 33 lines → 3
  lines. Removes the redundant inner `cases i` and the three duplicated branch
  blocks; no `match` boilerplate.
- The lemma statement (name, binders, type) is unchanged.

## Verification

- `lake build Physlib.Relativity.PauliMatrices.SelfAdjoint` succeeds.
- No `sorry`; axioms unchanged (`propext`, `Classical.choice`, `Quot.sound`
  only); no new errors or warnings.


jstoobysmith (approved):
looks good, thanks - approved.

PR #1271, opened 2026-06-25
auto-task(spectrum_subset_of_isOpen): flatten auto-generated proof scaffolding

mshibashish:
## Summary

Golfs the proof of `HermitianMat.spectrum_subset_of_isOpen` in
`QuantumInfo/ForMathlib/HermitianMat/CFC.lean`. Only the tactic block is
touched — the lemma statement (name, binders, type) is byte-for-byte
identical.

The original proof was machine-generated and carried a large amount of
scaffolding: **15** separate `generalize_proofs at *; (` invocations wrapping
nearly every step, together with a corresponding 5-level-deep tail of `)))))`.
Each `generalize_proofs at *` rescans the entire local context, but none of the
hypotheses here have proof terms embedded in their *types*, so every one of
these passes is a no-op.

## What changed

- **Removed all 15 redundant `generalize_proofs at *` passes** and the matching
  paren nesting, turning a deeply nested expression into a flat sequence of
  `have`/`obtain` steps. This is the bulk of the win on all three axes:
  - *speed* — no longer rescans the whole context 15 times during elaboration;
  - *length* — `47 insertions(+), 54 deletions(-)`;
  - *structure* — the proof now reads top-to-bottom instead of as a wall of
    `generalize_proofs at *; (` lines closed by a single `)))))`.
- Minor idiomatic cleanups, all behaviour-preserving:
  - dropped redundant `by exact` wrappers (`hK_compact`, `hK_disjoint`,
    `h_unitary_open`, `h_unit`) in favour of direct terms;
  - `spectrum_subset_closedBall B |> fun h => h.trans <| …` written as the
    direct `(spectrum_subset_closedBall B).trans …`;
  - `linarith!` → `linarith` (the goal is purely real-linear);
  - merged the `have := Metric.isOpen_iff.mp …` / `obtain … := this` pair into a
    single `obtain`;
  - rewrapped over-long lines and removed stray trailing `;`/`,`.

The mathematically load-bearing steps (the `simp_all`/`simpa` units argument,
the compactness subcover, the `min'` bound, the final `filter_upwards`) are kept
verbatim.

## Verification

- `lake build QuantumInfo.ForMathlib.HermitianMat.CFC` succeeds with no errors
  and no warnings.
- LSP diagnostics for the file are empty.
- `#print axioms` reports only `propext`, `Classical.choice`, `Quot.sound` — no
  `sorryAx`, no new axioms.

### DIFF
--- a/QuantumInfo/ForMathlib/HermitianMat/CFC.lean
+++ b/QuantumInfo/ForMathlib/HermitianMat/CFC.lean
@@ -640,75 +640,68 @@ set_option maxHeartbeats 400000 in
 lemma spectrum_subset_of_isOpen (A₀ : HermitianMat d ℂ) (U : Set ℝ)
     (hU : IsOpen U) (hAU : spectrum ℝ A₀.mat ⊆ U) :
     ∀ᶠ B in nhds A₀, spectrum ℝ B.mat ⊆ U := by
-  -- Let $M = \|A₀\| + 1$. For $B$ in a ball of radius $1$ around $A₀$, $\|B\| \leq M$, so $\sigma(B) \subseteq \overline{B(0, M)}$.
-  obtain ⟨M, hM⟩ : ∃ M : ℝ, ∀ B : HermitianMat d ℂ, ‖B - A₀‖ < 1 → spectrum ℝ B.mat ⊆ Metric.closedBall 0 M := by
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Many thanks for this PR and for running the script! This is very helpful. I've approved, and once the linters finish I will merge

PR #1272, opened 2026-06-25
feat(FluidDynamics): add pressure field API

Robby955:
This adds a small pressure-field layer to `FluidDynamics`, built on the existing `FluidState`, stress-tensor, and momentum-balance API. It is meant as a starting point for the pressure API discussed in #862; I kept it to the parts the current library supports cleanly, and can adjust the shape to fit the intended requirements.

### Added

`Physlib/FluidDynamics/Pressure.lean`:
- `PressureField`: a scalar pressure field.
- `FluidWithPressure`: a `FluidState` together with a pressure field.
- `pressureStress`: the isotropic pressure stress `-p • I`; `pressureStress_apply`, `pressureStress_offDiag`, and `pressureStress_symm` give its components and symmetry.
- `pressureForce`: the pressure force density `-∇p`, with `pressureForce_apply`.
- `matrixDiv_pressureStress`: the divergence of the pressure stress equals the pressure force, the bridge into the momentum balance.
- `toMomentumBalance`: packages a `FluidWithPressure` as momentum-balance data.

`Physlib.lean`: imports the new module.

### Reviewer map

`Pressure.lean` reads top to bottom: the field and `FluidWithPressure`, then the stress tensor and its lemmas, then the force and the `matrixDiv_pressureStress` bridge, and finally `toMomentumBalance`.

### DIFF
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -71,6 +71,7 @@ public import Physlib.FluidDynamics.FluidState
 public import Physlib.FluidDynamics.NavierStokes.Basic
 public import Physlib.FluidDynamics.NavierStokes.Continuity
 public import Physlib.FluidDynamics.NavierStokes.Momentum
+public import Physlib.FluidDynamics.Pressure
 public import Physlib.Mathematics.Calculus.AdjFDeriv
 public import Physlib.Mathematics.Calculus.Divergence
 public import Physlib.Mathematics.Calculus.ParametricIntegration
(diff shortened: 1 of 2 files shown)


jstoobysmith:
See the changes and discussion here #1125 related to setting up Fluid mechanics properly. Maybe if you want to go in this direction you can adapt some of the work of @FloWsnr there (adding them as a co-author on the relevant commits).

Robby955:
Yes sure, I will study #1125 and I am interested to see if my approach even makes sense with the discussion and @FloWsnr expert work.

Robby955:
I will read #1125 and @FloWsnr's fluid-mechanics setup and align the pressure layer with that direction rather than the standalone shape here.

Robby955:
closing this as I believe @FloWsnr approach already has this covered properly

PR #1274, opened 2026-06-25
feat(Relativity): add Levi-Civita Lorentz tensor and antisymmetry

Robby955:
Adds the rank-4 Levi-Civita tensor `εⁱʲᵏˡ` on the complex Lorentz tensor space and proves its antisymmetry under exchange of the first two indices.

### Added

- `leviCivita` (`Physlib/Relativity/Tensors/ComplexTensor/Metrics/Basic.lean:51`): the tensor `ℂT[.up, .up, .up, .up]` built through `ofRat`, sending a component map to `⟨1, 0⟩` on the 12 even permutations of `(0,1,2,3)`, to `⟨-1, 0⟩` on the 12 odd permutations, and to `⟨0, 0⟩` whenever any index repeats.
- `ε4` notation (`Metrics/Basic.lean:128`): scoped `complexLorentzTensor` notation for `leviCivita`, alongside the existing `εL` / `εR` metric notations in the same section.
- `leviCivita_eq_ofRat` (`Metrics/Basic.lean:455`): restates `ε4` as its explicit `ofRat` sign table, with proof `rfl`, so the table is definitionally the tensor and downstream component rewrites can target it directly.
- `leviCivita_antisymm` (`Metrics/Lemmas.lean:102`): the identity `{ε4 | μ ν ρ σ = - (ε4 | ν μ ρ σ)}ᵀ`, proved by reducing both sides to basis components, applying `← map_neg`, and closing the finite component case split with `decide`.

### Reviewer map

- The sign table and the notation sit next to the existing metric tensors (`dualLeftMetric`, `dualRightMetric`) and reuse their `ofRat` / `_eq_ofRat` shape, so the new declarations read against the surrounding file.
- `leviCivita_antisymm` follows `dualRightMetric_antisymm` in `Lemmas.lean`: the same `permT_basis_repr_symm_apply` then `ofRat_basis_repr_apply` then `decide` skeleton, with `set_option maxRecDepth 100000` for the wider case split over the four free indices.
- Both lemmas reduce to `decide`, so a reviewer can check the sign table component by component rather than trusting the by-hand permutation parities.

### Convention

The normalization here is all-up `ε^{0123} = +1`, totally antisymmetric, in the mostly-minus metric. If a different sign or a lowered-index height better fits the intended (2.23)-(2.30) identities of the reference note, the table can be flipped or the indices lowered to match, and the antisymmetry proof carries over unchanged.


Physlib/Relativity/Tensors/ComplexTensor/Metrics/Basic.lean
```diff
@@ -47,6 +47,59 @@ abbrev dualLeftMetric : ℂT[.downL, .downL] := complexLorentzTensor.metricTenso
 /-- The metric `ε_{dot a}_{dot a}` as a complex Lorentz tensor. -/
 abbrev dualRightMetric : ℂT[.downR, .downR] := complexLorentzTensor.metricTensor Color.downR
 
+/-- The Levi-Civita tensor `εⁱʲᵏˡ` as a complex Lorentz tensor, with `ε⁰¹²³ = 1`. -/
+noncomputable def leviCivita : ℂT[.up, .up, .up, .up] :=
```
jstoobysmith: This should likely go into its own file system. There may also be a more systematic way of defining it.

jstoobysmith:
I also think @wdconinc might have ideas about the Levi-Civita tensor

Robby955:
That makes sense. I will move leviCivita and its lemmas into their own file under the tensor metrics directory. For the full definition, I will wait for maintainers and @wdconinc's input before reworking it rather than lock in anything now.

Robby955:
I have moved leviCivita and its lemmas into their own file, Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean. On a more systematic definition, I will wait for @wdconinc's input before reworking it.

Robby955:
-awaiting-author

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
jstoobysmith, line 1: Maybe we should even put it in Tenors/LeviCivita/Basic.lean

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
```diff
@@ -0,0 +1,153 @@
...
+open TensorProduct
+noncomputable section
+
+namespace complexLorentzTensor
+open Fermion
+open TensorSpecies
+open Tensor
+
+/-- The Levi-Civita tensor `εⁱʲᵏˡ` as a complex Lorentz tensor, with `ε⁰¹²³ = 1`. -/
+noncomputable def leviCivita : ℂT[.up, .up, .up, .up] :=
```
jstoobysmith: I think the most important question here will be: "What is the best type for this". I think if we define it as a real tensor, we can define it in general dimensions.

wdconinc:
I'm away at a conference so I didn't see this earlier, but the Levi-Civita symbol and tensor was the reason I introduced the generalized Kronecker delta since it's the same object. Then the contractions follow from the generalized Kronecker delta contraction properties.

Robby955:
Thanks @wdconinc , and full credit to you for the generalized Kronecker delta. That makes sense, the Levi-Civita symbol is the same object as you pointed out.

 I'll rework this to define Levi-Civita through the generalized Kronecker delta so the contractions follow from its contraction properties, instead of duplicating it. If that all sounds good or wouldn't collide with your work of course.

wdconinc:
Pulled up my laptop now...

Based on the generalized Kronecker delta, I've got the symbol
```
abbrev leviCivita4Int (μ ν ρ σ : Fin 4) : ℤ :=
  generalizedKroneckerDelta 4 ![μ, ν, ρ, σ] id
```
and I define a covariant and contravariant tensor from this, definitionally.

The contraction identity for the Kronecker delta
```
∑ μ : Fin k → ι, δ^{(μ,lam)}_{(μ,ω)} = k! · δ^{lam}_{ω}
```
(with proof a bit annoying with many lemmas, but inductive based on Laplace's method for determinant).

Then this leads to proofs for `ε^{μνρσ} ε_{μντω} = -2 (g^ρ_τ g^σ_ω - g^ρ_ω g^σ_τ)`, `ε^{μνρσ} ε_{μνρτ} = -6 g^σ_τ`, and `ε^{μνρσ} ε_{μνρσ} = -24` where the magic integers are k!.

Robby955:
That's clearly cleaner through your generalized Kronecker delta, with the k! factors falling out of the contraction identity. Do you want to take it? I'll defer to you on the Kronecker delta and the contractions, since it's your object. If a thin Levi-Civita tensor and antisymmetry wrapper over your leviCivita4Int is useful, you should run with the whole thing. Full credit to you either way.

wdconinc:
> That's clearly cleaner through your generalized Kronecker delta, with the k! factors falling out of the contraction identity. Do you want to take it? I'll defer to you on the Kronecker delta and the contractions, since it's your object. If a thin Levi-Civita tensor and antisymmetry wrapper over your leviCivita4Int is useful, you should run with the whole thing. Full credit to you either way.

Either way. In very new to Lean so happy to defer. Even if GitHub commits or PRs counted as credit for tenure in my field, it's too late anyway for me to benefit from it :-D

I don't like my current proof of the generalized contraction. It seems support for minor sub matrices that are missing column,row = i,j should be easier to define and handle, or even a Laplace expansion should be an established pattern, but instead I'm kinda having to reinvent the wheel I feel, and I fear it's mostly due to my inexperience with Lean and/or mathlib.

Robby955:
Sounds good. I’ll try wiring the Levi-Civita tensor and antisymmetry directly off `leviCivita4Int`.

For the contraction, I think mathlib should save us from most of the manual determinant work. The useful lemmas look like:

`Matrix.det_succ_row` / `Matrix.det_succ_column` for the Laplace expansion, and
`det_eq_sum_row_mul_submatrix_succAbove_succAbove_det` in `Determinant/Misc.lean`, which writes the determinant as a row-sum of an entry times the delete-row/delete-column minor. That seems to me to match the submatrix we’re currently defining by hand.

For the `k!` factor, I’d probably use the permutation-sum form via `Matrix.det_apply`, then use `det_permute` to handle the sign cleanly.

wdconinc:
> `Matrix.det_succ_row` / `Matrix.det_succ_column` for the Laplace expansion, and
> `det_eq_sum_row_mul_submatrix_succAbove_succAbove_det` in `Determinant/Misc.lean`, which writes the determinant as a row-sum of an entry times the delete-row/delete-column minor. That seems to me to match the submatrix we’re currently defining by hand.

Yeah, that's the lemma I was clearly missing :-)

One thing that I've found hard in mathlib is searching for the useful patterns, whether as a physicist I'm used to calling it by a different name, or whether it's because in mathlib it is implemented a level of abstraction higher than I'm used to dealing with it, or sometimes because it's using a notation shortcut that is not quite as amenable to searching.

Robby955:
Reworked Levi-Civita through the generalized Kronecker delta: leviCivita is now ofRat (fun f => generalizedKroneckerDelta f id), and antisymmetry follows from a det row-swap (Matrix.det_permute) instead of the decide over 256 components.

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
```diff
@@ -0,0 +1,90 @@
...
+open Complex
+open TensorProduct
+noncomputable section
+
+namespace KroneckerDelta
+open Matrix
+
+/-- Swapping two of the upper indices of the generalized Kronecker delta negates it.
+This is one row transposition of the underlying determinant. -/
+lemma generalizedKroneckerDelta_swap {α ι : Type} [DecidableEq α] [DecidableEq ι] [Fintype ι]
```
wdconinc: Does it make more sense to put this in Kronecker.lean?

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
jstoobysmith: Think I would still make this move here.

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
```diff
@@ -0,0 +1,74 @@
+/-
+Copyright (c) 2024 Joseph Tooby-Smith. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joseph Tooby-Smith
```
jstoobysmith: Should not be me here.

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
```diff
@@ -0,0 +1,74 @@
...
+open Fermion
+open TensorSpecies
+open Tensor
+open KroneckerDelta
+
+/-- The Levi-Civita tensor `εⁱʲᵏˡ` as a complex Lorentz tensor, with `ε⁰¹²³ = 1`.
+
+The component on a multi-index `f` is the generalized Kronecker delta of `f` against the
+identity, i.e. the sign of `f` when `f` is a permutation and `0` otherwise. -/
+noncomputable def leviCivita : ℂT[.up, .up, .up, .up] :=
```
jstoobysmith: I think we could make this a real tensor, rather then a complex tensor, no?

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
```diff
@@ -0,0 +1,74 @@
...
+        = -{ fst := (↑y : ℚ), snd := 0 } := by
+    intro x y h; subst h
+    show (⟨(↑(-y) : ℚ), 0⟩ : Physlib.RatComplexNum) = ⟨-↑y, -0⟩
+    rw [Int.cast_neg, neg_zero]
+  apply hcast
+  rw [← generalizedKroneckerDelta_swap _ _ (Fin.zero_ne_one (n := 2))]
+  congr 1
+  funext i
+  fin_cases i <;> rfl
+
```
jstoobysmith: Would be nice to write down other properties of the leviCivita tensor, even in just with `sorry` proof for now.

Physlib/Relativity/Tensors/ComplexTensor/Metrics/LeviCivita.lean
Robby955: Good point let me take a look

Robby955:
- Moved the file to Physlib/Relativity/Tensors/LeviCivita/Basic.lean and updated the import tree.
- Fixed the header authorship.
- Switched it to a real Lorentz tensor (realLorentzTensor 3) as suggested. The integer components go through TensorInt.toTensor, and the antisymmetry argument carries over unchanged via permT_basis_repr_symm_apply and generalizedKroneckerDelta_swap.
- Added more properties: the explicit component formula, vanishing on a repeated index, and antisymmetry under all three adjacent index swaps (it previously only had the first pair).

For the epsilon-epsilon contraction identities, I think I would rather do those in a follow-up than stub them if that makes sense with you. I can add them regardless though. Let me know if I missed anything,

Robby955:
-awaiting-author

jstoobysmith (approved):
Approved - this now looks good to me1

Issue #353, opened 2025-02-27
Multigoal linter


jstoobysmith:
The whole of PhysLean needs the MultiGoal linter run on it and all the warnings dealt with.

In particular 

```
[leanOptions]
linter.style.multiGoal = true
```

should be added to `lakefile.toml` and the project built with `lake build`. 
This will generate warnings, some of which may be from Mathlib, but the ones from PhysLean should be dealt. For example:

```
   exact h 
   exact h2 
```

should be replaced with: 

```
  · exact h1 
  · exact h2 
```

PR #1277, opened 2026-06-25
refactor(SpaceAndTime): remove multiGoal warnings in CrossProduct

Robby955:
This focuses the post-split goals in `fderiv_cross_commute` with `·` bullets, which clears the `linter.style.multiGoal` warnings in `SpaceAndTime/Space/CrossProduct.lean`.

The edits stay inside the tactic blocks; no statement or definition is changed. The file builds with the multiGoal linter on and reports no warnings.

Toward #353.

### DIFF
--- a/Physlib/SpaceAndTime/Space/CrossProduct.lean
+++ b/Physlib/SpaceAndTime/Space/CrossProduct.lean
@@ -67,11 +67,16 @@ lemma fderiv_cross_commute {t : Time} {s : EuclideanSpace ℝ (Fin 3)}
       s j * (fderiv ℝ (fun u => f u) t) 1 i
       = (fderiv ℝ (fun t => s i * f t j - s j * f t i) t) 1:= by
     rw [fderiv_fun_sub, fderiv_const_mul, fderiv_const_mul]
-    simp only [FunLike.coe_sub, FunLike.coe_smul, Pi.sub_apply,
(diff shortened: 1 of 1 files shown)


jstoobysmith:
I'm not sure about this one. How did you check the linter.style.multiGoal warning for this file? If this actually does fix this, then I think it is an improvement.

jstoobysmith (approved):
Ok, I checked here with `set_option linter.style.multiGoal true`. So this looks good to me. Approved, and will merge shortly

Robby955:
Good question, and yes to do all of those accross the whole project, it would be similar to doing this for each one i think. 

I enabled the linter directly and recompiled the file: lake env lean -D linter.style.multiGoal=true Physlib/SpaceAndTime/Space/CrossProduct.lean. 

On the master version that reports three linter.style.multiGoal warnings in fderiv_cross_commute (lines 70, 72, 73: tactics that start with 5-6 goals and leave 4-5 unoperated after the split); after focusing those goals with the bullets it reports zero. The same command reproduces both sides.

Is that what would be intended for this issue, or maybe a detail I am missing on that one too.

jstoobysmith:
@Robby955 Many thanks for explaining ! Yep this is what is meant by this issue. I also just discovered you can see this with: 

```
set_option linter.style.multiGoal true
```
at the top of the corresponding file.

Issue #856, opened 2025-12-29
API: SpaceTime


jstoobysmith:
## Key data structure 

The key data structure is `SpaceTime`. This represents the `SpaceTime` with a given but arbitrary choice of units, with a given by arbitrary choice of zero. It is the usual notion of `SpaceTime ` used by physicists. It corresponds to Minkowski spacetime, and is used in relativity.

- [x] is defined

## Need

This API is heavily used in relativistic physics. 

## Requirements

- [x] The API shall contain a map from `SpaceTime` to `Space` and `Time`. 
- [x] The API shall contain the derivatives of functions from `SpaceTime` to a normed space.
- [ ] The API shall contain the derivatives of functions from `SpaceTime ` to manifolds. 
- [x] The API shall contain derivatives of distributions from `SpaceTime ` to normed spaces. 
- [ ] The API shall contain derivatives of distributions from `SpaceTime ` to manifolds.
- [x] The API shall contain the splitting of integrals over `SpaceTime` to integrals over `Space` and `Time`.
- [x] The API shall contain an action of the Lorentz group on `SpaceTime`. 

## Corresponding file system

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/SpaceAndTime/SpaceTime

## Parent APIs

#854 #855 #886

jstoobysmith:
Closing as included in code as API-map, this should be followed instead of this PR.

PR #1278, opened 2026-06-25
feat(SpaceAndTime): SpaceTime manifold-derivative API

Robby955:
This adds a derivative API for functions from `SpaceTime d` into a manifold, following the existing `Time.manifoldDeriv`.

`SpaceTime.manifoldDeriv I μ f y` is the tangent vector at `f y` obtained by applying `mfderiv` to the basis vector of the `μ` coordinate. The supporting lemmas are:

- `manifoldDeriv_eq` rewrites the definition back in terms of `mfderiv`.
- `deriv_eq_mfderiv` and `deriv_eq_manifoldDeriv` identify the existing `SpaceTime.deriv` with the manifold derivative when the codomain is a normed space, so the two notions agree on that overlap.
- `manifoldDeriv_const`, marked `@[simp]`, computes the derivative of a constant map.

This covers the "derivatives of functions from `SpaceTime` to manifolds" requirement of #856.

The remaining manifold requirement, derivatives of distributions from `SpaceTime` to manifolds, is not part of this PR. The current distribution API on `SpaceTime` is valued in normed/vector spaces, so a manifold version would need its own groundwork, and I think it is better tracked as a separate issue.

### DIFF
--- a/Physlib/SpaceAndTime/SpaceTime/Derivatives.lean
+++ b/Physlib/SpaceAndTime/SpaceTime/Derivatives.lean
@@ -22,6 +22,8 @@ distributions on `SpaceTime d`.
 ## ii. Key results
 
 - `deriv` : The derivative of a function `SpaceTime d → M` along the `μ` coordinate.
+- `manifoldDeriv` : The derivative of a function from `SpaceTime d` to a manifold along
+  the `μ` coordinate.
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Great! This looks good to me! Many thanks - approved.

PR #1279, opened 2026-06-26
auto-task(inner_eq_doubly_stochastic_sum): golf the doubly-stochastic inner-product proof

yidiq7:
## Summary

Golfs the proof of `HermitianMat.inner_eq_doubly_stochastic_sum` in
`QuantumInfo/ForMathlib/HermitianMat/Inner.lean` — the lemma carrying the
`--TODO: Cleanup, ew what?` marker. The statement (name, binders, type) is
byte-for-byte unchanged; only the tactic block is rewritten.

## What changed

The original proof reached the doubly-stochastic sum through heavy, opaque
machinery: three `convert!`s, a `grind`, two whole-context `simp +zetaDelta at *`
calls, a shadowed nested `have h_trace_diag`, and several 200–400 character `have`
statements spelling out the eigenvector matrices in full.

The new proof follows the same mathematics but makes every step explicit and
idiomatic:

- `U := A.H.eigenvectorUnitary`, `V := B.H.eigenvectorUnitary` are introduced with
  `set`; unitarity is just `U.2.1` / `U.2.2`.
- The diagonalizations `A = U · diag(a) · Uᴴ` (and likewise for `B`) come directly
  from `Matrix.IsHermitian.spectral_theorem` via
  `simpa [Unitary.conjStarAlgAut_apply]`.
- A single `key` step rewrites `A · B` as the conjugation
  `U · (diag(a) · C · diag(b) · Cᴴ) · Uᴴ`, after which `Matrix.trace_mul_cycle`
  together with unitarity collapses the trace to that of `diag(a)·C·diag(b)·Cᴴ`.
- That trace is expanded entrywise to the doubly-stochastic sum with targeted
  `simp only` sets, and the real part is taken with `Complex.ofReal_re`.

## Effect

- Proof body shrinks from ~3998 to ~2072 characters; the longest line drops from
  387 to 91 characters.
- All `convert!` (×3), `grind`, and `simp +zetaDelta at *` (×2) calls are removed,
  replaced by `rw`, `conv`, `simpa`, and targeted `simp only`. This is both faster
  to elaborate (no whole-context simp / `grind`) and far more readable.

## Verification

`lake build QuantumInfo.ForMathlib.HermitianMat.Inner` succeeds, as do the
downstream modules that consume the lemma
(`QuantumInfo.ForMathlib.HermitianMat.Schatten` and `QuantumInfo.Entropy.Relative`).
The proof closes with no `sorry`, no new errors or warnings, and `#print axioms`
reports only `propext`, `Classical.choice`, and `Quot.sound`.

### DIFF
--- a/QuantumInfo/ForMathlib/HermitianMat/Inner.lean
+++ b/QuantumInfo/ForMathlib/HermitianMat/Inner.lean
@@ -524,36 +524,38 @@ lemma inner_eq_doubly_stochastic_sum {d : Type*} [Fintype d] [DecidableEq d]
     let C := A.H.eigenvectorUnitary.val.conjTranspose * B.H.eigenvectorUnitary.val
     ⟪A, B⟫_ℝ = ∑ i, ∑ j,
       A.H.eigenvalues i * B.H.eigenvalues j * (‖C i j‖^2) := by
-  -- By the properties of the trace and diagonalization, we can rewrite the trace of AB as the sum of the products of the eigenvalues of A and B, multiplied by the squared norms of the entries of the product of their eigenvector matrices.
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good! APproved

PR #1281, opened 2026-06-26
auto-task(inner_zero_iff_aux_lemma): golf the PSD trace-zero kernel lemma

yidiq7:
## Summary

Golfs the proof of the private lemma `HermitianMat.inner_zero_iff_aux_lemma` in
`QuantumInfo/ForMathlib/HermitianMat/Inner.lean` (it was even flagged `--TODO cleanup`).

For PSD matrices `A`, `B` the lemma states

```
RCLike.re (A.val * B.val).trace = 0 ↔
  LinearMap.range (Matrix.toEuclideanLin A.val) ≤ LinearMap.ker (Matrix.toEuclideanLin B.val)
```

The **statement is unchanged**; only the proof body is rewritten. No other declaration,
import, or docstring is touched.

## What changed

The previous proof was auto-generated and ~79 lines of heavily repetitive tactics:
nested `have`s with the same name, repeated `obtain ⟨val, property⟩ := A` / `subst`,
several `simp_all only`, an `erw`, and a hand-rolled `nlinarith`/`RCLike.ext` argument
that `tr(MᴴM) = 0 ↔ M = 0`.

The new proof is ~36 lines (including comments) and follows the math directly:

1. **Kernel condition → matrix identity.** `LinearMap.range_le_ker_iff` turns the kernel
   inclusion into `toEuclideanLin B ∘ₗ toEuclideanLin A = 0`; this composition is
   `toEuclideanLin (B * A)` (via `mulVec_mulVec`), so by injectivity of the linear
   equiv it is `B * A = 0`, and by Hermitianness of `A`, `B` it is `A * B = 0`.
2. **Trace step via a single cyclic identity.** Factor `A = Cᴴ C`, `B = Dᴴ D`
   (`CStarAlgebra.nonneg_iff_eq_star_mul_self`). A short cyclic-trace rewrite gives
   `tr(A * B) = tr(EᴴE)` with `E = D Cᴴ`. Since `EᴴE` is PSD its trace is a nonnegative
   real, so `re = 0` forces the trace to vanish; `Matrix.trace_conjTranspose_mul_self_eq_zero_iff`
   then gives `E = 0`, hence `A * B = 0`. The converse is immediate.

This replaces the bespoke sum-of-squares `nlinarith` argument with the existing Mathlib
lemma `Matrix.trace_conjTranspose_mul_self_eq_zero_iff`, and removes all the
`simp_all only` / `obtain`/`subst` boilerplate.

## Verification

- `lake build QuantumInfo.ForMathlib.HermitianMat.Inner` succeeds with no errors or warnings.
- `#print axioms` reports only `propext`, `Classical.choice`, `Quot.sound` — no new axioms,
  no `sorry`.

### DIFF
--- a/QuantumInfo/ForMathlib/HermitianMat/Inner.lean
+++ b/QuantumInfo/ForMathlib/HermitianMat/Inner.lean
@@ -214,85 +214,42 @@ private theorem inner_zero_iff_aux_lemma [DecidableEq n] (hA₁ : A.mat.PosSemid
   RCLike.re (A.val * B.val).trace = 0 ↔
     LinearMap.range (Matrix.toEuclideanLin A.val) ≤
       LinearMap.ker (Matrix.toEuclideanLin B.val) := by
+  -- The kernel inclusion `range ≤ ker` says `(toEuclideanLin B) ∘ (toEuclideanLin A) = 0`, i.e.
+  -- `B * A = 0`, which by Hermitianness of `A` and `B` is `A * B = 0`.
+  rw [LinearMap.range_le_ker_iff,
+    show Matrix.toEuclideanLin B.val ∘ₗ Matrix.toEuclideanLin A.val
+        = Matrix.toEuclideanLin (B.val * A.val) from by
+      ext v; simp [Matrix.toLpLin_apply, Matrix.mulVec_mulVec, Matrix.toEuclideanLin],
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved. Btw some updates to this script, so maybe worth downloading a new version :).

PR #1284, opened 2026-06-26
feat: Close the easy FreeParticle sorryful gap

NicolaBernini:
# Overview

feat: Close the easy FreeParticle sorryful gap


jstoobysmith:
What has this got to do with sorryful? I think this is a good change nevertheless as this is likely throwing a warning

jstoobysmith (approved):
Approved - will merge now.

PR #1288, opened 2026-06-26
feat: Refactor downstream use away from FieldStrengthMatrix

NicolaBernini:
# Overview

feat: Refactor downstream use away from FieldStrengthMatrix

### DIFF
--- a/Physlib/ClassicalMechanics/Pendulum/SlidingPendulum.lean
+++ b/Physlib/ClassicalMechanics/Pendulum/SlidingPendulum.lean
@@ -5,7 +5,7 @@ Authors: Shlok Vaibhav Singh
 -/
 module
 
-public import Mathlib.Data.Real.Basic
+public import Mathlib.Analysis.SpecialFunctions.Trigonometric.Angle
 /-!
 # Sliding Pendulum
 ### Tag: LnL_1.5.2
@@ -77,7 +77,7 @@ structure ConfigurationSpace where
   /-- The horizontal position `x₁` of the support mass. -/
   supportPosition : ℝ
   /-- The angle `φ` that the string makes with the vertical. -/
-  angle : ℝ
+  angle : Real.Angle
 
 end SlidingPendulum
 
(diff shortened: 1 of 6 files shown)


jstoobysmith:
@NicolaBernini  I think this now has a merge conflict with your other PR

NicolaBernini:
> @NicolaBernini I think this now has a merge conflict with your other PR

Hi @jstoobysmith should be fixed now

Physlib/Electromagnetism/Kinematics/FieldStrength.lean
```diff
@@ -373,6 +373,11 @@ These lemmas express the components of the field strength tensor using index eva
 
 -/
 
+/-- The field-strength component obtained by evaluating both tensor indices. -/
+noncomputable abbrev fieldStrengthComponent {d} (A : ElectromagneticPotential d) (x : SpaceTime d)
```
jstoobysmith: This I'm not so sure about as it seems to be just a rewriting of fieldStrengthMatrix. I think what is needed here is to write the results in terms of `toField {toFieldStrength A [mu] [nu]}` (not quite the right syntax but hopefully obvious what I mean). 

I think the idea is to base everything more concretely on `toFieldStrength A` rather than introducing a new version of it.

Physlib/ClassicalMechanics/Pendulum/SlidingPendulum.lean
```diff
@@ -63,9 +63,21 @@ namespace ClassicalMechanics
...
+The configuration space of the sliding pendulum system.
+
+The generalized coordinates are the horizontal position of the support mass and the angle
+that the string makes with the vertical.
+-/
+structure ConfigurationSpace where
+  /-- The horizontal position `x₁` of the support mass. -/
+  supportPosition : ℝ
+  /-- The angle `φ` that the string makes with the vertical. -/
+  angle : ℝ
```
jstoobysmith: I think I would make this `[Real.Angle](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecialFunctions/Trigonometric/Angle.html#Real.Angle)` rather then `Real`.

jstoobysmith:
(Think we should remove the sliding pendulum changes from this PR)

PR #1294, opened 2026-06-26
feat(ClassicalMechanics): add amplitude–phase parametrization of the harmonic oscillator

nateabr:
Adds the amplitude–phase parametrization — the final alternate initial-condition form for the
harmonic oscillator. Lives in a new, self-contained **section E** of
`HarmonicOscillator/Solution.lean`; **no existing declarations are changed**.

### New API
The new section introduces `AmplitudePhase`, an initial-condition type carrying an amplitude
`A : ℝ` and a phase `φ : ℝ`. The forward conversion `toInitialConditions` maps `(A, φ)` to the
standard initial position and velocity `(A cos φ, A ω sin φ)`, and the lemmas
`toInitialConditions_trajectory_eq_cos` and `toInitialConditions_velocity_eq_sin` show the
resulting trajectory is the normal form `x(t) = A cos(ω t − φ)`, with velocity
`−A ω sin(ω t − φ)`. The reverse conversion `fromInitialConditions` recovers `(A, φ)` as the polar
coordinates of the phase vector `(x₀, v₀ / ω)` via `Complex.arg`, and
`toInitialConditions_fromInitialConditions` proves it is a right inverse — converting initial
conditions to amplitude–phase form and back recovers them.

### What to look at
1. **Scalars, not vectors** — `A`, `φ` are `ℝ` (a magnitude and an angle), unlike the
   `EuclideanSpace ℝ (Fin 1)` fields of the other IC types.
2. **`Complex.arg` in the reverse map only** — used to avoid the `arctan` quadrant case-split; the
   trajectory itself stays real-valued. First use of `Complex.arg` in `ClassicalMechanics`.

Minor (separate commit): §D.1–§D.3 headings promoted `##` → `###` to match the TOC.

---
*Developed with assistance from Claude (Anthropic); all mathematics and proofs were
reviewed and verified to compile.*

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
@@ -589,7 +589,7 @@ end InitialConditions
 
 /-!
 
-## D.1. Correctness of InitialConditionsAtTime conversion
+### D.1. Correctness of InitialConditionsAtTime conversion
 
 We now prove the correctness lemmas for the `InitialConditionsAtTime.toInitialConditions`
 conversion function. These show that the conversion produces a trajectory that passes through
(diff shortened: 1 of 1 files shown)


nateabr:
t-classical-mechanics

jstoobysmith (approved):
This looks great! I’ve approved

PR #1296, opened 2026-06-27
refactor(Mathematics): remove erws from LieTrace

giuseppesorge:
## Summary
Removes four `erw` calls in `Physlib/Mathematics/DataStructures/Matrix/LieTrace.lean` (toward #385),
replacing each with `rw`. **No statement is changed** — the edits are purely proof-internal coercion
normalizations.

## Changes
| Lemma | Fix |
|---|---|
| `det_exp_of_blockTriangular_id` | `erw [← NormedSpace.exp_sum …]` → plain `rw` (the `erw` was unnecessary) |
| `det_exp` | `erw [h_conj, trace_unitary_conj]` → `rw [h_conj, Unitary.coe_star, trace_unitary_conj]` (normalize `↑(star U)` vs `star ↑U`) |
| `det_exp` | `erw [h_conj, det_exp_unitary_conj]` → `rw [h_conj, Unitary.coe_star, det_exp_unitary_conj]` (same coercion) |
| `det_exp_real` | `erw [← Complex.ofReal_exp]` → `rw [Complex.coe_algebraMap, ← Complex.ofReal_exp]`; drops a now-redundant `simp_all only [...]` |

## Scope
- **Zero statement changes**; the two public theorems (`det_exp`, `det_exp_real`) and all helpers keep identical signatures. Diff is `+4 / −6`, one file.
- The bridging lemmas used are the canonical `@[simp, norm_cast]` ones: `Unitary.coe_star`, `Complex.coe_algebraMap`.

## Deliberately left (separate follow-up)
The file still has two `erw`s in `exp_map_algebraMap`. Removing them cleanly would require specializing the
private helper `map_tsum` from `AddMonoidHom` to `RingHom` — but its proof only uses additivity, so that
would weaken a general helper just to dodge a coercion. I'd rather leave those two for a dedicated
follow-up than regress the helper here. (Happy to do that follow-up if you'd like.)

## Verification (Lean v4.31.0)
- `lake build Physlib.Mathematics.DataStructures.Matrix.LieTrace` succeeds; 0 errors.
- `lake env lean -Dlinter.style.multiGoal=true -Dlinter.style.longLine=true …` → 0 warnings.
- No `sorry`/`admit`; no trailing whitespace.

Toward #385.

### DIFF
--- a/Physlib/Mathematics/DataStructures/Matrix/LieTrace.lean
+++ b/Physlib/Mathematics/DataStructures/Matrix/LieTrace.lean
@@ -158,7 +158,7 @@ lemma det_exp_of_blockTriangular_id {A : Matrix m m 𝕂} (hA : BlockTriangular 
     diag_exp_of_blockTriangular_id hA
   simp_rw [← diag_apply]
   simp_rw [h_diag_exp]
-  erw [← NormedSpace.exp_sum Finset.univ]
+  rw [← NormedSpace.exp_sum Finset.univ]
   congr 1
 
 /-- The trace is invariant under unitary conjugation. -/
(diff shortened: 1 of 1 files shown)


jstoobysmith:
```quote 
The two erws in exp_map_algebraMap cannot be removed without restating the private helper map_tsum
(a →+ → →+* generalization). It is single-use and the restate is the cleanest fix, but if you'd prefer
a strictly zero-statement-change PR I'm happy to split this into the four plain rw swaps now and the
map_tsum restate as a follow-up.
``` 
Please could we do this. I'm a bit confused about whether the change to `map_tsum` is a generalization or a restriction. If a generalization then great.

giuseppesorge:
Done — I've split it: this PR is now just the four plain `erw → rw` swaps, with no statement changes.

Good catch on `map_tsum`: the change I'd made was a **restriction**, not a generalization — `→+*` (`RingHom`, needs semiring structure) is *more* demanding than the original `→+` (`AddMonoidHom`), and since the proof only uses additivity it bought nothing. So I've dropped it entirely; `map_tsum` is left exactly as it was. The two remaining `erw`s in `exp_map_algebraMap` are left for a dedicated follow-up.

Removing the `awaiting-author` tag.

giuseppesorge:
-awaiting-author

jstoobysmith (approved):
Many thanks - approved.

PR #1297, opened 2026-06-27
feat: Add a general lemma specifying the derivative of distributions obtained from functions.

NicolaBernini:
# Overview

feat: Add a general lemma specifying the derivative of distributions obtained from functions.


jstoobysmith (commented):
Awesome! This is great! Many thanks.

jstoobysmith (approved):

PR #1300, opened 2026-06-27
refactor(ClassicalMechanics): model harmonic oscillator configuration space by EuclideanSpace

nateabr:
## Summary

Refactors `ClassicalMechanics.HarmonicOscillator.ConfigurationSpace` so its chosen
coordinate value is `EuclideanSpace ℝ (Fin 1)` instead of `ℝ`, resolving the in-file TODO
and aligning the configuration-space coordinate with the model used for position and
velocity in `Basic.lean` and `Solution.lean`. The existing one-dimensional
configuration-space API is preserved.

## Changes

The structure field `ConfigurationSpace.val` changes from `ℝ` to `EuclideanSpace ℝ (Fin 1)`,
and the rest of the file is adjusted to that coordinate. The algebraic instances now act
coordinatewise on the underlying Euclidean value, while the metric and inner-product
instances delegate to the underlying coordinate; these become `noncomputable`, since the
`PiLp` norm and inner product are.

The previous real-valued coordinate maps are replaced by a single linear isometry,
`valLinearIsometryEquiv : ConfigurationSpace ≃ₗᵢ[ℝ] EuclideanSpace ℝ (Fin 1)`, from which
`valHomeomorphism` is derived; because the norm is defined to be that of the coordinate,
this identification is definitional. The chart and smooth-manifold instances are now modeled
on `EuclideanSpace ℝ (Fin 1)`, and finite-dimensionality and completeness are re-established
through the isometry. Finally, `toSpace` and the coordinate-access lemmas are updated to read
off the Euclidean coordinate, and the file gains module-level and section-level documentation
matching the adjacent harmonic-oscillator files.

---
*Developed with assistance from Claude (Anthropic); all mathematics and proofs were
reviewed and verified to compile.*

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2026 Nicola Bernini. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Nicola Bernini
+Authors: Nicola Bernini, Nathaneal Sajan
 -/
 module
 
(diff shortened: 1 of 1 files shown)


nateabr:
t-classical-mechanics

nateabr:
The first CI run failed the Lean based style linters job (the simpNF linter); all other checks passed. It flagged two @[simp] lemmas in my target file ConfigurationSpace.lean: coe_apply : x i = x.val i and toSpace_apply : q.toSpace i = q.val i.

Both are projection lemmas whose right-hand side evaluates a coordinate of the underlying EuclideanSpace ℝ (Fin 1) value (x.val i). Indexing an EuclideanSpace coordinate goes through the PiLp/WithLp coercion machinery, and as simp rewrite rules these lemmas drive simp to recurse through that coercion until it hits the recursion limit.

This is a direct consequence of the coordinate-model change in this PR. With the previous val : ℝ, the right-hand side was a plain real and the lemmas were well-behaved as simp rules; once val : EuclideanSpace ℝ (Fin 1), componentwise indexing introduces the looping coercion.

For the fix I just naively dropped the @[simp] from both lemmas for now. Rather than spot-fixing the two lemmas, I suspect this can be addressed in a more robust way as part of the broader geometric API work I plan to do for the next in-file TODO — and I'd like that to build cumulatively on the work in this file. Hope this clears up the ad-hoc change.

Physlib/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
```diff
@@ -98,75 +136,88 @@ instance : AddCommGroup ConfigurationSpace where
...
+The inner product is that of the underlying Euclidean coordinate, making
+`ConfigurationSpace` a real inner-product space. Physically this is the metric used to form
+kinetic energy and to identify configuration space with its dual.
+-/
 
 open InnerProductSpace
 
-instance : Inner ℝ ConfigurationSpace where
-  inner x y := x.val * y.val
+noncomputable instance : Inner ℝ ConfigurationSpace where
```
jstoobysmith, line 198: For a future PR it would be good to remove all of the algebraic and inner-product structure on `ConfigurationSpace`, and go straight into defining it as a manifold. This is because this approach will generalize to other classical mechanic systems, whilst defining an inner-product structure won't.

jstoobysmith (approved):
This looks good to me! Approved. Thanks.

Physlib/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean
nateabr: Thanks, a really sharp and neccessary comment as I see now that the manifold structure is effectively defined "in reverse". Will get on this change immediately.

PR #1302, opened 2026-06-28
Add total-divergence equivalence API to PhyslibAlpha

juanjfndz:
This PR adds the next small local Classical Field Theory API layer to `PhyslibAlpha`: lagrangian equivalence up to total divergences.

It builds directly on the total-divergence layer from #1291 and introduces:

- `HasTotalDivergenceDifference`, recording that a target lagrangian density differs from a source density by a packaged total-divergence density;
- `IsEulerLagrangeEquivalent`, recording equality of local Euler--Lagrange operators along every field;
- `TotalDivergenceEquivalence`, bundling the source lagrangian, target lagrangian, total-divergence witness, density-difference identity, and Euler--Lagrange equivalence data;
- basic API lemmas for the density identity, equality of Euler--Lagrange operators, and equivalence of criticality under the existing admissibility/smoothness hypotheses.

As in the previous PR, this is intentionally a packaging/API layer for the current Alpha local stack. Since `Lagrangian` carries coordinate derivatives explicitly, this PR does not try to build the modified lagrangian syntactically by adding a divergence or derive the Euler--Lagrange equality from symbolic coordinate calculus. It records the data needed by later equivalence and first-variation layers while keeping the API honest about what is currently proved.

The module follows the local coordinate setup from Cortés--Haupt, arXiv:1612.03100v2, Chapter 5.

Validation:

- `lake build PhyslibAlpha.ClassicalFieldTheory.Local.TotalDivergenceEquivalence`
- `lake build PhyslibAlpha`
- `lake exe runPhyslibAlphaLinters`
- `./scripts/PhyslibAlpha/alphaPythonLinters.sh`
- `python3 ./scripts/PhyslibAlpha/alphaFileImports.py`
- `python3 ./scripts/PhyslibAlpha/noAlphaImports.py`

PR #1307, opened 2026-06-28
Add Euler-Lagrange equivalence API to PhyslibAlpha

juanjfndz:
This PR adds a small API layer for `IsEulerLagrangeEquivalent` in the local Classical Field Theory Alpha stack.

It builds directly on the total-divergence equivalence layer from #1302 and introduces:

- reflexivity, symmetry, and transitivity for `IsEulerLagrangeEquivalent`;
- a reusable `IsEulerLagrangeEquivalent.isCritical_iff` lemma transferring criticality between equivalent lagrangians under the existing admissibility and smoothness hypotheses;
- a simplification of `TotalDivergenceEquivalence.isCritical_iff` through this new predicate-level API.

This PR does not introduce new mathematical objects or symbolic calculus. It only exposes the minimal algebraic API around the already-approved Euler-Lagrange equivalence predicate, so later total-divergence and equivalence layers can reuse the predicate directly instead of duplicating the same proof pattern.

Validation:

- `lake build PhyslibAlpha.ClassicalFieldTheory.Local.TotalDivergenceEquivalence`
- `lake build PhyslibAlpha`
- `lake exe runPhyslibAlphaLinters`
- `./scripts/PhyslibAlpha/alphaPythonLinters.sh`
- `python3 ./scripts/PhyslibAlpha/alphaFileImports.py`
- `python3 ./scripts/PhyslibAlpha/noAlphaImports.py`

### DIFF
--- a/PhyslibAlpha/ClassicalFieldTheory/Local/TotalDivergenceEquivalence.lean
+++ b/PhyslibAlpha/ClassicalFieldTheory/Local/TotalDivergenceEquivalence.lean
@@ -29,13 +29,15 @@ which facts are data and which facts are proved.
 
 - `ClassicalFieldTheory.Local.HasTotalDivergenceDifference`
 - `ClassicalFieldTheory.Local.IsEulerLagrangeEquivalent`
+- `ClassicalFieldTheory.Local.IsEulerLagrangeEquivalent.isCritical_iff`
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved

PR #1308, opened 2026-06-28
refactor(classical-mechanics): characterize special times of harmonic oscillator trajectories

nateabr:
Builds on the amplitude-phase parametrization from #1294 to characterize special times of a harmonic-oscillator trajectory in a new section F, and closes the open zero-crossing TODO. It also groups the existing periodicity declarations into a labelled section G.

### New API

Section F first transports the amplitude-phase normal form back to ordinary initial conditions:

- `trajectory_eq_cos` rewrites the trajectory as
  `x(t) = A cos(ω t - φ)`, where `A` and `φ` come from
  `AmplitudePhase.fromInitialConditions`.
- `trajectory_velocity_eq_sin` rewrites the velocity as
  `v(t) = -Aω sin(ω t - φ)`.

Using these two normal forms, section F then proves the special-time characterizations:

- `trajectory_velocity_eq_zero_iff_sin_eq_zero` says that, for nonzero amplitude, the velocity is
  zero exactly when the sine factor in the amplitude-phase velocity vanishes.
- `trajectory_velocity_eq_zero_iff_exists_int` converts this into the explicit time family
  `t = (φ + nπ) / ω`.
- `trajectory_velocity_eq_zero_iff_norm_eq_amplitude` identifies the velocity-zero times with the
  turning points of the motion: these are exactly the times when `‖x(t)‖ = A`.
- `trajectory_eq_zero_iff_cos_eq_zero` says that, for nonzero amplitude, the trajectory passes
  through the origin exactly when the cosine factor in the amplitude-phase trajectory vanishes.
- `trajectory_eq_zero_iff_exists_int` converts this into the explicit time family
  `t = (φ + (2n + 1)π / 2) / ω`.

Together these describe the velocity-zero times, the position at those times, and the zero-crossing times in terms of the phase `ω t - φ`.

### Removed API

The old zero-velocity section is replaced by the amplitude-phase formulation. The following lemmas are removed:

- `tan_time_eq_of_trajectory_velocity_eq_zero`
- `trajectory_velocity_eq_zero_at_arctan`
- `trajectory_velocity_eq_zero_iff`

The new statements give exact `iff` characterizations and integer parametrizations of the relevant times. The norm result is now stated directly as `‖x(t)‖ = A`.

### Motivation

In the raw initial-condition form, the special-time calculations involve tangent/arctangent algebra and separate case analysis. In amplitude-phase form, they reduce to standard zero sets of sine and cosine:

- velocity-zero times are zeros of `sin(ω t - φ)`;
- zero crossings are zeros of `cos(ω t - φ)`;
- turning-point norms follow from `|cos(ω t - φ)| = 1`.

This gives a smaller and more uniform API for later trajectory results.

### Section G

The existing declarations

- `period`
- `trajectory_periodic`
- `return_time`

are grouped under `G. Periodicity and recurrence` with section documentation. Their statements and proofs are unchanged.

### Reviewer map

1. Start with `trajectory_eq_cos` and `trajectory_velocity_eq_sin`.
2. Review the velocity-zero and turning-point lemmas in F.2-F.3.
3. Review the zero-crossing lemmas in F.4; these close the TODO.
4. Skim section G, which is documentation and organization only.

Developed with AI assistance; all mathematics and proofs were reviewed by the human author.

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Solution.lean
@@ -51,11 +51,15 @@ prove that they satisfy the equation of motion, and prove some properties of the
   - E.2. Conversion to standard initial conditions
   - E.3. The trajectory in normal form
   - E.4. Recovering the amplitude and phase
-- F. The trajectories at zero velocity
-  - F.1. The times at which the velocity is zero
-  - F.2. A time when the velocity is zero
+- F. Special conditions of the trajectory
+  - F.1. Normal form for standard initial conditions
+  - F.2. Times at which the velocity is zero
   - F.3. The position when the velocity is zero
-- G. Some open TODOs
+  - F.4. Times at which the trajectory passes through zero
+- G. Periodicity and recurrence
+  - G.1. The period
+  - G.2. Periodicity of the trajectory
+  - G.3. Return to the initial state
 
 ## iv. References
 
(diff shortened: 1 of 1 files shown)


nateabr:
t-classical-mechanics

jstoobysmith (approved):
Maybe we should think about splitting this file up into different parts. Though I've approved this for now and will merge shortly.

PR #1315, opened 2026-06-29
refactor(Relativity): remove erws from Weyl/Metric.lean

giuseppesorge:
Convert the 9 `erw` in ComplexTensor/Weyl/Metric.lean to `rw`/`simp`. The `*_symm_expand_tmul` site is a direct `rw`. In the four `*Contraction_apply_metric` lemmas the erws were forced by a preceding `simp only [← IntertwiningMap.toLinearMap_apply]` that re-wrapped every application into `.toLinearMap` form, no longer matching the `⇑`-stated `*Contraction_basis`/`*Unit_apply_one` lemmas; dropping that line lets `simp only [*Contraction_basis]` and `rw [*Unit_apply_one, ...]` match directly. No statements changed; Unit.lean and Contraction.lean untouched. (erw in file: 9 → 0)


jstoobysmith (approved):
Looks good - approved.

PR #1316, opened 2026-06-29
refactor(Relativity): remove erws from Weyl/Two.lean

giuseppesorge:
Convert the 20 `erw` in the `*ToMatrix_ρ` group-action lemmas of ComplexTensor/Weyl/Two.lean to `rw`. The sites needing a defeq bridge over `Finsupp.linearEquivFunOnFinite` are handled by a single private `coe_linearEquivFunOnFinite` rfl-lemma. No statements changed; the affected lemmas still depend only on [propext, Classical.choice, Quot.sound]. (erw in file: 20 → 0)


jstoobysmith (approved):
Looks good - approved.

PR #1318, opened 2026-06-29
feat: Improvements to Elaboration of tensor expressions

jstoobysmith:
Two improvements to Elaboration of tensor expressions: 

- Terms of type `Tensor S _` now no longer get a redundant `toTensor` inserted when they appear in tensor expressions. 
- Redundant permutations of indices (which trivially correspond to the identity) are no longer added. 

co-authored with claude.


morrison-daniel (approved):
Looks good

PR #1319, opened 2026-06-29
feat: Adding inherited order instances for WithDim

NicolaBernini:
# Overview

feat: Adding inherited order instances for WithDim


jstoobysmith (commented):
Looks good! Approved

jstoobysmith (approved):

PR #1323, opened 2026-06-29
refactor(ClassicalMechanics): present harmonic oscillator configuration space as a manifold

nateabr:
## Summary

Follow-up to #1300, addressing [the review comment](https://github.com/leanprover-community/physlib/pull/1300#discussion_r3487457120) by @jstoobysmith: remove all algebraic and inner-product structure from `ClassicalMechanics.HarmonicOscillator.ConfigurationSpace` and present `Q` **directly as a smooth manifold** with a chosen global Euclidean coordinate.

Previously the manifold structure was built "in reverse," through a vector → norm → inner-product → linear-isometry stack. That inner-product structure does not generalize to other configuration spaces.

## Changes

**Topology is now induced directly by the coordinate**, and the global chart comes from a plain coordinate homeomorphism: a set in `Q` is open exactly when its image under `val` is open, and continuity of the chart both ways is the universal property of the induced topology — no norm or isometry involved.

Removed (file: `Physlib/ClassicalMechanics/HarmonicOscillator/ConfigurationSpace.lean`):
- **Algebraic** — instances `Zero`, `OfNat _ 0`, `Add`, `Neg`, `Sub`, `SMul ℝ`, `AddGroup`, `AddCommGroup`, `Module ℝ`; lemmas `zero_val`, `add_val`, `neg_val`, `sub_val`, `smul_val`. (`Q` is not a vector space.)
- **Normed/metric** — instances `Norm`, `Dist`, `SeminormedAddCommGroup`, `NormedAddCommGroup`, `NormedSpace ℝ`; lemmas `norm_val`, `dist_val`. (No length/distance on `Q` itself.)
- **Inner product** — instances `Inner ℝ`, `InnerProductSpace ℝ`; lemma `inner_def`; differentiability lemmas `differentiable_inner_self`, `differentiableAt_inner_self`, `contDiff_inner_self`.
- **Isometry bridge** — `valLinearIsometryEquiv` (`≃ₗᵢ`).
- **Vector-space aftereffects** — instances `FiniteDimensional ℝ`, `CompleteSpace`.
- Import `Mathlib.Analysis.InnerProductSpace.Calculus` (no longer needed).

Added (same file):
- `instance : TopologicalSpace ConfigurationSpace` — induced by `ConfigurationSpace.val`.
- `def valEquiv : ConfigurationSpace ≃ EuclideanSpace ℝ (Fin 1)` — plain coordinate equivalence.
[...]

PR #1329, opened 2026-06-30
feat(ClassicalMechanics): geometric trajectories and velocity of the harmonic oscillator

nateabr:
## Summary

Builds the next layer of the manifold-first harmonic-oscillator model (follow-up to #1323): trajectories as curves in the configuration manifold `Q`, their coordinate representation, the smoothness bridge to ordinary calculus, and geometric velocity in the tangent bundle.

### `Geometric/Basic.lean` &nbsp;·&nbsp; moved from `ConfigurationSpace.lean`

| Declaration | Kind | What it does |
|---|---|---|
| `valDiffeomorph` | def | The global coordinate chart as an analytic diffeomorphism `Q ≃ₘ EuclideanSpace ℝ (Fin 1)` — the smooth upgrade of `valHomeomorphism`. Transfers smoothness/differentiability between `Q` and its coordinate model, and underpins every Trajectory result. |

### `Geometric/Trajectory.lean` &nbsp;·&nbsp; new file

| Declaration | Kind | What it does |
|---|---|---|
| `Trajectory` | abbrev | A curve `Time → ConfigurationSpace` into the configuration manifold. |
| `coord` | def | The trajectory in the global coordinate: `t ↦ (γ t).val`, a curve into `EuclideanSpace ℝ (Fin 1)`. Definitionally `valDiffeomorph ∘ γ`. |
| `toSpace` | def | The physical-space position `Time → Space 1` along the trajectory. |
| `coord_apply` | lemma | Pointwise `coord γ t = (γ t).val`. Non-`@[simp]` (the `PiLp`/`WithLp` recursion from #1300). |
| `contMDiff_iff_contDiff_coord` | lemma | Manifold smoothness of `γ` ⟺ ordinary `ContDiff` smoothness of `coord γ`, at every order `n`. The coordinate-side bridge for later regularity. |
| `velocity` | def *(noncomp.)* | Geometric velocity at `t` as a tangent vector in `T_{γ t}Q`, via `mfderiv` in the unit-time direction. First use of the tangent bundle `TQ` here. |
| `velocity_eq_deriv_coord` | lemma | In the global coordinate, velocity equals the time derivative of the coordinate curve. |
| `mdifferentiableAt_iff_coord`, `mfderiv_eq_mfderiv_coord` | private | Support for `velocity_eq_deriv_coord` — see the reviewer note. |

### `Physlib.lean`

Aggregator imports updated for the `Geometric/` folder: `…HarmonicOscillator.ConfigurationSpace` → `…Geometric.Basic`, plus `…Geometric.Trajectory`, kept in sorted order.

## Reviewer map

1. `Geometric/Basic.lean` — `valDiffeomorph` (section D).
2. `Trajectory.lean` §A — `Trajectory`, `coord`, `toSpace`.
3. `Trajectory.lean` §B — the smoothness bridge.
4. `Trajectory.lean` §C — `velocity` and `velocity_eq_deriv_coord` (see the note below).

## Reviewer note — Section C proof (`velocity_eq_deriv_coord`)

I first had a single self-contained proof of `velocity_eq_deriv_coord` that compiled. I've since refactored it into the private lemmas in §C. The original is below, followed by the concerns that led to the refactor. I'm happy to revert to the single proof if that is preferred.

Original proof (compiled):
```lean
lemma velocity_eq_deriv_coord (γ : Trajectory) (t : Time) :
    velocity γ t = ∂ₜ (coord γ) t := by
  rw [Time.deriv_eq_mfderiv]
  have hmdiff :
      MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) γ t ↔
        MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) (coord γ) t := by
    constructor
    · intro hγ
      rw [show coord γ = ConfigurationSpace.valDiffeomorph ∘ γ from rfl]
      exact MDifferentiableAt.comp t
        ((ConfigurationSpace.valDiffeomorph.mdifferentiable WithTop.top_ne_zero).mdifferentiableAt)
        hγ
    · intro hcoord
      have hγ : MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1))
          (ConfigurationSpace.valDiffeomorph.symm ∘ coord γ) t :=
        MDifferentiableAt.comp t
          ((ConfigurationSpace.valDiffeomorph.symm.mdifferentiable
            WithTop.top_ne_zero).mdifferentiableAt)
          hcoord
      convert hγ using 1
      funext t
      change γ t = ConfigurationSpace.valDiffeomorph.symm (coord γ t)
      rw [coord]
      exact (ConfigurationSpace.valDiffeomorph.symm_apply_apply (γ t)).symm
  by_cases hγ :
      MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) γ t
  · have hcoord := hmdiff.mp hγ
    simp [velocity, coord, mfderiv, writtenInExtChartAt, hγ, hcoord,
      ConfigurationSpace.valHomeomorphism, ConfigurationSpace.valEquiv]
    change (fderiv ℝ (coord γ) t) 1 = (fderiv ℝ (coord γ) t) 1
    rfl
  · have hcoord :
        ¬ MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) (coord γ) t :=
      fun hcoord => hγ (hmdiff.mpr hcoord)
    simp [velocity, coord, mfderiv, hγ, hcoord]
    change (0 : EuclideanSpace ℝ (Fin 1)) =
      (0 : Time →L[ℝ] EuclideanSpace ℝ (Fin 1)) 1
    simp
```

Concerns with the original proof:

1. Both `simp [… mfderiv, writtenInExtChartAt, valHomeomorphism, valEquiv]` calls unfold the definitions of `mfderiv` and `writtenInExtChartAt`, so the proof relies on the current internal form of the manifold derivative. A change to those definitions in Mathlib could stop the `simp` from reducing, without an informative error.

2. Each branch closes with `change X = X; rfl` (and `change 0 = …; simp`), where the `change` restates a goal that `simp` reduced to but did not discharge. This pins the exact term form by hand instead of closing the goal through a stable lemma.

The refactor splits the proof into three parts.

- `mdifferentiableAt_iff_coord` establishes `MDifferentiableAt γ t ↔ MDifferentiableAt (coord γ) t`. Each direction composes with `valDiffeomorph` (resp. its inverse) through `Diffeomorph.mdifferentiable` and `MDifferentiableAt.comp`, so this step uses no manifold-derivative internals.

- `mfderiv_eq_mfderiv_coord` proves `mfderiv γ t = mfderiv (coord γ) t`, by cases on `MDifferentiableAt γ t`. In the differentiable case the equality reduces to a coordinate computation — the one place that unfolds `mfderiv`/`writtenInExtChartAt`. In the non-differentiable case both sides are `0` by `mfderiv_zero_of_not_mdifferentiableAt`, with the two differentiability hypotheses matched through `mdifferentiableAt_iff_coord`.

- `velocity_eq_deriv_coord` rewrites `velocity` to `mfderiv γ t 1`, applies `mfderiv_eq_mfderiv_coord`, and closes against `Time.deriv_eq_mfderiv`. The §B smoothness bridge uses `mfderiv_eq_fderiv`.

`Time.deriv_eq_mfderiv` and `mfderiv_eq_fderiv` are existing Physlib/Mathlib lemmas whose statements are stable under internal changes to `mfderiv`. The dependence on the manifold-derivative internals is therefore confined to the differentiable branch of `mfderiv_eq_mfderiv_coord` rather than appearing in the public statement, and the `change … rfl` steps are removed. The unfold itself is not eliminated: the equality of the two `mfderiv`s would follow from the differential of `valDiffeomorph` being the identity on the model space — that is, its tangent map sending each tangent vector to itself under the identification `TangentSpace 𝓘(ℝ, E) x = E` — but no such result is available, so it reduces to the coordinate computation in that single lemma. If a future Mathlib bump breaks that computation, the breakage is contained to `mfderiv_eq_mfderiv_coord`, which I expect to be simpler to repair than the original proof, where the same unfolding appeared inline in both branches of the public statement.

---
*Developed with assistance from AI; all mathematics and proofs were reviewed and verified to compile.*


nateabr:
t-classical-mechanics

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/Trajectory.lean
```diff
@@ -0,0 +1,156 @@
...
+forgets the manifold structure and returns the corresponding physical-space position.
+-/
+
+/-- A trajectory of the oscillator: a curve in the configuration manifold `Q`. -/
+abbrev Trajectory := Time → ConfigurationSpace
+
+namespace Trajectory
+
+/-- The coordinate curve of a trajectory: its reading in the global Euclidean coordinate. -/
+def coord (γ : Trajectory) : Time → EuclideanSpace ℝ (Fin 1) := fun t => (γ t).val
```
jstoobysmith: I think this should be more properly defined using the coordinate map at `γ t`. Maybe not needed here, but I think this is what will generalize to other cases.

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/Trajectory.lean
```diff
@@ -0,0 +1,156 @@
...
+noncomputable def velocity (γ : Trajectory) (t : Time) :
+    TangentSpace 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) (γ t) :=
+  mfderiv 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) γ t
+    ((1 : Time) : TangentSpace 𝓘(ℝ, Time) t)
+
+/-- A trajectory is `MDifferentiableAt` a time iff its coordinate curve is, since the two
+differ by the coordinate diffeomorphism. -/
+private lemma mdifferentiableAt_iff_coord (γ : Trajectory) (t : Time) :
+    MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) γ t ↔
+      MDifferentiableAt 𝓘(ℝ, Time) 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) (coord γ) t := by
```
jstoobysmith: I think this latter one could just be `DifferentiableAt` (i.e. no `M`), also, I think this doesn't need to be private.

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/Trajectory.lean
nateabr: I have made the change, this definitely is the more accurate definition.

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/Trajectory.lean
nateabr: Made the changes :)

nateabr:
-awaiting-author

jstoobysmith (approved):
Great! Looks good. Approved.

Issue #1048, opened 2026-04-21
Removing pre-Lorentz vectors `Co`, `Contr` etc.


jstoobysmith:
The current implementation of Lorentz vectors in Lean, defines some preliminary modules and representations, these include `Co`, `Contr` etc. 

These can be removed and simply replaced with `Vector` and `CoVector`, with a careful seperation of definitions into the required modules. Similar for the complex case.

PR #1336, opened 2026-06-30
refactor(Tensors): remove pre-Lorentz Contr/Co reps in favour of ContrMod.rep/CoMod.rep

Robby955:
Towards #1048.

Implements what's flagged in `RealTensor/Vector/Pre/Basic.lean`: 

- removes the vestigial `Contr`/`Co` representation aliases and replaces every call site with `ContrMod.rep`/`CoMod.rep` directly, since `Contr d` was `Rep.of ContrMod.rep` (defeq-preserving throughout).

- dropped a duplicate `TopologicalSpace (Contr d)` instance that was a verbatim copy of the existing one on `ContrMod d`.

- `Contr.toCo`/`Co.toContr` (the intertwining maps) are untouched they were already stated against `ContrMod.rep`/`CoMod.rep`, not the aliases, so the names survive independently.

One note: the issue text also mentions replacing with `Vector`/`CoVector`. Should those be different, higher-level abstractions (`Lorentz.Vector d := Fin 1 ⊕ Fin d → ℝ`, a plain function type) from `ContrMod.rep`/`CoMod.rep` (`Representation` objects), or is it meant to be an equivalence or a renaming of the same thing? Migrating the `Pre/` machinery onto that carrier could be a separate, larger refactor after this. This PR follows what's already flagged in the file, but for now it is the narrower scope not a full closure yet.

### DIFF
--- a/Physlib/Relativity/Tensors/ComplexTensor/Vector/Pre/Basic.lean
+++ b/Physlib/Relativity/Tensors/ComplexTensor/Vector/Pre/Basic.lean
@@ -194,7 +194,7 @@ lemma complexContrBasis_of_real (i : Fin 1 ⊕ Fin 3) :
 
 lemma inclCongrRealLorentz_ρ (M : SL(2, ℂ)) (v : ContrMod 3) :
     (ContrℂModule.SL2CRep M) (inclCongrRealLorentz v) =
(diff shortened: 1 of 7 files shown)


jstoobysmith (approved):
Great! Approved! Many thanks

Issue #833, opened 2025-12-13
Lorentz algebra and Lorentz Group


jstoobysmith:
As suggested by @PrParadoxy in #832, the Lorentz algebra should be derived from the Lorentz group. 

- [ ] Include a TODO item in the Lorentz algebra file about this. 
- [ ] Introduce the Lorentz algebra from the Lorentz group.

PrParadoxy:
@jstoobysmith 
I also suggest a major change in the current implementation of `LorentzGroup `. Right now, it is defined as a bare set, a lot of lemmas and instances are created just to show it behaves as expected. But all of that is avoidable by defining LorentzGroup as a SubGroup of GL. I get topology, group structure, invariability, and so on for free with no effort. More importantly, it becomes easier to make LorentzGroup a LieGroup, and the reason is, LorentzGroup is manifestly a Subgroup of  GL. One does not have to go back and forth between LorentzGroup and GL for topological reasons through homorophisms. 


```lean
import Mathlib.Geometry.Manifold.Algebra.LieGroup
import PhysLean.Relativity.MinkowskiMatrix

open minkowskiMatrix
open Matrix

variable (d)

def LorentzGroup : Subgroup (GL (Fin 1 ⊕ Fin d) ℝ) where
  carrier := {Λ | (dual Λ.val) * Λ = 1}
  one_mem' := by simp
  mul_mem' := by
    simp only [Set.mem_setOf_eq, Units.val_mul, dual_mul]
    grind
  inv_mem' := by
    intro a h
    apply Matrix.inv_inj
    . replace h := congr_arg (· * a.val⁻¹) h
      simp only [mul_inv_cancel_right_of_invertible, one_mul] at h
      simp only [coe_units_inv, Matrix.mul_inv_rev, inv_inv_of_invertible, inv_one]
      rw [←h, dual_dual]
      simp
    . simp [Matrix.GeneralLinearGroup.det_ne_zero a]

#synth (TopologicalSpace (LorentzGroup d))
#synth (IsTopologicalGroup (LorentzGroup d))
#synth (∀ (a : LorentzGroup d), Invertible a.val.val)
#synth (Group (LorentzGroup d))
```

jstoobysmith:
I agree that if this change can be made and does not affect anything else downstream then we should make it. Is this something you would be happy to take a shot at @PrParadoxy ?

PrParadoxy:
[...]

PR #1337, opened 2026-06-30
feat(LorentzGroup): realize as a closed subgroup of GL

Robby955:
Partial progress on #833.

Follows the plan @PrParadoxy proposed on the issue thread: realize `LorentzGroup d` as a closed subgroup of `GL (Fin 1 ⊕ Fin d) ℝ` rather than a bare `Set`, as the route to a `LieGroup` instance.

Done here:
- `toGL_isClosedEmbedding`: promotes the existing `toGL_embedding` to a closed embedding, using the existing matrix-level closedness (`isClosed`) and the fact that the range of `toGL` is the preimage of `LorentzGroup d` under the continuous `Units.val` coercion.
- `toGLSubgroup`: `LorentzGroup d` as an actual `Subgroup (GL ...)` via `toGL.range`.
- `toClosedSubgroup`: the same, packaged as a `ClosedSubgroup`.

- getting the `LieGroup` instance itself needs Cartan's closed-subgroup theorem (a closed subgroup of a Lie group is an embedded Lie subgroup), and mathlib doesn't have it yet. There's no `LieSubgroup` type, no closed-embedding analogue of `IsOpenEmbedding.singletonChartedSpace`, and no infrastructure to derive a `ChartedSpace`/`IsManifold` from a closed embedding (the immersion/smooth-embedding scaffolding that does exist all presupposes a `ChartedSpace` on both sides already). `LorentzAlgebra` depends on that same structure, so it's blocked too for now.
- `API-map.yaml` updated to reflect this the closed-subgroup realization is marked done, "Instance as a Lie group" stays `false` with the blocker noted.

### DIFF
--- a/Physlib/Relativity/LorentzGroup/API-map.yaml
+++ b/Physlib/Relativity/LorentzGroup/API-map.yaml
@@ -24,7 +24,11 @@ Requirements:
     done: true
     location: Physlib/Relativity/LorentzGroup/Basic.lean (lorentzGroupIsGroup)
 
-  - description: Instance as a Lie group
+  - description: Realization as a closed subgroup of GL (closed embedding toGL, toGLSubgroup, toClosedSubgroup)
+    done: true
+    location: Physlib/Relativity/LorentzGroup/Basic.lean (toGL_isClosedEmbedding, toGLSubgroup, toClosedSubgroup)
+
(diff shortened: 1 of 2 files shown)


jstoobysmith (approved):
Looks good! Approved.

PR #1346, opened 2026-07-01
feat: Document succSuccAbove implementation tradeoff

NicolaBernini:
# Overview

Document succSuccAbove implementation tradeoff


jstoobysmith (approved):
Approved - looks good.

PR #1324, opened 2026-06-29
feat(Mathematics): add the hat map realizing ℝ³ as skew-symmetric 3×3 matrices

giuseppesorge:
## Summary

Toward #893 (API: Rigid body). Adds the **hat map** `[ω]ₓ` — the skew-symmetric `3 × 3` matrix realizing the cross product — the algebraic foundation for the rigid-body angular velocity (a later PR). This is general linear algebra, so it lives in `Physlib/Mathematics/`; it is independent of the other #893 PRs.

## Added — `Physlib/Mathematics/CrossProductMatrix.lean` (new file, `Matrix` namespace)

| Declaration | Meaning |
|---|---|
| `crossProductMatrix ω` | the hat matrix `[ω]ₓ = !![0,-ω 2,ω 1; ω 2,0,-ω 0; -ω 1,ω 0,0]` |
| `crossProductMatrix_mulVec` | `[ω]ₓ *ᵥ v = ω ⨯₃ v` (the defining property) |
| `crossProductMatrix_transpose` | `[ω]ₓᵀ = -[ω]ₓ` (skew-symmetry) |
| `crossProductVee A` | the vee map `![A 2 1, A 0 2, A 1 0]` |
| `crossProductVee_crossProductMatrix` | `vee (hat ω) = ω` (left inverse) |

One sorted `import` is added to `Physlib.lean`.

## Reviewer map

1. `crossProductMatrix` and `crossProductMatrix_mulVec` — the definition and its characterizing property.
2. `crossProductMatrix_transpose`.
3. `crossProductVee` and the round-trip lemma.

## Verification

- `lake build Physlib.Mathematics.CrossProductMatrix` — succeeds.
- `lake exe runLinter Physlib.Mathematics.CrossProductMatrix` — "Linting passed".
- `./scripts/lint-style.sh` — passes; no lines over 100 chars; `git diff --check` clean.
- `#print axioms` on the three lemmas — only `[propext, Classical.choice, Quot.sound]`.

Toward #893.

PR #1320, opened 2026-06-29
feat(ClassicalMechanics): add RigidBodyMotion with centre-of-mass velocity

giuseppesorge:
## Summary

Toward #893 (API: Rigid body). Adds the time-dependent configuration of a rigid body and the kinematics of its centre of mass — the first of the four open Landau–Lifshitz §32 items. This PR covers the **centre-of-mass velocity**; angular velocity, kinetic energy and Euler angles will follow in separate PRs.

## Added — `Physlib/ClassicalMechanics/RigidBody/Motion.lean` (new file)

| Declaration | Kind | Meaning |
|---|---|---|
| `RigidBodyMotion d` | structure, extends `RigidBody d` | a moving rigid body: the body plus the inertial-frame centre-of-mass trajectory `comTrajectory : Time → Space d` and orientation `orientation : Time → SO(d)` |
| `centerOfMassVelocity` | def | `∂ₜ comTrajectory` — the velocity `V` of the centre of mass (§32, `v = V + Ω × r`) |
| `centerOfMassVelocity_eq` | lemma | definitional unfolding |
| `centerOfMassVelocity_of_comTrajectory_const` | lemma | a stationary centre of mass has zero velocity |
| `linearMomentum` | def | `mass • centerOfMassVelocity` — total linear momentum `P = M V` |
| `linearMomentum_eq` | lemma | definitional unfolding |

One sorted `import` line is added to `Physlib.lean`.

## Reviewer map

1. `RigidBodyMotion` — the configuration data type (the core contribution).
2. `centerOfMassVelocity` and its two lemmas.
3. `linearMomentum` and its lemma.

`orientation` is intentionally included now even though it is unused in this PR: it is the body's time-dependent rotation about the centre of mass, and is the foundation for the angular-velocity and kinetic-energy follow-up PRs toward #893.

## Scope — what is not changed

No existing declaration is modified; this only adds one new file plus one import. The corresponding `informal_*` stubs in `RigidBody/Basic.lean` are left in place.

## Verification

- `lake build Physlib.ClassicalMechanics.RigidBody.Motion` — succeeds.
- `lake exe runLinter Physlib.ClassicalMechanics.RigidBody.Motion` — zero errors originating from the new file.
- `./scripts/lint-style.sh` — passes; no trailing whitespace, no lines over 100 chars; `git diff --check` clean.
- `#print axioms` on all four new declarations — only `[propext, Classical.choice, Quot.sound]`.

Toward #893.

Issue #893, opened 2026-01-09
API: Rigid body


jstoobysmith:
## Key data structure

The key data structure shall be the definition of a Rigid Body specified by  its mass distribution. 

- [x] is defined

## Need

This API is needed to study many examples in classical mechanics, for example rigid spheres.

## Requirements

- [x] The API shall allow the calculation of the total mass of the rigid body.
- [x] The API shall allow the calculation of the center of mass of the rigid body. 
- [x] The API shall allow the calculation of the moment of inertia of the rigid body. 
- [ ] The API shall contain the definition of the angular velocity of rotation of the rigid body. 
- [ ] The API shall contain the definition of the velocity of the center of mass of the rigid body. 
- [ ] The API shall contain the definition of the kinetic energy of the rigid body. 
- [ ] The API shall contain the definition of Euler angles for a rigid body.

## Corresponding file system

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/ClassicalMechanics/RigidBody

## References 

- Landau and Lifshitz, Mechanics, Section 32. 

## Parent APIs 

#854 #855

giuseppesorge:
Progress update on this API. Two PRs are currently open:

- **#1320** — the rigid-body motion configuration and centre-of-mass kinematics: `RigidBodyMotion` (body + centre-of-mass trajectory + time-dependent orientation), the velocity of the centre of mass, the linear momentum, and the time-dependent mass distribution (with mass conservation and centre-of-mass tracking as sanity checks).
- **#1324** — the `so(3)` hat map `[ω]ₓ` (the skew-symmetric matrix realizing the cross product), the algebraic foundation needed for the angular velocity.

The remaining requirements — the **angular velocity** (`Ω = Ṙ Rᵀ`, with the vector `ω` recovered via the hat/vee map), the **kinetic energy**, and the **Euler angles** — build directly on the `RigidBodyMotion` structure and the hat map. To avoid stacking PRs on top of unmerged work, I'll hold off on them until #1320 and #1324 are merged, then continue in that order.

PR #1353, opened 2026-07-01
feat(ClassicalMechanics): add rigid-body angular velocity tensor

giuseppesorge:
## Summary

Adds the **angular velocity tensor** of a rigid body in motion, `Ω(t) = Ṙ(t) R(t)ᵀ` where
`R(t) = orientation t` (Landau–Lifshitz, *Mechanics* §31), building on `RigidBodyMotion` from #1320.

New declarations in `Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean`:
- `RigidBodyMotion.orientationMatrix` — the orientation as a matrix-valued function of time.
- `RigidBodyMotion.angularVelocityTensor` — `Ω(t) = Ṙ(t) R(t)ᵀ`.
- `angularVelocityTensor_transpose` — **litmus test**: `Ω` is skew-symmetric (`Ωᵀ = -Ω`, i.e.
  `Ω ∈ 𝔰𝔬(d)`), obtained by differentiating the orthogonality identity `R Rᵀ = 1`.
- `angularVelocityTensor_of_orientation_const` — a body of constant orientation has `Ω = 0`.

The supporting general matrix-calculus lemmas live in a new general file
`Physlib/SpaceAndTime/Time/MatrixDerivatives.lean` (kept out of the classical-mechanics file, per the
AGENTS.md file-placement guideline):
- `Time.deriv_matrix_mul` — product rule for `∂ₜ (A * B)`.
- `Time.deriv_matrix_transpose` — `∂ₜ` commutes with transpose.
- `Matrix.transposeCLM` — transpose as a continuous linear map on square real matrices.

The `d = 3` angular-velocity *vector* `ω` (via the hat/vee map from #1324) and the identity
`[ω]ₓ = Ω` are left to a follow-up.

Toward #893.

### DIFF
--- a/Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean
+++ b/Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean
@@ -36,17 +36,9 @@ namespace RigidBodyMotion
 
 variable {d : ℕ}
 
-/-- The orientation of the body as a matrix-valued function of time: the underlying matrix of
-`orientation`. -/
-noncomputable def orientationMatrix (M : RigidBodyMotion d) : Time → Matrix (Fin d) (Fin d) ℝ :=
-  fun s => (M.orientation s).val
-
-lemma orientationMatrix_apply (M : RigidBodyMotion d) (t : Time) :
-    M.orientationMatrix t = (M.orientation t).val := rfl
-
 /-- The orientation matrix is special orthogonal, so `R Rᵀ = 1`. -/
-lemma orientationMatrix_mul_transpose (M : RigidBodyMotion d) (t : Time) :
-    M.orientationMatrix t * (M.orientationMatrix t)ᵀ = 1 :=
+lemma orientation_mul_transpose (M : RigidBodyMotion d) (t : Time) :
+    (M.orientation t).1 * ((M.orientation t).1)ᵀ = 1 :=
   (mem_orthogonalGroup_iff (Fin d) ℝ).mp
     (mem_specialOrthogonalGroup_iff.mp (M.orientation t).2).1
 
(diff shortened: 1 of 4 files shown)


Physlib/SpaceAndTime/Time/MatrixDerivatives.lean
```diff
@@ -0,0 +1,69 @@
...
+
+attribute [local instance] Matrix.linftyOpNormedAddCommGroup Matrix.linftyOpNormedSpace
+  Matrix.linftyOpNormedRing Matrix.linftyOpNormedAlgebra
+
+namespace Matrix
+
+variable {d : ℕ}
+
+/-- Transpose as a continuous linear map on square real matrices. -/
+noncomputable def transposeCLM :
```
jstoobysmith: If we can help it i wouldn't make this definition, I would just use what is already in Mathlib where needed

Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean
```diff
@@ -0,0 +1,95 @@
...
+attribute [local instance] Matrix.linftyOpNormedAddCommGroup Matrix.linftyOpNormedSpace
+  Matrix.linftyOpNormedRing Matrix.linftyOpNormedAlgebra
+
+namespace RigidBodyMotion
+
+variable {d : ℕ}
+
+/-- The orientation of the body as a matrix-valued function of time: the underlying matrix of
+`orientation`. -/
+noncomputable def orientationMatrix (M : RigidBodyMotion d) : Time → Matrix (Fin d) (Fin d) ℝ :=
```
jstoobysmith: If you can get away without this, and just use `(M.orientation s).1` everywhere, I think that would be better (it may even be shorter).

jstoobysmith (commented):
Some comments, but otherwise looks good.

Physlib/SpaceAndTime/Time/MatrixDerivatives.lean
giuseppesorge: Removed. The lemmas now use `(Matrix.transposeLinearEquiv (Fin d) (Fin d) ℝ ℝ).toLinearMap.toContinuousLinearMap` from Mathlib directly. Mathlib has no differentiability analogue of `Continuous.matrix_transpose`, so that step is a lemma (`DifferentiableAt.matrix_transpose`, stated for a general normed domain) rather than a definition.

Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean
giuseppesorge: Done - `orientationMatrix` is gone and everything is stated on `(M.orientation s).1` directly. You were right, it is shorter: net -13 lines.

giuseppesorge:
-awaiting-author

Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean
```diff
@@ -0,0 +1,86 @@
...
+
+attribute [local instance] Matrix.linftyOpNormedAddCommGroup Matrix.linftyOpNormedSpace
+  Matrix.linftyOpNormedRing Matrix.linftyOpNormedAlgebra
+
+namespace RigidBodyMotion
+
+variable {d : ℕ}
+
+/-- The orientation matrix is special orthogonal, so `R Rᵀ = 1`. -/
+lemma orientation_mul_transpose (M : RigidBodyMotion d) (t : Time) :
```
jstoobysmith: One last thing, I would now move this to the file where orientation is defined (maybe the `Basic.lean` file)?

giuseppesorge: Done - moved to Motion.lean, which is where the `orientation` field of `RigidBodyMotion` is defined (Basic.lean only has the static `RigidBody`, with no orientation).

giuseppesorge:
-awaiting-author

jstoobysmith (approved):
Approved - looks good to me. Will merge shortly.

Issue #905, opened 2026-01-14
API:  Damped harmonic oscillator


jstoobysmith:
## Key data structure

The key data structure shall be the input data of the damped harmonic oscillator. Namely, the mass, spring constant and damping coefficient. 

- [x] is defined 
## Need

The damped harmonic oscillator is a key example of a classical mechanical system, and a simple generalisation of the harmonic oscillator. 

## Requirements

- [ ] The API shall contain the definition of the kinetic energy. 
- [ ] The API shall contain the definition of the potential energy. 
- [ ] The API shall contain the definition of the lagrangian. 
- [ ] The API shall contain a simplification of the variational gradient of the lagrangian to the Euler-Lagrange equations. 
- [x] The API shall contain a definition of the equation of motion of the system. 
- [x] The API shall contain a definition of the initial conditions of the damped harmonic oscillator. 
- [x] The API shall contain the definition of a trajectory of the damped harmonic oscillator. 

## Corresponding file system

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/ClassicalMechanics/DampedHarmonicOscillator

## Parent APIs 

#846

FloWsnr:
Hey @jstoobysmith do we need to update this? I just now stumbled upon this issue.

jstoobysmith:
Hey @FloWsnr, yeah we should  - I've updated the things I think are now included! Though these issues never really took off, and it is maybe best to try and construct a replacement for them (in the file system themselves).

FloWsnr:
I agree. I mean, there are the TODOs which already represent the unchecked boxes, but I guess there is nothing yet to document what is already present, correct?

FloWsnr:
P.s.
- The API shall contain the definition of the kinetic energy.
- The API shall contain the definition of the potential energy.

These should be checked now since each Damped Oscillator is now extending the HarmonicOscillator class for which the energies are already defined and proven.

giuseppesorge:
I'd like to take the remaining requirements of this API — the lagrangian and the simplification of its variational gradient to the Euler-Lagrange equations — following the structure already used in `HarmonicOscillator/Basic.lean` (sections D.1–D.3).

Since the damped oscillator is non-conservative, the plain `T − V` lagrangian doesn't reproduce the damped equation of motion. My plan is to use the Caldirola–Kanai lagrangian

[...]

PR #1354, opened 2026-07-01
feat(ClassicalMechanics): add Caldirola-Kanai lagrangian for the damped harmonic oscillator

giuseppesorge:
## Summary

Adds the Caldirola–Kanai lagrangian for the damped harmonic oscillator and the simplification of its variational gradient to the Euler-Lagrange equations, completing two open requirements of #905. Since the damped oscillator is non-conservative, the plain `T − V` lagrangian does not reproduce the damped equation of motion; the Caldirola–Kanai lagrangian `exp (γ/m * t) * (T − V)` does: its Euler-Lagrange equation is exactly `m ẍ + γ ẋ + k x = 0`, the `EquationOfMotion` already in the file.

The new section F mirrors the structure of `HarmonicOscillator/Basic.lean` (sections D.1–D.3).

## Changes

| Declaration | File | Change |
|---|---|---|
| `lagrangian` | `DampedHarmonicOscillator/Basic.lean` | new def: `exp (γ/m * t) *` undamped lagrangian |
| `lagrangian_eq`, `lagrangian_eq_exp_mul_kineticEnergy_sub_potentialEnergy`, `lagrangian_of_isUndamped` | 〃 | equalities; reduction to the undamped lagrangian at `γ = 0` |
| `contDiff_lagrangian` (`@[fun_prop]`) | 〃 | smoothness in all arguments |
| `gradient_const_mul` (private) | 〃 | helper: `gradient (c * f) = c • gradient f` |
| `gradient_lagrangian_position_eq`, `gradient_lagrangian_velocity_eq` | 〃 | gradients, reusing the `HarmonicOscillator` gradient lemmas |
| `gradLagrangian` | 〃 | new def: variational gradient `δ ∫ L` of the action |
| `gradLagrangian_eq_eulerLagrangeOp` | 〃 | via `euler_lagrange_varGradient` |
| `deriv_exp_smul` (private) | 〃 | `∂ₜ (exp (a t) • y t)` product rule |
| `gradLagrangian_eq_force` | 〃 | simplification to `exp (γ/m * t) • (F − m a)` |
| `equationOfMotion_iff_gradLagrangian_zero` | 〃 | EoM ⟺ variational gradient vanishes |
| `Time.val_contDiff` (`@[fun_prop]`) | `SpaceAndTime/Time/Derivatives.lean` | completes the `val_measurable`/`val_differentiable` API; needed for `contDiff_lagrangian` (section D retitled "Smoothness properties" to cover it) |
| requirements | `DampedHarmonicOscillator/API-map.yaml` | lagrangian + variational-gradient marked done; kinetic/potential energy marked done as inherited from `HarmonicOscillator` (per the discussion on #905); Overview and References synced |

Reviewer reading order: `Time.val_contDiff` → `lagrangian` → F.1.1–F.1.3 → `gradLagrangian` → `gradLagrangian_eq_force` → `equationOfMotion_iff_gradLagrangian_zero`.

## Scope — what is NOT changed

- No existing statement, definition, or proof is changed; besides the new declarations, only the `Time/Derivatives.lean` section-D title and the API-map prose are touched.
- The variational-gradient formula in `HarmonicOscillator/Basic.lean` section D.2 transposes the position and velocity arguments of `L(t, x, v)` (doc-only typo). The new section F.2 here states it correctly; I can fix the `HarmonicOscillator` docstring in a small follow-up rather than widen this diff.
- `Solution.lean` has a private `exp_decay_smul_velocity` similar in spirit to the private `deriv_exp_smul` added here (decay vs growth sign, function-level vs pointwise). Unifying them would mean touching `Solution.lean`; happy to do that as a follow-up if preferred.
- The uniqueness-of-solution requirement of #905 is untouched (tracked by the existing TODO in `Solution.lean`).

On size: the diff is 236 gross lines, but roughly half is module-doc/section-header text required by the numbered template; the net Lean code is ~110 lines forming one coherent concept (the CK lagrangian and its Euler-Lagrange derivation), which loses its point if split.

## Verification

- `lake build` (full library): success; touched modules re-built clean after the final docs pass.
- `lake exe lint_all`: passes (the "transitive imports" shake messages are the known pre-existing false positives, incl. on `Time/Derivatives.lean` before this change).
- `./scripts/lint-style.sh` (on the committed state): clean.
- `#print axioms` on all 12 new public declarations: only `[propext, Classical.choice, Quot.sound]`.

Toward #905.

### DIFF
--- a/Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
+++ b/Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
@@ -87,15 +87,11 @@ Requirements:
 
   - description: The API shall contain the definition of the kinetic energy.
     done: true
-    location: >
-      Inherited from Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (kineticEnergy),
-      since DampedHarmonicOscillator extends HarmonicOscillator.
+    location: Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (kineticEnergy)
 
   - description: The API shall contain the definition of the potential energy.
     done: true
-    location: >
-      Inherited from Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (potentialEnergy),
-      since DampedHarmonicOscillator extends HarmonicOscillator.
+    location: Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (potentialEnergy)
 
   - description: The API shall contain the definition of the Lagrangian.
     done: true
--- a/Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
+++ b/Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
@@ -15,8 +15,10 @@ Overview: |
     equivalence with Newton's second law, the rate of mechanical-energy dissipation
     along solutions, the three damping regimes (underdamped, critically damped,
     overdamped) selected by the sign of the discriminant `γ² - 4 m k` together with
-    the decay rate and regime-selected angular frequency, and the reduction to the
-    undamped harmonic oscillator when `γ = 0`. The `Solution` module defines the
+    the decay rate and regime-selected angular frequency, the reduction to the
+    undamped harmonic oscillator when `γ = 0`, and the Caldirola–Kanai lagrangian
+    `exp (γ/m * t) * (T - V)`, whose variational gradient vanishes if and only if
+    the equation of motion holds. The `Solution` module defines the
     initial conditions, the explicit closed-form `trajectory` selected from the
     damping regime (trigonometric, polynomial, or hyperbolic, each times an
     exponential decay factor), and proves that the selected trajectory satisfies the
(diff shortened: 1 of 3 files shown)


Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
```diff
@@ -82,22 +86,28 @@ Requirements:
     location: Physlib/ClassicalMechanics/DampedHarmonicOscillator/Solution.lean (trajectory_equationOfMotion)
 
   - description: The API shall contain the definition of the kinetic energy.
-    done: false
-    location: N/A
+    done: true
+    location: >
+      Inherited from Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (kineticEnergy),
+      since DampedHarmonicOscillator extends HarmonicOscillator.
```
jstoobysmith: I would keep this as: 

```
Physlib/ClassicalMechanics/HarmonicOscillator/Basic.lean (kineticEnergy)
```
just because I think we will likely use a script at some point to make sure these definitions go to a place which actually exists. (likewise with below).

jstoobysmith (commented):
Great! This is really nice! Just a couple of convention comments.

Physlib/ClassicalMechanics/DampedHarmonicOscillator/API-map.yaml
giuseppesorge: Done - both inherited-energy `location` fields now use the plain `file (name)` format; the inheritance rationale stays in the Overview.

giuseppesorge:
-awaiting-author

jstoobysmith (approved):
Approved - will merge shortly. Many thanks.

PR #1355, opened 2026-07-02
feat: add /check-golf bot verifying statements are unchanged

Vilin97:
## What

Adds a `/check-golf` bot that verifies a PR only *golfs* proofs: that no declaration **statement** (its signature/type) changed and only proofs and definition bodies changed. Comment `/check-golf` on any PR and the bot posts its findings as a single comment, editing that same comment on subsequent runs.

## Files

- `scripts/check_golf.py` — compares two revisions of the repo. For every named declaration in the changed `.lean` files it decides whether the statement was preserved and only the proof/body changed. It also renders the Markdown report and upserts it as a PR comment (POST first, PATCH the marked comment afterwards).
- `.github/workflows/check-golf.yml` — triggers on an `issue_comment` containing `/check-golf` on a PR, resolves the merge-base and head, runs the script, and comments.
- `scripts/README.md` — documents the tool and how to run it locally.

## How it works

The Lean sources are parsed **textually** (no build required, so the check is fast):

- comments and string literals are blanked, whitespace is ignored;
- top-level declarations are segmented on **column-0** command starts (robust even when a proof body contains brackets that cannot be balanced textually, e.g. a `by` block closed by indentation);
- each statement is split from its body at the top-level `:=` / `where` / equation `|` arm, so `:=` in default binders and `by` blocks *inside* a type (`⟨x, by omega⟩`) do not confuse the split;
- `by` proof terms embedded inside a type are treated as proof-irrelevant, so golfing them counts as a proof change, not a statement change.

Anonymous instances and `example`s are not tracked (they have no stable name).

## Testing

Run locally against two revisions:

```
scripts/check_golf.py --base <merge-base> --head <head-sha>
```

Validated against a large real golf PR: the checker reports **0 statement changes** across 312 changed files, and a name-level cross-check confirms no named declaration is dropped by the parser. The post→edit comment logic is covered by a mocked-API test.

PR #1356, opened 2026-07-02
feat: report golf compile cost (heartbeats + time) in /check-golf

Vilin97:
**Stacked on #1355 — please review that first.** This PR adds one commit on top of #1355; its net diff over `master` therefore also contains #1355's script. Once #1355 merges, this becomes a small delta.

## What

Extends the `/check-golf` bot so its comment also reports the **compile cost** of a golf: for every changed file it compiles the base and head versions and diffs their **heartbeats** and **wall-clock time**.

## Changes (on top of #1355)

- `scripts/check_golf.py`: new `--measure` flag.
  - Instruments each file by importing `Mathlib.Util.CountHeartbeats`, forcing `set_option Elab.async false`, and prefixing `#count_heartbeats in` before every declaration (placed above any attached doc comment so the declaration stays well-formed).
  - Heartbeats are deterministic; wall-clock time is reported as a secondary, noisier signal.
  - Adds a "Compile cost (base → head)" section: totals (with %) and a table of the largest heartbeat movers.
- `.github/workflows/check-golf.yml`: builds the head revision (`lake exe cache get` + `lake build`) before measuring, then runs with `--measure`. The base sources are compiled against the built head environment, which is sound because the check has already verified the statements are unchanged.
- `scripts/README.md`: documents `--measure`.

## Why the async gymnastics

`#count_heartbeats` only sees ~0 heartbeats under Lean's default async elaboration (the proof runs in a background task). `set_option Elab.async false` makes the count accurate — confirmed by measuring real proofs (e.g. a `ring` proof reports ~1.5k heartbeats instead of ~40).

## Testing

Validated locally on golfed files, e.g. `HarmonicOscillator/Solution.lean`: base **71,914** → head **71,979** heartbeats (+65) and 7.39s → 7.17s — i.e. this golf left compile cost essentially unchanged.

### DIFF
--- a/.github/workflows/check-golf.yml
+++ b/.github/workflows/check-golf.yml
@@ -1,11 +1,12 @@
 # Reacts to a `/check-golf` comment on a pull request. It verifies that the PR
 # only golfs proofs -- i.e. that no declaration *statement* (signature/type)
-# changed, only proofs and bodies -- and posts the findings as a single PR
-# comment, editing that comment in place on subsequent runs.
+# changed, only proofs and bodies -- and reports the compile-cost of the golf
+# (heartbeats and time, base vs head). The findings are posted as a single PR
+# comment, edited in place on subsequent runs.
 #
 # The check runs on `issue_comment`, so it uses the workflow file from the
-# default branch and only parses the PR's Lean sources textually (it does not
-# build them).
+# default branch. Measuring compile cost needs oleans, so this workflow builds
+# the head revision before measuring.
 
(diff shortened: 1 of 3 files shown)


jstoobysmith:
@Vilin97 Merge conflict with this now. 

awaiting-author

Vilin97:
-awaiting-author

jstoobysmith (approved):
Approved. Will merge now

PR #1357, opened 2026-07-02
feat(PhyslibAlpha/Particles): The effective potential of the 2HDM

jstoobysmith:
The effective potential of the 2HDM model, formalised using AI (Claude Opus 4.8) taking about 10 hours. Some commits have the prompts which lead to them in their comments (the earlier ones do not). 

I wrote the initial theorem and gave some minor guidance when it got stuck.


zhikaip (approved):
approved

PR #1359, opened 2026-07-03
feat(QuantumMechanics): Add TODOs

gloges:
Adds a number of quantum mechanics TODOs.
Some are currently possible to complete (e.g. the operator examples), but others depend on areas such as (degenerate) perturbation theory which are currently undeveloped and can help guide further development.

Q: Is there a better way to refer to other TODOs? In one of the momentum TODOs I have used the hash of the other.

### DIFF
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -274,6 +274,7 @@ public import Physlib.QuantumMechanics.DDimensions.Hydrogen.LaplaceRungeLenzVect
 public import Physlib.QuantumMechanics.DDimensions.Operators.AngularMomentum
 public import Physlib.QuantumMechanics.DDimensions.Operators.Commutation
 public import Physlib.QuantumMechanics.DDimensions.Operators.Covariance
+public import Physlib.QuantumMechanics.DDimensions.Operators.Examples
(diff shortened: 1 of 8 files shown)


gloges:
t-quantum-mechanics

jstoobysmith (approved):
Nice! Looks good. APproved.

jstoobysmith:
```quote 
Is there a better way to refer to other TODOs? In one of the momentum TODOs I have used the hash of the other.
```
The hash is probably the unique way to identify them, although we should probably make a tool which allows people to find TODOs by their hashes.

gloges:
Thanks! Perhaps TODOs could have an (optional) second ID in addition to the hash which is user-picked and human readable, but not necessarily unique. e.g. something like "QM-momentum-domain" for the one I mentioned. The non-uniqueness may even be a _pro_ since it allows related TODOs to be found simultaneously in a search.

jstoobysmith:
Yeah agreed! For now you could probably preface your TODO string with "QM-momentum-domain: ....", and use that instead. Won't merge just yet in case you want to do this.

gloges:
I think it's okay for now - merge away!

PR #1369, opened 2026-07-05
feat(ClassicalMechanics): add geometric kinetic energy for the harmonic oscillator

nateabr:
I add a geometric notion of kinetic energy for the harmonic oscillator in a new file, building on the geometric `ConfigurationSpace`. The metric-induced kinetic energy is defined coordinate-free and parametric in any pseudo-Riemannian metric, and I then construct the oscillator's own mass metric and show it reproduces the standard KE formula.

Some challenges I ran into:

- **No inner product on tangent spaces.** My first version wrote `⟪v, w⟫` directly on velocities, which doesn't compile, as Mathlib deliberately withholds inner-product/normed instances from `TangentSpace`. I resolved this with an explicit coordinate identification `tangentCoord`. I tried installing instances but the instance route creates diamonds.
- **Smoothness of the metric.** `fun_prop` couldn't discharge this, because it can't see that the global chart inverse has identity derivative. I added a dedicated lemma establishing exactly that, which collapses the metric to a constant in the chart, after which smoothness is immediate.
- **Declaration ordering / defeq.** The tangent identification and its finite-dimensional instance had to precede `geometricKineticEnergy` (whose type already demands them), and I needed a file-scoped `set_option backward.isDefEq.respectTransparency false` for the tangent≡model identification.

---
*Developed with assistance from AI; all mathematics and proofs were reviewed and verified to compile.*

### DIFF
--- a/Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
+++ b/Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
@@ -7,64 +7,44 @@ module
 
 public import Physlib.ClassicalMechanics.HarmonicOscillator.Basic
 public import Physlib.ClassicalMechanics.HarmonicOscillator.Geometric.Basic
-public import Physlib.Mathematics.Geometry.Metric.PseudoRiemannian.Defs
-public import Mathlib.Geometry.Manifold.MFDeriv.Atlas
+public import Mathlib.Geometry.Manifold.VectorBundle.Riemannian
 
 /-!
 # Geometric kinetic energy of the harmonic oscillator
 
 ## i. Overview
 
-The configuration space `Q` of the geometric harmonic oscillator is `ConfigurationSpace`.
-Velocities lie in the tangent bundle `TQ`; at a configuration `q`, a velocity is a tangent
-vector in `TangentSpace I q`.
+The configuration space of the geometric harmonic oscillator is `ConfigurationSpace`.
+At a configuration `q`, velocities are tangent vectors in
+`TangentSpace 𝓘(ℝ, EuclideanSpace ℝ (Fin 1)) q`.
 
-Given a pseudo-Riemannian metric `g` on `Q`, the definition of the kinetic-energy function
-on `TQ` is given by `geometricKineticEnergy g q v = 1 / 2 * g.inner q v v`. This is the
-metric-induced kinetic term used for natural mechanical systems.
+The oscillator mass determines a Riemannian metric on `ConfigurationSpace`. At each
+configuration `q`, the metric is the mass-scaled Euclidean inner product on tangent
+vectors, recorded by `massMetricVal S q`.
 
-For the oscillator mass metric, tangent vectors are first read in the chosen global
-Euclidean coordinate. The map `tangentCoord q : TangentSpace I q ≃L[ℝ] Model` is the
-chart-induced continuous linear equivalence from the tangent space at `q` to the Euclidean
-coordinate model `Model := EuclideanSpace ℝ (Fin 1)`. We use it to form the pullback of
-the Euclidean inner product to `TangentSpace I q`.
-
-The mass metric `massPseudoRiemannianMetric S` is the mass-scaled pullback of the Euclidean
-inner product along `tangentCoord q`: `S.m * ⟪tangentCoord q v, tangentCoord q w⟫_ℝ`.
-Positivity of `S.m`, together with positive-definiteness of the Euclidean inner product on
-`Model`, gives nondegeneracy and negative index `0`.
-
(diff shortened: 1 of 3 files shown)


nateabr:
t-classical-mechanics

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
```diff
@@ -0,0 +1,302 @@
...
+the same Euclidean coordinate model. The continuous linear equivalence `tangentCoord q`
+gives the coordinate representative of a tangent vector at `q`; it does not install an
+intrinsic inner-product structure on tangent spaces.
+-/
+
+-- Let Lean use the definitional tangent/model identification used by `tangentCoord`
+set_option backward.isDefEq.respectTransparency false
+
+/-- The Euclidean coordinate model for the harmonic oscillator configuration space. -/
+abbrev Model := EuclideanSpace ℝ (Fin 1)
```
jstoobysmith: I think abbreviation is wrong here, I think you should be using local notation.

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
```diff
@@ -0,0 +1,302 @@
...
+
+/-- The model-with-corners structure for the global chart on `ConfigurationSpace`. -/
+abbrev I := 𝓘(ℝ, Model)
+
+/-- The chart-induced continuous linear equivalence from the tangent space at `q` to the
+Euclidean coordinate model. It sends a tangent vector to its coordinate representative in
+`Model`. -/
+def tangentCoord (q : ConfigurationSpace) :
```
jstoobysmith: This should likely go in the `Basic.lean` file

jstoobysmith: Along with the results below (they are not about the kinetic energy)

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
```diff
@@ -0,0 +1,302 @@
...
+
+A pseudo-Riemannian metric `g` on `Q` assigns a bilinear form `g_q` to each tangent space
+`TangentSpace I q`. The kinetic-energy function associated to `g` is
+`K_g(q, v) = 1 / 2 * g_q(v, v)`.
+-/
+
+/-- For a pseudo-Riemannian metric `g` on `Q`, the kinetic-energy function
+`K_g(q, v) = 1 / 2 * g.inner q v v` on the tangent bundle `TQ`. -/
+noncomputable def geometricKineticEnergy
+    (g : PseudoRiemannianMetric
```
jstoobysmith: I think Mathlib now has `RiemannianMetric`, so we should probably use this. But should we define an explicit version of this on `ConfigurationSpace` rather than keeping it a free variable (i.e. move below the definition of the metric).

Physlib/ClassicalMechanics/HarmonicOscillator/Geometric/KineticEnergy.lean
```diff
@@ -0,0 +1,302 @@
...
+  apply QuadraticForm.rankNeg_eq_zero
+  intro v hv
+  change 0 < massMetricVal S q v v
+  rw [massMetricVal_apply]
+  exact mul_pos S.m_pos (real_inner_self_pos.mpr (tangentCoord_ne_zero hv))
+
+/-- The mass-scaled pullback of the Euclidean inner product, as a pseudo-Riemannian metric on
+the oscillator configuration space. -/
+noncomputable def massPseudoRiemannianMetric (S : HarmonicOscillator) :
+    PseudoRiemannianMetric
```
jstoobysmith: Would probably make this a RiemannianMetric. Also would collapse this down to one line, instead of having the arguments spread across different lines.

nateabr:
Thank you for your comments @jstoobysmith. I hope I have addressed them in the latest commit.

I have moved the parts not pertaining to KE into basic.lean. 

Moving to the mathlib remannian metric required a significant rewrite and a new set of proof obligations, therefore I would appreciate if you could take another look to see if I've done things correctly.

I do believe that it was the right design choice to switch as it did not lead to any diamonds, maybe we should rethink physlib's own implementation of the metric in light of this.

-awaiting-author

jstoobysmith (approved):
This looks good! We probably actually want such a metric on things like Space aswell. I agree we need to rethink the pseudo-reimannian metric in Physlib.

nateabr:
I will try to add a todo with regards to this after scoping out the relevant files.

Thank you for your review.

PR #1371, opened 2026-07-06
auto-task(generalizedKroneckerDelta_sum_snoc): golf the snoc contraction proof

doxtor6:
## Summary

Golfs the proof of `generalizedKroneckerDelta_sum_snoc` in
`Physlib/Mathematics/KroneckerDelta/Contraction.lean` from 95 lines down to 46
(net −49 lines), keeping the statement byte-for-byte identical and following the
same mathematical route (Schur-complement bordering + the file's rank-one
determinant update `Matrix.det_add_rankOne`).

## What changed

Only the tactic block of `generalizedKroneckerDelta_sum_snoc` was rewritten; no
other declaration, statement, or file is touched.

- The 40-line manual block-matrix identification (`hblk`) is collapsed to a
  4-case `ext (i | i) (j | j)` with one-line `simp` discharges. The key trick is
  rewriting `Fin.snoc μ a` to `Fin.append μ (fun _ => a)` via
  `← Fin.append_right_eq_snoc`, so that the `@[simp]` lemmas `Fin.append_left` /
  `Fin.append_right` handle the `Fin.castAdd`/`Fin.natAdd` indices produced by
  `finSumFinEquiv` directly (the old proof needed manual `hcast`/`hlast` casts and
  a 4-way `cases … with` tree).
- `step1`/`step2` are merged into a single `have key`, with the repeated
  δ-column/row lambdas abstracted once via `set b`/`set B`/`set C`, and the
  `det`-level bridge done by a single `show … from (Matrix.det_submatrix_equiv_self _ _).symm`
  instead of a separate `rfl`-rewrite plus `← det_submatrix_equiv_self` pattern rewrite.
- The 10-line `hkey` (killing the δ-weighted sum over the shared index) becomes a
  2-tactic `hrow`: `simp_rw [← nsmul_eq_mul, KroneckerDelta.sum_smul]` reuses the
  file's own `@[simp]` sum lemma instead of a manual `Finset.sum_eq_single` case
  split.
- The final 14-line `calc` bookkeeping is replaced by one `rw` chain plus a
  closing `simp only [nsmul_eq_mul, ← sub_mul]` (no `ring` needed), and the
  unused `classical` is dropped.

## Verification

- `lake build` succeeds: `Build completed successfully (9121 jobs)`, with the
  golfed file compiling warning-free and all downstream users (the Levi-Civita
  epsilon-epsilon contraction identities) unchanged.
- Elaboration cost is unchanged (tactic execution 230ms → 190ms per
  `-Dprofiler=true` on the file; wall time dominated by imports in both cases).

### DIFF
--- a/Physlib/Mathematics/KroneckerDelta/Contraction.lean
+++ b/Physlib/Mathematics/KroneckerDelta/Contraction.lean
@@ -159,97 +159,48 @@ This is the reusable combinatorial fact behind all epsilon-epsilon identities. -
 lemma generalizedKroneckerDelta_sum_snoc {n : ℕ} (μ ν : Fin n → α) :
     ∑ a : α, generalizedKroneckerDelta (Fin.snoc μ a) (Fin.snoc ν a)
       = ((Fintype.card α : ℤ) - n) * generalizedKroneckerDelta μ ν := by
-  classical
   set A : Matrix (Fin n) (Fin n) ℤ :=
     Matrix.of fun i j => ((kroneckerDelta (μ i) (ν j) : ℕ) : ℤ) with hA
-  -- The `n × n` δ-matrix is exactly `generalizedKroneckerDelta μ ν`.
-  have hAdet : A.det = generalizedKroneckerDelta μ ν := rfl
-  -- Step 1: Schur complement. Border the matrix with the appended index and reduce dimension.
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved.

PR #1379, opened 2026-07-07
feat: Add action of gauge group on Quark Doublets

jstoobysmith:
Adding the action of the gauge group (with possible quotients) on the quark doublets, extending the result related to the action of the full gauge group.

### DIFF
--- a/Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
+++ b/Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
@@ -185,6 +185,9 @@ lemma repGaugeGroupI_eq_iff_mul_eq {g1 g2 : GaugeGroupI} :
     simp only [EuclideanSpace.basisFun_apply, Fin.sum_univ_two, Fin.isValue] at h1 h2
     simp [valLinEquiv_symm_apply, h1, h2, b, h]
 
+TODO "Improve the efficiency of `mem_repGaugeGroupI_ker_iff_eq` by removing the
+  `grind`s and replacing them with a more direct argument."
+
 lemma mem_repGaugeGroupI_ker_iff_eq {g : GaugeGroupI} :
     g ∈ repGaugeGroupI.ker ↔ ∃ a b : ℂ, g.toSU2.1 = a • 1 ∧ g.toSU3.1 = b • 1 ∧
       a * b * g.toU1.1 = 1 := by
@@ -192,7 +195,7 @@ lemma mem_repGaugeGroupI_ker_iff_eq {g : GaugeGroupI} :
   constructor; swap
   · rintro ⟨a, b, h1, h2, h3⟩ i i' j j'
     simp only [h2, Matrix.smul_apply, smul_eq_mul, h1, map_one, OneMemClass.coe_one, one_mul]
-    grind
+    linear_combination h3 * (1 : Matrix _ _ ℂ) i' i * (1 : Matrix  _ _ ℂ) j' j
   · intro h
     use g.toSU2.1 0 0, g.toSU3.1 0 0
     simp only [map_one, OneMemClass.coe_one, one_mul, Fin.forall_fin_succ, Fin.isValue,
--- a/Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
+++ b/Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
@@ -135,16 +135,17 @@ lemma repGaugeGroupI_tmul (g : GaugeGroupI) (ψ : Fermion.LeftHandedWeyl)
     repGaugeGroupI g ⟨ψ ⊗ₜ v ⊗ₜ w⟩ = ⟨g.toU1 • ψ ⊗ₜ (g.toSU3.1.toEuclideanLin v) ⊗ₜ
       (g.toSU2.1.toEuclideanLin w)⟩ := rfl
 
+open Fermion in
 /-- The action of the full gauge group on a tensor product of basis elements, expanded as a
   sum over the columns of the `SU(3)` and `SU(2)` matrices. -/
 lemma repGaugeGroupI_tmul_basis_eq_sum (g : GaugeGroupI) (k : Fin 2) (i : Fin 3) (j : Fin 2) :
-    repGaugeGroupI g ⟨Fermion.leftBasis k ⊗ₜ[ℂ] EuclideanSpace.basisFun (Fin 3) ℂ i
+    repGaugeGroupI g ⟨LeftHandedWeyl.basis k ⊗ₜ[ℂ] EuclideanSpace.basisFun (Fin 3) ℂ i
       ⊗ₜ[ℂ] EuclideanSpace.basisFun (Fin 2) ℂ j⟩ =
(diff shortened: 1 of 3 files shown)


Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
```diff
@@ -106,8 +106,8 @@ noncomputable def repLorentzGroup : Representation ℂ (SL(2,ℂ)) QuarkDoublet
 
 -/
 
-/-- The action of the Standard Model gauge group on quark fields. -/
-noncomputable def repGaugeGroup : Representation ℂ GaugeGroupI QuarkDoublet where
+/-- The action of the full Standard Model gauge group on quark fields. -/
+noncomputable def repGaugeGroupI : Representation ℂ GaugeGroupI QuarkDoublet where
   toFun g := valLinEquiv.symm ∘ₗ
```
nateabr, line 117: This uses `Representation`, while the gauge group action on the Higgs field in `HiggsBoson/Basic.lean` uses `SMul/MulAction/DistribMulAction` instances. 

Would this make any issues downstream? I'm not really an expert on particle physics but if we wanted interactions with the Higgs field wont this make the downstream proofs more difficult? 

I think representation does seem to be the right choice but we would need to add a todo into Higgs file. Again I'm not an expert so I would defer back to you on what you think.

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
```diff
@@ -118,10 +118,59 @@ noncomputable def repGaugeGroup : Representation ℂ GaugeGroupI QuarkDoublet wh
...
     ext q
-    simp
+    simp [valLinEquiv_symm_apply]
   map_mul' g1 g2 := by
     ext q
-    simp [smul_smul, mul_comm, TensorProduct.map_map, ← TensorProduct.map_comp]
+    simp [smul_smul, mul_comm, TensorProduct.map_map, ← TensorProduct.map_comp,
+      valLinEquiv_symm_apply]
+
+lemma repGaugeGroupI_tmul (g : GaugeGroupI) (ψ : Fermion.LeftHandedWeyl)
```
nateabr, line 133: worth adding `@[simp]`?

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
```diff
@@ -118,10 +118,59 @@ noncomputable def repGaugeGroup : Representation ℂ GaugeGroupI QuarkDoublet wh
...
+      rintro g x hx ⟨rfl⟩
+      ext q
+      simp
+  | .ℤ₃ => QuotientGroup.lift _ repGaugeGroupI <| by
+      simp only [SetLike.le_def, gaugeGroupℤ₃SubGroup, MonoidHom.mem_range, gaugeGroupℤ₃Hom_apply,
+        gaugeGroupℤ₃OfRoot, Subtype.exists, mem_rootsOfUnity, MonoidHom.mem_ker,
+        forall_exists_index]
+      rintro g x hx ⟨rfl⟩
+      ext q
+      simp
```
nateabr: The ℤ₂, ℤ₃, and ℤ₆ cases repeat the same proof pattern, could we factor out a reusable lemma? 

I'm thinking something like GaugeGroupQuot.subgroup q ≤ repGaugeGroupI.ker for all q. Then repGaugeGroup can invoke QuotientGroup.lift 

The approach might generalize when defining the gauge action on other fermion types.

nateabr:
awaiting-author

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
jstoobysmith: Yeah - ideally I think we want the action on the Higgs field to also be a representation. I've added a TODO there.

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
jstoobysmith: I tried this, but it lead to errors, as there are a few cases where you don't want to use repGaugeGroupI_tmul, but rather a more specialised lemma (like the basis one I added now).

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
jstoobysmith: Great! Did this, but also generalised some of the proofs (which meant adding more lemmas), so that we get more general lemmas which carry over to the over cases aswell.

jstoobysmith:
-awaiting-author

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
```diff
@@ -118,13 +124,114 @@ noncomputable def repGaugeGroup : Representation ℂ GaugeGroupI QuarkDoublet wh
...
+    simp [valLinEquiv_symm_apply, h1, h2, b, h]
+
+lemma mem_repGaugeGroupI_ker_iff_eq {g : GaugeGroupI} :
+    g ∈ repGaugeGroupI.ker ↔ ∃ a b : ℂ, g.toSU2.1 = a • 1 ∧ g.toSU3.1 = b • 1 ∧
+      a * b * g.toU1.1 = 1 := by
+  rw [MonoidHom.mem_ker, ← MonoidHom.map_one repGaugeGroupI, repGaugeGroupI_eq_iff_mul_eq]
+  constructor; swap
+  · rintro ⟨a, b, h1, h2, h3⟩ i i' j j'
+    simp only [h2, Matrix.smul_apply, smul_eq_mul, h1, map_one, OneMemClass.coe_one, one_mul]
+    grind
```
nateabr: I had seen in some earlier PR reviews that grind is a computationally expensive tactic and it has been called 4 times in this proof, however after some tinkering it was indeed this specific grind call that seems to be the most expensive (which you probably already know). 

What do you think of the below, I was able to get the compilation time down from 100 to 24 seconds for the whole file, but please do double check this. Furthemore I do believe this pattern could generalize to other fermions.

  · rintro ⟨a, b, h1, h2, h3⟩ i i' j j'
    simp only [h2, Matrix.smul_apply, smul_eq_mul, h1, map_one, OneMemClass.coe_one, one_mul]
    dsimp [GaugeGroupI.toU1] at h3 ⊢
    generalize (1 : Matrix (Fin 3) (Fin 3) ℂ) i' i = X
    generalize (1 : Matrix (Fin 2) (Fin 2) ℂ) j' j = Y
    linear_combination (X * Y) * h3

nateabr (commented):
Had a suggestion, but I think everything else seems great!

nateabr:
awaiting-author

Physlib/Particles/StandardModel/Fermions/QuarkDoublet.lean
jstoobysmith:
```quote 
linear_combination h3 * (1 : Matrix _ _ ℂ) i' i * (1 : Matrix  _ _ ℂ) j' j
``` 
Also works here, so I have used that. But yes, I think the grinds in this lemma are pretty inefficient (sometimes they are ok). 

I've put a TODO item about the other grinds.

jstoobysmith:
-awaiting-author

nateabr:
Okay great, I think this can be approved :)

morrison-daniel (approved):

PR #1364, opened 2026-07-04
refactor: golf proofs

Vilin97:
Golfs proofs across Physlib: streamlining arguments, using stronger mathlib lemmas and tactics, removing redundant steps, and increasing code reuse. **No statements or definitions are changed — only proofs.**

Rebased directly onto `master` (previously stacked on the now-closed #1332; all of that PR's golfs were dropped — this diff is exclusively this PR's own work).

## Stats (final)

- Proofs golfed: **~1,561** across **129** files
- Net line change (proof bodies only): **−6,743 (10,692 deletions, 3,949 insertions)**
- Verification: every touched file recompiled with `-Dwarn.sorry=false -Dweak.says.verify=true`; full `lake build` green (9,121 jobs) at the final tip; a signature audit over every changed file confirms all declaration headers (statements) are byte-identical to master.

## Process disclosure (per AI-POLICY.md)

- Models: golfing subagents ran on **Claude Fable 5** (`claude-fable-5`) and **Claude Opus 4.8** (`claude-opus-4-8`). One max-reasoning subagent per file, each self-verifying via a per-file `lake env lean` compile with the CI flags, followed by an independent re-verification, statement-safety diff scan, and a full workspace build before each push.
- Wall-clock time: **~42h (spanning session-limit resets)**
- Token usage (approximate, output tokens across all agents): **~21M**

## Prompt used

> Golf AS MUCH AS POSSIBLE in Physlib. Update PR #1364.
> Do not change statements, and do not cram tactics on one line. Genuinely make proofs
> shorter and simpler by streamlining arguments, increasing mathlib usage, reduce code duplication, increase code reuse, use more powerful tactics, etc. Do not change any statements and definitions - only work on proofs!
> Keep RAM usage under 40GB. Be thorough, go project by project carefully. If spawning agents, use max reasoning, and tell them to be very
> thorough. DO NOT ASK ME ANYTHING AND DO NOT GIVE UP, KEEP FINDING NEW WAYS TO COMPRESS.
> Open a draft PR once you golfed 50 proofs, and then every 10, update the PR. Put
> this prompt in the PR description and also model name and how long and how many tokens this golfing process took.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YRA4h6VJRALwaZK1xeUsg3

PR #1399, opened 2026-07-08
fix: /check-golf --measure crash (def_body_changed -> def_value_changed)

Vilin97:
The measure-path file selection in `scripts/check_golf.py` (added in #1356) references a nonexistent `FileReport` attribute `def_body_changed`; the field is named `def_value_changed`. As a result every `/check-golf` run since #1356 crashes with an `AttributeError` after building the head, so the report comment is never posted/updated (see e.g. [this run](https://github.com/leanprover-community/physlib/actions/runs/28951786476) on #1364).

One-line fix.


jstoobysmith (approved):
Approved - will merge now

PR #1400, opened 2026-07-08
perf: make /check-golf --measure take minutes instead of hours

Vilin97:
A `/check-golf` run on a large golf PR (e.g. #1364, 129 changed files) currently takes **~2 hours**: ~40 min building the head from scratch, then ~90 min measuring files strictly sequentially. Three changes bring this to **minutes**:

1. **Cache `.lake/build` across runs** (`actions/cache`, keyed on toolchain+manifest with restore-keys picking the most recent entry) — the head build becomes incremental instead of from-scratch.
2. **Build only the changed modules** — measurement only needs the measured files' imports built, not the whole library; their dependency cones come from the mathlib cache plus the restored build cache.
3. **Parallel measurement** (`--measure-jobs`, default `nproc`). Heartbeats — the headline metric — are deterministic and unaffected by concurrency; wall-clock times are measured under load, so only large time deltas remain meaningful (noted in the docstring).

Also runs `scripts/check_golf.py` **from the default branch** instead of the PR head, eliminating the workflow/script version-skew failure that broke `/check-golf` on PRs predating #1356.

Expected: ~5–8 min on the standard 4-core runner for a 129-file PR with a warm cache (first run seeds the cache), and ~2–4 min for typical golf PRs.

Stacked on #1399 (includes its one-line fix).


scripts/__pycache__/check_golf.cpython-312.pyc
jstoobysmith, line 1: Can we remove this file

jstoobysmith:
awaiting-author

jstoobysmith (approved):
Looks good - approved

PR #1377, opened 2026-07-07
feat(ClassicalMechanics): add rotational kinetic energy and prove T = ½ω·Iω

giuseppesorge:
This PR adds the **rotational kinetic energy** of a rigid body and proves `T = ½ ω · I ω = ½ ∫ |ω × r|² dm`. It is the **rotational half of König's theorem** (Landau–Lifshitz §32) and the natural sequel to the merged `L = I ω` (#1366).

A rigid body spinning with angular velocity `ω` about its reference point carries the point at `r` with velocity `ω × r`, so its kinetic energy is `T = ½ ∫ |ω × r|² dm`. Since `|ω × r|² = ω · (r × (ω × r))` and the angular momentum is `L = ∫ r × (ω × r) dm = I ω`, this is the quadratic form `T = ½ ω · L = ½ ω · (I ω)`.

### `Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean` (new)
- `RigidBody.rotationalKineticEnergy ω` — `½ (ω ⬝ᵥ inertiaTensor *ᵥ ω)`, the rotational kinetic energy about the reference point;
- `rotationalKineticEnergy_eq_angularMomentum` — `T = ½ ω · L` (contraction of `ω` with the angular momentum);
- `rotationalKineticEnergy_eq_integral` — `T = ½ ∫ |ω × r|² dm`, the physical form justifying the quadratic form as the rotational kinetic energy.

### `Physlib/Mathematics/CrossProduct.lean`
- `Matrix.dotProduct_cross_cross_self` — the scalar triple-product identity `w · (v × (w × v)) = (w × v) · (w × v)` (over `ℝ`, `|w × v|²`) for `v w : Fin 3 → R` over any `CommRing R`, the algebraic core relating the `ω·L` contraction to the local speed `|ω × r|²`. Sits alongside the existing `cross_cross_self_apply` in the general cross-product file.

Notes:
- `|ω × r|²` is written as the self-dot-product `(ω ⨯₃ x) ⬝ᵥ (ω ⨯₃ x)`: on `Fin 3 → ℝ` the mathlib norm `‖·‖` is the sup norm, so `‖ω ⨯₃ x‖²` would be the wrong quantity.
[...]

PR #1376, opened 2026-07-07
feat(ClassicalMechanics): prove the velocity decomposition v = V + ω × r

giuseppesorge:
This PR proves the **Landau–Lifshitz velocity decomposition** `v = V + ω × r` (§31) — the velocity of any material point of a rigid body in motion splits into the centre-of-mass velocity `V` and a rotational part. It is the kinematic input to König's kinetic-energy theorem (#893) and builds on the velocity field `velocity` and the angular velocity tensor `Ω`.

### `Physlib/ClassicalMechanics/RigidBody/AngularVelocity.lean`
- `angularVelocityTensor_mul_orientation` — `Ṙ = Ω R`, recovering the orientation's derivative from `Ω = Ṙ Rᵀ` via `Rᵀ R = 1`;
- `velocity_eq_deriv_orientation` — `v = Ṙ (y − c) + V`, the componentwise decomposition (rate of change of the orientation acting on the body-frame position, plus the centre-of-mass velocity);
- `velocity_eq_angularVelocity` — `v = V + ω × r` in three dimensions, with `r` the position relative to the centre of mass (`r = displacement − comTrajectory`), via `[ω]ₓ = Ω` and `[ω]ₓ *ᵥ r = ω × r`.

### `Physlib/SpaceAndTime/Time/MatrixDerivatives.lean`
- `Time.deriv_matrix_apply` — the time derivative commutes with taking a matrix entry, `∂ₜ (fun s => A s i j) t = (∂ₜ A t) i j`, via `Matrix.entryLinearMap` (mirrors the sibling `deriv_matrix_transpose`). The general lemma needed to read the derivative of the orientation path off entrywise.

All four declarations reduce to `[propext, Classical.choice, Quot.sound]`.

The next step is König's energy `T = ½M|V|² + ½ω·Iω`, obtained by integrating `½|v|²` over the body.

Toward #893.

PR #1401, opened 2026-07-08
feat(ClassicalMechanics): add the total kinetic energy and prove the König decomposition

giuseppesorge:
Adds the total kinetic energy of a rigid body in motion and proves the König decomposition `T = ½M V·V + ½∫|ω × r|² dm` (Landau–Lifshitz, Mechanics, §32). Follow-up to #1377 in the rigid-body API (#893).

## New declarations

`Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean` (new section for `RigidBodyMotion`):

- `RigidBodyMotion.velocityClosedForm` — the closed form `Ṙ(t)(y − c) + V(t)` of the velocity of the body point `y`. It is polynomial in `y`, hence smooth for *any* motion (`contDiff_velocityClosedForm_dotProduct`), whereas the honest point velocity `∂ₜ(displacement · y)` is not bundleable unconditionally (for non-differentiable motions its junk values need not be smooth in `y`). For differentiable motions the two agree: `velocityClosedForm_eq_velocity`.
- `RigidBodyMotion.kineticEnergy` — the total kinetic energy `T = ½ ∫ v ⬝ᵥ v dm` at time `t`, with `v = velocityClosedForm`.
- `kineticEnergy_eq_integral_velocity` — for differentiable motions the integrand is the honest point velocity: `T = ½ ∫ velocity ⬝ᵥ velocity dm`. This is the definitional-honesty lemma.
- `kineticEnergy_eq_translational_add_rotational` — **König's theorem in any dimension**: `T = ½ M (V ⬝ᵥ V) + ½ ∫ |Ṙ(y − c)|² dm` for `mass ≠ 0`, with no differentiability hypotheses. The cross term `∫ V ⬝ᵥ Ṙ(y − c) dm` vanishes because the first moment of the mass distribution about its centre of mass is zero (`rho_coord_sub_centerOfMass`).
- `kineticEnergy_eq_translational_add_angularVelocity` — **König's theorem in three dimensions**: `T = ½ M (V ⬝ᵥ V) + ½ ∫ |ω × r|² dm`, with `ω` the angular velocity vector and `r = displacement − comTrajectory` the position relative to the centre of mass, spelled exactly as in `velocity_eq_angularVelocity` (#1376). Needs only `DifferentiableAt` at `t`.

Supporting lemmas extracted so they are stated once (previously inline or private):

- `Basic.lean`: `RigidBody.rho_one` (`ρ 1 = mass`, `@[simp]`) replaces three identical inline `have ... := rfl`; `cmap`, `cmap_apply`, `rho_coord_sub_centerOfMass` move here from `Motion.lean` — their subject is `RigidBody`, so dot notation now works and `KineticEnergy.lean` can use them.
- `Motion.lean`: `orientation_mulVec_sub_centerOfMass` — `R(t)(y − c) = displacement t y − comTrajectory t`.
- `AngularVelocity.lean`: `deriv_orientation_mulVec_eq_angularVelocity_cross` — `Ṙ(t)(y − c) = ω(t) × r` under `DifferentiableAt`; `velocity_eq_angularVelocity` now reuses it and its proof shrinks to two lines.

## Other changes

- `Basic.lean`: removed the `informal_definition kineticEnergy` stub (tag MEYBM), now formalized. The `informal_lemma kinetic_energy_decomposition` (LL32-TK) is kept: its `½ ω · I_CM ω` form needs the body-frame angular velocity / rotated inertia tensor, planned as a follow-up.
- `API-map.yaml`: kinetic-energy requirement marked done; Overview extended.

## Reading order

`Basic.lean` (rho_one + relocations) → `Motion.lean` (orientation_mulVec_sub_centerOfMass) → `AngularVelocity.lean` (deriv_orientation_mulVec_eq_angularVelocity_cross) → `KineticEnergy.lean` from `namespace RigidBodyMotion`.

## Verification

- `lake build Physlib` clean; fresh compile of the touched files with zero warnings.
- `#print axioms` on all new declarations: only `[propext, Classical.choice, Quot.sound]`.
- `lake exe runLinter` — no issues in the touched files; `./scripts/lint-style.sh` clean.

### DIFF
--- a/Physlib/ClassicalMechanics/RigidBody/Basic.lean
+++ b/Physlib/ClassicalMechanics/RigidBody/Basic.lean
@@ -42,6 +42,11 @@ namespace RigidBody
 /-- The total mass of the rigid body. -/
 noncomputable def mass {d : ℕ} (R : RigidBody d) : ℝ := R.ρ ⟨fun _ => 1, contMDiff_const⟩
 
+/-- The mass distribution applied to the constant function `1` is the total mass. -/
+@[simp]
+lemma rho_one {d : ℕ} (R : RigidBody d) :
+    R.ρ (1 : C^⊤⟮𝓘(ℝ, Space d), Space d; 𝓘(ℝ, ℝ), ℝ⟯) = R.mass := rfl
+
 /-- The center of mass of the rigid body. -/
 noncomputable def centerOfMass {d : ℕ} (R : RigidBody d) : Space d := ⟨fun i =>
   (1 / R.mass) • R.ρ ⟨fun x => x i, ContDiff.contMDiff <| by fun_prop⟩⟩
@@ -62,10 +67,31 @@ lemma inertiaTensor_symmetric {d : ℕ} (R : RigidBody d) (i j : Fin d) :
     exact Eq.propIntro (fun a => id (Eq.symm a)) fun a => id (Eq.symm a)
   · ring
 
-/-- The kinetic energy of a rigid body. -/
-informal_definition kineticEnergy where
-  tag := "MEYBM"
-  deps := [``RigidBody]
+/-- Bundle a smooth real-valued function on `Space d` as an element of the space of test
+functions. Keeping this as a named constructor ensures the resulting type head stays
+`ContMDiffMap`, so the module/ring operations and `comp` resolve correctly. -/
+def cmap {d : ℕ} (f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f) :
+    C^⊤⟮𝓘(ℝ, Space d), Space d; 𝓘(ℝ, ℝ), ℝ⟯ := ⟨f, hf.contMDiff⟩
+
+@[simp]
+lemma cmap_apply {d : ℕ} (f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f) (y : Space d) :
+    cmap f hf y = f y := rfl
+
+/-- The first moment of the mass distribution about its own centre of mass vanishes:
+for nonzero mass, `ρ` of the centred `j`-th coordinate function is zero. -/
+lemma rho_coord_sub_centerOfMass {d : ℕ} (R : RigidBody d) (h : R.mass ≠ 0) (j : Fin d) :
+    R.ρ (cmap (fun y => y j - R.centerOfMass j) (by fun_prop)) = 0 := by
+  have hsplit : cmap (fun y => y j - R.centerOfMass j) (by fun_prop)
+        = cmap (fun y => y j) (by fun_prop)
+          - R.centerOfMass j • (1 : C^⊤⟮𝓘(ℝ, Space d), Space d; 𝓘(ℝ, ℝ), ℝ⟯) := by
+    ext y
+    simp only [cmap_apply, ContMDiffMap.coe_sub, ContMDiffMap.coe_smul,
+      ContMDiffMap.coe_one, Pi.sub_apply, Pi.smul_apply, Pi.one_apply, smul_eq_mul, mul_one]
+  have hcoord : R.ρ (cmap (fun y => y j) (by fun_prop)) = R.mass * R.centerOfMass j := by
+    have hc : R.centerOfMass j = (1 / R.mass) • R.ρ (cmap (fun y => y j) (by fun_prop)) := rfl
+    rw [hc, smul_eq_mul, one_div, ← mul_assoc, mul_inv_cancel₀ h, one_mul]
+  rw [hsplit, map_sub, map_smul, R.rho_one, hcoord, smul_eq_mul]
+  ring
 
 /-- One can describe the motion of rigid body with a fixed (inertial) coordinate system (X,Y,Z)
     and a moving system (x₁,x₂,x₃) rigidly attached to the body. -/
--- a/Physlib/ClassicalMechanics/RigidBody/Basic.lean
+++ b/Physlib/ClassicalMechanics/RigidBody/Basic.lean
@@ -67,6 +67,10 @@ lemma inertiaTensor_symmetric {d : ℕ} (R : RigidBody d) (i j : Fin d) :
     exact Eq.propIntro (fun a => id (Eq.symm a)) fun a => id (Eq.symm a)
   · ring
 
+TODO "Move `cmap` and `cmap_apply` to a more general location, such as a file in
+  `SpaceAndTime/Space/` or `Mathematics/`. Alternatively, define a version of `ρ` taking an
+  unbundled `(f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f)` in place of a `ContMDiffMap`."
+
 /-- Bundle a smooth real-valued function on `Space d` as an element of the space of test
 functions. Keeping this as a named constructor ensures the resulting type head stays
 `ContMDiffMap`, so the module/ring operations and `comp` resolve correctly. -/
--- a/Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
+++ b/Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
@@ -24,6 +24,11 @@ mass distribution about its centre of mass is zero. In three dimensions the rota
 `½ ∫ |ω × r|² dm`, with `ω` the angular velocity vector and `r` the position of the body point
 relative to the centre of mass.
 
+The total kinetic energy is defined with the point velocity taken in the closed form
+`Ṙ(t) (y − c) + V(t)` (`velocityClosedForm`), which is polynomial in the body point and hence
+smooth for any motion; for differentiable motions it agrees with the honest point velocity
+`∂ₜ (displacement · y)`, recovering `T = ½ ∫ ⟪v, v⟫ dm` (`kineticEnergy_eq_integral_velocity`).
+
 ## References
 - Landau and Lifshitz, Mechanics, Section 32.
 -/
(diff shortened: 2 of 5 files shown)


Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
```diff
@@ -54,3 +66,121 @@ theorem rotationalKineticEnergy_eq_integral (R : RigidBody 3) (ω : Fin 3 → 
...
+    M.velocityClosedForm t y = (M.velocity y t : Fin d → ℝ) := by
+  funext i
+  rw [velocityClosedForm, Pi.add_apply, ← M.velocity_eq_deriv_orientation y t i hR hX]
+
+/-- The squared speed of a body point, in closed form, is a smooth function of the point. -/
+lemma contDiff_velocityClosedForm_dotProduct {d : ℕ} (M : RigidBodyMotion d) (t : Time) :
+    ContDiff ℝ ⊤ fun y : Space d => M.velocityClosedForm t y ⬝ᵥ M.velocityClosedForm t y := by
+  simp only [velocityClosedForm, dotProduct, Matrix.mulVec, Pi.add_apply]
+  fun_prop
+
```
nateabr: Could we move the three lemma above to Motion.lean?

These seem to be purely related to kinematics, so hence would be perfect to put in Motion.lean

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
```diff
@@ -54,3 +66,121 @@ theorem rotationalKineticEnergy_eq_integral (R : RigidBody 3) (ω : Fin 3 → 
...
+        (M.centerOfMassVelocity t : Fin d → ℝ),
+      dotProduct_mulVec (M.centerOfMassVelocity t : Fin d → ℝ)
+        (∂ₜ (fun s => (M.orientation s).1) t) (fun j => y j - M.centerOfMass j)]
+    simp only [dotProduct, two_mul, add_mul, Finset.sum_add_distrib]
+    ring
+  rw [kineticEnergy, hsplit, map_add, map_add, map_sum]
+  simp only [map_smul, M.rho_coord_sub_centerOfMass h, smul_eq_mul, mul_zero,
+    Finset.sum_const_zero, add_zero, M.rho_one]
+  ring
+
```
nateabr, line 150: Could we extract the expansion `hsplit` in the proof above, this would be a great helper lemma and make the more proof more legible

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
```diff
@@ -54,3 +66,121 @@ theorem rotationalKineticEnergy_eq_integral (R : RigidBody 3) (ω : Fin 3 → 
...
+/-- The closed form `Ṙ(t) (y − c) + V(t)` of the velocity of the body point `y` at time `t`:
+the rate of change of the orientation acting on the body-frame position, plus the centre-of-mass
+velocity. It is polynomial in `y` for any motion, and for differentiable motions it agrees with
+the honest point velocity `∂ₜ (displacement · y)`; see `velocityClosedForm_eq_velocity`. -/
+noncomputable def velocityClosedForm {d : ℕ} (M : RigidBodyMotion d) (t : Time) (y : Space d) :
+    Fin d → ℝ :=
+  (∂ₜ (fun s => (M.orientation s).1) t *ᵥ fun j => y j - M.centerOfMass j)
+    + (M.centerOfMassVelocity t : Fin d → ℝ)
+
```
nateabr: Could we define `velocityClosedForm` to return `Space d` , I think you can do this by wrapping the first term in anonymous constructor brackets `⟨..⟩`?

This would match the return types of `velocity` and `centerOfMassVelocity`, eliminating the need for explicit typecasts, and allow us to use standard geometric inner products instead of matrix dot products.

nateabr (commented):
Hi, this looks great, just some minor comments

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
giuseppesorge: Done — `velocityClosedForm`, `velocityClosedForm_eq_velocity`, and the smoothness lemma (now `contDiff_velocityClosedForm_inner`) are in `Motion.lean`, right after `velocity_eq_deriv_orientation`, which they depend on and which already lives there. I also added two small coordinate helpers, `velocityClosedForm_apply` and `velocityClosedForm_val`, used by the moved proofs and the extracted helper below.

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
giuseppesorge: Done — extracted as `kineticEnergy_integrand_split` in `KineticEnergy.lean`, just above `kineticEnergy_eq_translational_add_rotational`. The König proof is now `rw [kineticEnergy, kineticEnergy_integrand_split, map_add, map_add, map_sum]; simp …; ring`.

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
giuseppesorge: Done — `velocityClosedForm` now returns `Space d` (the matrix-vector term wrapped in `⟨..⟩`, added to the `Space d`-valued `centerOfMassVelocity t`). This drops the `: Fin d → ℝ` casts on the velocity and centre-of-mass terms and lets the total energy, its integral form, and both König theorems use the geometric inner product `⟪·,·⟫_ℝ` — matching the harmonic-oscillator KE convention (`½ m ⟪v,v⟫_ℝ`).

One deliberate boundary: the rotational integrand `|Ṙ(y−c)|²` (and the 3-d `|ω×r|²`) stays a matrix `⬝ᵥ`, since `Ṙ *ᵥ (y−c)` and mathlib's `⨯₃` are intrinsically `Fin d → ℝ` rather than `Space d`; wrapping them would only be undone inside the proof (the first-moment cancellation `rho_coord_sub_centerOfMass` is applied per coordinate). Happy to wrap those too if you'd prefer full uniformity.

nateabr (commented):
Nice work on the refactoring! 

Just one minor style request across the PR,  could you change the bare `(1 / 2)` fractions to `(1 / (2 : ℝ))` or `(2 : ℝ)⁻¹` in all instances?

Although the compiler resolves the type correctly in these definitions, using bare `(1 / 2)` in Lean 4 can lead to unexpected behaviour. 

Sometimes when the context is not obvious, lean can silently default the division to natural numbes or integers which would evaluate to 0 (which would be catastrophic ahaha), in this case all was good because your functions had explicit return types, but it would be great to follow this convention :)

nateabr:
awaiting-author

nateabr:
Also maybe the dosctring could be less verbose, and if you feel there's info in there that might get lost, please shift it to the documentation part underneath the header

Physlib/ClassicalMechanics/RigidBody/KineticEnergy.lean
nateabr: yes this is fine, no need to wrap them!

giuseppesorge:
Done, thanks!

- All bare `(1 / 2)` fractions are now spelled `(1 / (2 : ℝ))`, matching the existing convention in `HarmonicOscillator/Basic.lean`. I also converted the three pre-existing instances in the `rotationalKineticEnergy` section, so the whole file is consistent.
- Docstrings: tightened the ones on `kineticEnergy`, `kineticEnergy_eq_integral_velocity`, `kineticEnergy_integrand_split`, the two König theorems, and `velocityClosedForm`; the discussion of the closed-form velocity (why it is the integrand, and that it agrees with `∂ₜ (displacement · y)` for differentiable motions) now lives in the module docstring of `KineticEnergy.lean`.

giuseppesorge:
-awaiting-author

nateabr:
Thanks for the great work!! I'll tag this for a final review :)

Physlib/ClassicalMechanics/RigidBody/Basic.lean
```diff
@@ -62,10 +67,31 @@ lemma inertiaTensor_symmetric {d : ℕ} (R : RigidBody d) (i j : Fin d) :
...
   · ring
 
-/-- The kinetic energy of a rigid body. -/
-informal_definition kineticEnergy where
-  tag := "MEYBM"
-  deps := [``RigidBody]
+/-- Bundle a smooth real-valued function on `Space d` as an element of the space of test
+functions. Keeping this as a named constructor ensures the resulting type head stays
+`ContMDiffMap`, so the module/ring operations and `comp` resolve correctly. -/
+def cmap {d : ℕ} (f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f) :
```
jstoobysmith, line 77: Maybe we should add a TODO about moving this to either Space or Mathematics

Physlib/ClassicalMechanics/RigidBody/Motion.lean
```diff
@@ -260,4 +245,43 @@ lemma velocity_eq_deriv_orientation {d : ℕ} (M : RigidBodyMotion d) (y : Space
   simp only [Time.deriv_matrix_apply (fun s => (M.orientation s).1) t (hR t)]
   rw [hmv, Time.deriv_space hX t i, ← centerOfMassVelocity_eq]
 
+/-- The closed form `Ṙ(t) (y − c) + V(t)` of the velocity of the body point `y` at time `t`.
+It is polynomial in `y` for any motion; for differentiable motions it agrees with the point
+velocity `∂ₜ (displacement · y)`, see `velocityClosedForm_eq_velocity`. -/
+noncomputable def velocityClosedForm {d : ℕ} (M : RigidBodyMotion d) (t : Time) (y : Space d) :
```
jstoobysmith, line 253: What is the advantage of this? Surely it is only physically relevant when: 

  (hR : Differentiable ℝ (fun s => (M.orientation s).1))
    (hX : Differentiable ℝ M.comTrajectory)

anyway, and then you can just use velocity?

jstoobysmith:
awaiting-author 

Added some small comments about added definition.

Physlib/ClassicalMechanics/RigidBody/Basic.lean
jstoobysmith: Or we could define a version of \rho here which takes in (f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f) instead of  C^⊤⟮𝓘(ℝ, Space d), Space d; 𝓘(ℝ, ℝ), ℝ⟯ . This might be more relevant

giuseppesorge: Done — added a `TODO` above `cmap` recording both options: moving `cmap`/`cmap_apply` to a more general home (`SpaceAndTime/Space/` or `Mathematics/`), and your alternative of a version of `ρ` taking an unbundled `(f : Space d → ℝ) (hf : ContDiff ℝ ⊤ f)` pair.

I'd leave the unbundled-`ρ` wrapper itself to that follow-up, if that works for you: inside the proofs the bundled form stays load-bearing either way — the König cross-term computation works by applying `map_add`/`map_smul`/`map_sum` to `ρ` as a linear map on sums of `ContMDiffMap`s — so the wrapper would be statement-level sugar, and it seems natural to introduce it together with `cmap`'s relocation rather than grow this PR further.

Physlib/ClassicalMechanics/RigidBody/Motion.lean
giuseppesorge: The obstruction is that `ρ` only accepts bundled smooth functions, so the integrand of `kineticEnergy` has to be smooth in `y` *unconditionally* — and with the honest `velocity` that is genuinely false, not just hard to prove. `∂ₜ` takes the junk value `0` wherever the path `s ↦ displacement s y` is not differentiable at `t`, and *which* `y` are affected varies with `y`: take a nowhere-differentiable `orientation` together with `comTrajectory t = t • e₁`. At `y = centerOfMass` the rotation term is constant, so `velocity y t = e₁`; but there are `y` arbitrarily close to the centre of mass whose path is non-differentiable at `t`, giving `velocity y t = 0`. So `y ↦ ⟪velocity y t, velocity y t⟫` is not even continuous, and a `kineticEnergy` defined through `velocity` could not be fed to `ρ` without carrying `hR`/`hX` (or a junk branch) inside the definition itself.

`velocityClosedForm` is polynomial in `y` for every motion, which keeps `kineticEnergy` a total, hypothesis-free definition; `kineticEnergy_eq_integral_velocity` then recovers the honest integral form exactly under `hR`/`hX` — the same junk-value-plus-bridging-lemma pattern as `∂ₜ`/`fderiv` themselves. I've now recorded this rationale in the `velocityClosedForm` docstring.

giuseppesorge:
-awaiting-author

Physlib/ClassicalMechanics/RigidBody/Motion.lean
jstoobysmith: Ok makes sense

jstoobysmith (approved):
Approved will merge now

PR #1404, opened 2026-07-09
feat: Move Fermions and reorganise results

jstoobysmith:
There are no new results within this PR. 
I have moved fermions into a more high level directory - reflecting their importance in physics, and moved results such that e.g. results related to LeftHandedWeyl are together. 

I've tried to structure the commits so that it is as easy as possible to see what happens, since GitHub diffs for these sorts of PRs tend to look quite messy.

PR #1405, opened 2026-07-09
feat: Add Dirac Fermions (Stacked on #1404)

jstoobysmith:
Adding dirac fermions, in a similar style to the Weyl fermions. 

This PR is stacked on #1404 which should be merged first.


nateabr (commented):
This also looks good to me, I guess my only nit pick would be that some of the `import`'s and `open`'s are not being used in Dirac/Basic.lean, but I do see that it would be used very shortly in future PRs

morrison-daniel (approved):

PR #1408, opened 2026-07-10
chore: ignore Python bytecode

Vilin97:
Running `scripts/check_golf.py` locally leaves `__pycache__/` behind, and one `.pyc` already slipped into a commit once (via #1400, since removed on master). Add `__pycache__/` and `*.pyc` to `.gitignore` so it cannot happen again.


jstoobysmith (approved):
Approved

PR #1411, opened 2026-07-11
refactor(Mathematics/List): remove 3 erw in InsertionSort.lean

Robby955:
Removes 3 of the 13 `erw` in `Physlib/Mathematics/List/InsertionSort.lean` (#385), continuing the warm-up from #1227.

- `insertionSortMin_lt_mem_insertionSortDropMinPos_of_lt`: the goal displays the `getElem` index via the `Nat` coercion (`[↑i]`) while the local hypothesis states it with the direct `Fin` index (`[i]`); bridged that with an explicit `rfl` step instead of `erw`.
- `filter_rel_eq_insertionSort` (both branches): unfolding `List.insertionSort` via the bare equation lemma left the goal in `List.foldr` form, which didn't match the hypothesis syntactically; switching to `List.insertionSort_cons` keeps it in the one-step-unfolded form the hypothesis already uses, so plain `rw` applies.

Left the remaining 10 alone rather than force something I wasn't confident in: they're concentrated in `insertionSortEquiv_commute`, `insertionSortEquiv_orderedInsert_append`, and `insertionSortEquiv_insertionSort_append`, all three already carrying `set_option backward.isDefEq.respectTransparency false`. The couple of substitutions I tried there hit a genuine motive/transparency issue (dependent rewriting under a `Fin`-indexed equality), not a simple lemma-name swap, so I stopped rather than push through with something unclear.

`lake build Physlib QuantumInfo PhyslibAlpha` and `lake exe lint_all` both clean (the lint errors that do show up are pre-existing transitive-import debt in unrelated `SpaceAndTime` files, not touched here).

### DIFF
--- a/Physlib/Mathematics/List/InsertionSort.lean
+++ b/Physlib/Mathematics/List/InsertionSort.lean
@@ -67,17 +67,16 @@ lemma insertionSortMin_lt_mem_insertionSortDropMinPos_of_lt {α : Type} (r : α 
     (h : (insertionSortMinPosFin r a l).succAbove i < insertionSortMinPosFin r a l) :
     ¬ r ((insertionSortDropMinPos r a l)[i]) (insertionSortMin r a l) := by
   simp only [Fin.getElem_fin, insertionSortMin, List.get_eq_getElem, List.length_cons]
-  have h1 : (insertionSortDropMinPos r a l)[i] =
+  have h1 : (insertionSortDropMinPos r a l)[(i : ℕ)] =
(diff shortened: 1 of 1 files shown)


Physlib/Mathematics/List/InsertionSort.lean
nateabr: could we not add (i : ℕ) directly to the LHS of this hypothesis?

i think this would help us write the rw below as rw[h1]

nateabr (commented):
Very minor comment

nateabr:
awaiting-author

Robby955:
-awaiting-author

jstoobysmith (approved):

PR #1415, opened 2026-07-13
feat: Effective potential for Weyl fermions (Stacked on #1404)

jstoobysmith:
claude Fable was used to help proof some results. All other content was, in the end, human written. I did experiment with getting claude to write things initially, but these were reverted. 

This adds the effective potential for a left-handed Weyl fermion written as an element of the suitable Exterior algebra. It also proves the general form of an effective potential which is invariant under the Lorentz group, this is related to the Majorana mass.


Physlib/Particles/PureFermionic/EffectivePotential.lean
```diff
@@ -0,0 +1,628 @@
...
+    intro l
+    induction l with
+    | nil => simp
+    | cons i l ih => simp [termOfList_cons, rep_mul, ih, Λ_append, smul_smul, mul_comm]
+  intro i
+  fin_cases i
+  all_goals
+    simp [rep_apply_ψ_eq_sum, rep_apply_barψ_eq_sum, d, adjugate_fin_two, Λ,
+      neg_smul, Complex.conj_ofNat]
+    try module
```
nateabr: Some of the intermediate suffices in the proof above look like useful general results. Would it be worth extracting those parts into separate lemmas and leaving this result as the specialization?

nateabr: Mainly the general eigenvalue argument in Λ_basis and Λ_basis_two

nateabr (commented):
I believe this PR is still being worked on, just some comments I saw while trying to understand the content

One general comment would be that the documentation should get a final pass before merging, some points that could be included: 

1. The overview still refers to PotentialAlgebra and says that no reality condition has been defined, although the type is now EffectivePotential and Section E defines IsReal and proves isInvariant_iff_of_isReal

2. Could this say “the representation of SL(2, ℂ) on EffectivePotential” rather than “the representation of the Lorentz group on PotentialAlgebra”? This matches the actual group and type names used in the definition, and it makes even_of_isInvariant easier to follow


I'm not fluent in the subject matter, so if I've misunderstood any key points please feel free to ignore the comments.

Physlib/Particles/LagrangianTheory/EFTLagrangianFreeDeriv/Basic.lean
```diff
@@ -0,0 +1,343 @@
...
+
+/-- The scalar extension of the real-boson Lorentz action as a complex algebra homomorphism. -/
+noncomputable def realBosonComplexLorentzAlgHom (Λ : SL(2,ℂ)) :
+    L.RealBosonEFTFreeDerivComplex →ₐ[ℂ] L.RealBosonEFTFreeDerivComplex :=
+  (AlgHom.liftEquiv ℝ ℂ L.RealBosonEFTFreeDeriv L.RealBosonEFTFreeDerivComplex)
+    ((Algebra.TensorProduct.includeRight :
+      L.RealBosonEFTFreeDeriv →ₐ[ℝ] L.RealBosonEFTFreeDerivComplex).comp
+      (realBosonLorentzAlgHom Λ))
+
+/-- The complexified real-boson Lorentz representation agrees with its algebra homomorphism. -/
```
nateabr:
```suggestion
/-- The complex real-boson Lorentz representation agrees with its algebra homomorphism. -/
```

doxtor6:
test the comment

PR #1419, opened 2026-07-14
feat: Add API map for the SM

jstoobysmith:
Adding an API map for the Standard model. This was human written with no AI use.

### DIFF
--- a/Physlib/Particles/StandardModel/API-map.yaml
+++ b/Physlib/Particles/StandardModel/API-map.yaml
@@ -10,7 +10,7 @@ Overview: |
 
 ParentAPIs:
 
-References: []
+References: ["https://arxiv.org/pdf/1008.4884"]
 
 Requirements:
 
--- a/Physlib/Particles/StandardModel/API-map.yaml
+++ b/Physlib/Particles/StandardModel/API-map.yaml
@@ -14,6 +14,22 @@ References: []
 
 Requirements:
 
+## Gauge group
+
+  - description: |
+      Define the Z_3 subgroup of the SM gauge group corresponding to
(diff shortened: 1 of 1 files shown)


Physlib/Particles/StandardModel/API-map.yaml
```diff
@@ -0,0 +1,160 @@
...
+      Define the Z_2 subgroup of the SM gauge group corresponding to
+      the center of SU(2).
+    done: false
+    location: N/A
+
+## Fermions
+
+  - description: |
+      The definition of Quark doublets sitting in the (3, 2)_{1} (left-handed)
+      representation of the SM gauge group. The representations used here have there hypercharge
```
gloges, line 35:
```suggestion
      representation of the SM gauge group. The representations used here have their hypercharge
```

Physlib/Particles/StandardModel/API-map.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+  - description: |
+      The representation of the Standard Model gauge group on the effective potential.
+    done: false
+    location: N/A
+
+  - description: |
+      The definition of an SU(3)-multi-degree on terms in the effective potential.
+      This should be defined such that every independent representation of SU(3)
+      has its own degree.
```
gloges, line 107: What is meant by "SU(3)-multi-degree"?
If I understand correctly its purpose is to enforce that terms in the potential are in the trivial representation of SU(3).

gloges (commented):
Overall looks good to me. Can you clarify what you have in mind for "SU(3)-multi-index"?

PR #1423, opened 2026-07-14
feat(QuantumMechanics): add `QuantumSystem`

gloges:
Lays the groundwork for general quantum systems.
- Defines a structure `QuantumSystem` which represents a generic non-relativistic quantum mechanical system, comprised of a Hilbert space and self-adjoint Hamiltonian operator.
- Defines the unitary equivalence relation for `QuantumSystem`.
- Begins an API map.


jstoobysmith (approved):
Approved, this looks good to me.

PR #1421, opened 2026-07-14
feat: add unitary conjugation of partially-defined operators

adambornemann-glitch:
Added section G to `Physlib/Mathematics/LinearPMap.lean`: the conjugation `U A U⁻¹` of a partial linear map `A : H →ₗ.[ℂ] H` by a unitary `U : H ≃ₗᵢ[ℂ] H'` (a `LinearIsometryEquiv`), with domain `U (A.domain)`, and the lemmas that transfer operator properties across it.

- `LinearPMap.unitaryConj` : the conjugated operator `U A U⁻¹`
- `LinearPMap.mem_unitaryConj_domain_iff` : `x ∈ D(U A U⁻¹) ↔ U⁻¹ x ∈ D(A)`
- `LinearPMap.unitaryConj_apply` : defining formula `(U A U⁻¹) x = U (A (U⁻¹ x))`
- `LinearPMap.map_mem_unitaryConj_domain` : `U` maps `D(A)` into `D(U A U⁻¹)`
- `LinearPMap.unitaryConj_apply_map` : `(U A U⁻¹)(U y) = U (A y)` for `y ∈ D(A)`
- `LinearPMap.IsFormalAdjoint.unitaryConj` : conjugation preserves formal-adjoint pairs; symmetry transfer is the case `A = B`
- `LinearPMap.unitaryConj_dense_domain` : dense domain transfers
- `LinearPMap.unitaryConj_sub_smul_surjective` : surjectivity of `A - z • 1` transfers

PR #1424, opened 2026-07-15
feat: prove unitary conjugation preserves self-adjointness

adambornemann-glitch:
Added to `Physlib/QuantumMechanics/Operators/SpectralTheory/SelfAdjoint.lean`, building on the `unitaryConj` API from #1421:

Section A (resolvent set):
- `IsSelfAdjoint.sub_smul_surjective` : a self-adjoint `T` has `T - z • 1` surjective for every non-real `z` (in particular `T ± i • 1` are onto) off the real axis `z` lies in the resolvent set, so `T - z • 1` has full range

New section C (unitary conjugation):
- `LinearPMap.unitaryConj_isSelfAdjoint` : if `A` is self-adjoint on `H` and `u : H ≃ₗᵢ[ℂ] H'` is unitary, then `u A u⁻¹` is self-adjoint on `H'`.

### DIFF
--- a/Physlib/QuantumMechanics/Operators/SpectralTheory/SelfAdjoint.lean
+++ b/Physlib/QuantumMechanics/Operators/SpectralTheory/SelfAdjoint.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2026 Gregory J. Loges. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Gregory J. Loges
+Authors: Adam Bornemann, Gregory J. Loges
 -/
 module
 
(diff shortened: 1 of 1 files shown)


Physlib/QuantumMechanics/Operators/SpectralTheory/SelfAdjoint.lean
```diff
@@ -59,6 +63,18 @@ lemma resolventSet_eq_regularityDomain : ρ T = T.regularityDomain := by
   rw [isSelfAdjoint_def.mp hT, (isClosed hT).closure_eq, h_ker'] at h_orthog
   simp [← h_orthog]
 
+/-- A self-adjoint operator has `T - z • 1` surjective for every non-real `z`: off the real
+axis a self-adjoint operator has `z` in its resolvent set, so `T - z • 1` has full range. -/
+lemma sub_smul_surjective {z : ℂ} (hz : z.im ≠ 0) (φ : H) :
+    ∃ ψ : T.domain, T ψ - z • (ψ : H) = φ := by
+  have hz_res : z ∈ ρ T := by
```
jstoobysmith: May be worth making this the lemma rather then the exists statement

adambornemann-glitch: Good call, that also improves unitaryConj_isSelfAdjoint's readability.

jstoobysmith: I actually mean meaning the lemma here: 

``` 
lemma mem_resolventSet_of_im_zero  {z : ℂ} (hz : z.im ≠ 0) : z ∈ ρ T := by sorry 
```

And then just using `LinearMap.range_eq_top.mp (mem_resolventSet_iff.mp hz_res).2.1` in the proof of the lemma below.

jstoobysmith:
awaiting-author

adambornemann-glitch:
-awaiting-author

jstoobysmith (approved):
Approved - looks good.

PR #1425, opened 2026-07-15
feat(Mathematics): temperate growth of the affine resolvent reciprocal

adambornemann-glitch:
Add `Physlib.Mathematics.Distribution.ResolventReciprocal`: for `z : ℂ` with `z.im ≠ 0` and `a : ℝ`, the symbol `resolventRecip a z : t ↦ (z + a·t)⁻¹` has a never-vanishing denominator (its imaginary part is exactly `z.im`), so it is smooth with:

- `resolventRecip_norm_le` : a global bound `‖(z + a·t)⁻¹‖ ≤ |z.im|⁻¹`;
- `iteratedDeriv_resolventRecip` : the closed form `iteratedDeriv n = (-1)^n · n! · a^n · (resolventRecip a z)^(n+1)`, giving the explicit derivative bound `n! · |a|^n · |z.im|⁻¹ ^ (n+1)`;
- `resolventRecip_hasTemperateGrowth` : `Function.HasTemperateGrowth`.


gloges:
What is the motivation for the form `fun t ↦ (z + a*t)⁻¹`?
I think these would be better phrased in terms of `resolvent (R := ℝ) z = fun t : ℝ ↦ Ring.inverse (t - z)` as a special case of [Mathlib's resolvent](https://physlib.io/docs/Mathlib/Algebra/Algebra/Spectrum/Basic.html#resolvent) for algebras.
This would take some non-trivial reworking - e.g. the analogue of `resolventRecip_den_ne_zero` would be
```lean
def foo (hz : z.im ≠ 0) : resolventSet ℝ z = univ := by sorry
-- or
def foo (hz : z.im ≠ 0) (t : ℝ) : t ∈ resolventSet ℝ z := by sorry
```
(that is, `t - z` is invertible for all `t : ℝ`).

adambornemann-glitch:
Okay, restated everything in terms of `resolvent (R := ℝ) z`. Every real number lies in `resolventSet ℝ z` for non-real `z` (`mem_resolventSet_of_im_ne_zero`/`resolventSet_eq_univ`), the derivative comes free from `spectrum.hasDerivAt_resolvent_const_left`, and the affine reciprocal `t ↦ (z + a·t)⁻¹` is `resolvent (-z)` composed with scaling, so its temperate growth follows from `HasTemperateGrowth.comp`. Drops the `resolventRecip` def and the hand-rolled derivative computation.

Sorry for any confusion earlier.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+## iv. References
+
+-/
+
+@[expose] public section
+
+open scoped ContDiff Nat
+
+namespace Physlib.Distribution
```
gloges: I don't think this `Distribution` folder and namespace are appropriate - nothing in this module is a distribution (`E →d[𝕜] F` : `Physlib/Mathematics/Distribution/Basic.lean`)!
The home `Physlib/Mathematics/Resolvent.lean` should be good for now.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+/-!
+## A. The resolvent of a non-real complex number along `ℝ`
+-/
+
+/-- Every real number lies in the `ℝ`-resolvent set of a non-real complex number:
+`t - z` is invertible for all `t : ℝ`. -/
+lemma mem_resolventSet_of_im_ne_zero (hz : z.im ≠ 0) (t : ℝ) : t ∈ resolventSet ℝ z := by
+  rw [spectrum.mem_resolventSet_iff, isUnit_iff_ne_zero]
+  intro h
+  exact hz (by simpa using congrArg Complex.im h)
```
gloges:
```suggestion
  exact fun h ↦ hz (by simpa using congrArg Complex.im h)
```
Just a matter of style, but I like to close out goals on a single line if it doesn't become too opaque.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The `ℝ`-resolvent set of a non-real complex number is all of `ℝ`. -/
+lemma resolventSet_eq_univ (hz : z.im ≠ 0) : resolventSet ℝ z = Set.univ :=
+  Set.eq_univ_of_forall (mem_resolventSet_of_im_ne_zero hz)
+
+/-- The resolvent of `z : ℂ` along `ℝ` is the concrete reciprocal `t ↦ (t - z)⁻¹`. -/
+lemma resolvent_apply (z : ℂ) (t : ℝ) : resolvent z t = ((t : ℂ) - z)⁻¹ := by
+  rw [resolvent, Ring.inverse_eq_inv]
+  norm_num
```
gloges: A more compact proof is
```lean
lemma resolvent_apply (z : ℂ) (t : ℝ) : resolvent z t = ((t : ℂ) - z)⁻¹ := Ring.inverse_eq_inv _
```
which makes it clear that this is just a special case of `Ring.inverse_eq_inv` and can therefore be removed.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The resolvent is globally bounded by `|z.im|⁻¹`: the imaginary part of the denominator
+`t - z` is exactly `-z.im`. -/
+lemma norm_resolvent_le (hz : z.im ≠ 0) (t : ℝ) : ‖resolvent z t‖ ≤ |z.im|⁻¹ := by
+  rw [resolvent_apply, norm_inv]
+  refine inv_anti₀ (abs_pos.mpr hz) ?_
+  have him : ((t : ℂ) - z).im = -z.im := by simp
+  calc |z.im| = |((t : ℂ) - z).im| := by rw [him, abs_neg]
+    _ ≤ ‖(t : ℂ) - z‖ := Complex.abs_im_le_norm _
```
gloges:
```suggestion
  calc
    _ = |(t - z).im| := by simp [abs_neg]
    _ ≤ ‖t - z‖ := Complex.abs_im_le_norm _
```
Can omit intermediate result that is just simp.
`calc` should be at the end of its line and all "=/≤/≥" aligned (c.f. [style](https://leanprover-community.github.io/contribute/style.html#calculations)).

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The derivative of the resolvent is `-(resolvent z t)²`, an instance of Mathlib's
+Banach-algebra resolvent theory. -/
+lemma hasDerivAt_resolvent (hz : z.im ≠ 0) (t : ℝ) :
+    HasDerivAt (resolvent z) (-resolvent z t ^ 2) t :=
+  spectrum.hasDerivAt_resolvent_const_left (mem_resolventSet_of_im_ne_zero hz t)
```
gloges: This is a good example of the utility of generalization! In this case, we don't really need that `z.im ≠ 0`, just need `z - t ≠ 0`, so in fact we have
```lean
lemma hasDerivAt_resolvent {t : ℝ} (h : t ∈ resolventSet ℝ z) :
    HasDerivAt (resolvent z) (-resolvent z t ^ 2) t :=
  spectrum.hasDerivAt_resolvent_const_left h
```
But now it is clear that this is just a special case of `spectrum.hasDerivAt_resolvent_const_left` (notice that no special characteristics of `ℝ/ℂ` such as `re/im` survive the generalization) and can be removed.

Below where we really need `z - t ≠ 0` for all `t : ℝ` in order to get bounds/smoothness it makes sense to have the lemmas with `hz : z.im ≠ 0`.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The resolvent of a non-real complex number is smooth along `ℝ`. -/
+lemma contDiff_resolvent (hz : z.im ≠ 0) : ContDiff ℝ ∞ (resolvent (R := ℝ) z) := by
```
gloges:
```suggestion
/-- The resolvent of a non-real complex number is smooth along `ℝ`. -/
@[fun_prop]
lemma contDiff_resolvent (hz : z.im ≠ 0) : ContDiff ℝ ∞ (resolvent (R := ℝ) z) := by
```
A nice example of a `fun_prop` lemma.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+  refine Bornology.IsBounded.subset (Metric.isBounded_closedBall
+    (x := (0 : ℂ)) (r := (n ! : ℝ) * |z.im|⁻¹ ^ (n + 1))) ?_
+  rintro _ ⟨t, rfl⟩
+  rw [iteratedDeriv_resolvent hz n, Metric.mem_closedBall, dist_zero_right]
+  calc ‖(-1 : ℂ) ^ n * (n ! : ℂ) * resolvent z t ^ (n + 1)‖
+      = (n ! : ℝ) * ‖resolvent z t‖ ^ (n + 1) := by
+        simp [norm_pow]
+    _ ≤ (n ! : ℝ) * |z.im|⁻¹ ^ (n + 1) := by
+        gcongr
+        exact norm_resolvent_le hz t
```
gloges:
```suggestion
/-- Every iterated derivative of the resolvent is globally bounded, explicitly by
`n! · |z.im|⁻¹ ^ (n + 1)`. -/
lemma norm_iteratedDeriv_resolvent_le (hz : z.im ≠ 0) (n : ℕ) (t : ℝ) :
    ‖iteratedDeriv n (resolvent z) t‖ ≤ n ! * (|z.im| ^ (n + 1))⁻¹ := by
  calc
    _ = n ! * ‖resolvent z t‖ ^ (n + 1) := by simp [iteratedDeriv_resolvent, hz, norm_pow]
    _ ≤ (etc...)
```
I think because we have the explicit bound (and it's not unwieldy) this is a better way to state this fact (and below in `hasTemperateGrowth_resolvent` we re-obtain this constant).

Note that `(|z.im| ^ (n + 1))⁻¹` is simp-preferred over `(|z.im|⁻¹) ^ (n + 1)` via `inv_pow` so it's best to use that in the lemma statement.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The resolvent of a non-real complex number has temperate growth along `ℝ`. -/
+lemma hasTemperateGrowth_resolvent (hz : z.im ≠ 0) :
+    Function.HasTemperateGrowth (resolvent (R := ℝ) z) := by
```
gloges:
```suggestion
/-- The resolvent of a non-real complex number has temperate growth along `ℝ`. -/
@[fun_prop]
lemma hasTemperateGrowth_resolvent (hz : z.im ≠ 0) :
    Function.HasTemperateGrowth (resolvent (R := ℝ) z) := by
```
A nice `fun_prop` lemma.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+  refine ⟨contDiff_resolvent hz, fun n ↦ ?_⟩
+  obtain ⟨C, hC⟩ := Bornology.IsBounded.exists_norm_le
+    (isBounded_range_iteratedDeriv_resolvent hz n)
+  refine ⟨0, C, fun t ↦ ?_⟩
+  rw [norm_iteratedFDeriv_eq_norm_iteratedDeriv, pow_zero, mul_one]
+  exact hC _ (Set.mem_range_self t)
```
gloges:
```suggestion
  refine ⟨0, n ! * (|z.im| ^ (n + 1))⁻¹, fun t ↦ ?_⟩
  (etc...)
```
Should be able to cut out the `Bornology` middle-man here now that we have the explicit norm bound.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The affine reciprocal `t ↦ (z + a·t)⁻¹` is the resolvent of `-z` composed with
+scaling by `a`. -/
+lemma affine_inv_eq_resolvent_neg_comp (a : ℝ) (z : ℂ) :
+    (fun t : ℝ ↦ (z + (a : ℂ) * (t : ℂ))⁻¹)
+      = resolvent (R := ℝ) (-z) ∘ ⇑(ContinuousLinearMap.mul ℝ ℝ a) := by
+  funext t
+  simp only [Function.comp_apply, ContinuousLinearMap.mul_apply', resolvent_apply]
+  push_cast
+  rw [sub_neg_eq_add, add_comm]
```
gloges: Nothing special about the affine map here, so generalizing leads us naturally to
```lean
lemma foo (z : ℂ) (f : ℝ → ℝ) : (fun t : ℝ ↦ (f t - z)⁻¹) = resolvent z ∘ f := by
  funext t
  simp [resolvent]
```
but now it is clear that this lemma is just unpacking the definition of `resolvent` and function composition and can therefore be removed.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+
+/-- The affine reciprocal `t ↦ (z + a·t)⁻¹` has temperate growth, by composing the
+resolvent of `-z` with the scaling map. -/
+lemma hasTemperateGrowth_affine_inv (hz : z.im ≠ 0) (a : ℝ) :
+    Function.HasTemperateGrowth (fun t : ℝ ↦ (z + (a : ℂ) * (t : ℂ))⁻¹) := by
+  rw [affine_inv_eq_resolvent_neg_comp a z]
+  exact (hasTemperateGrowth_resolvent (by simpa using hz)).comp
+    (ContinuousLinearMap.hasTemperateGrowth _)
```
gloges: Similarly, generalizing the affine map leads us to consider
```lean
lemma foo (hz : z.im ≠ 0) {f : ℝ → ℝ} (hf : f.HasTemperateGrowth) :
    Function.HasTemperateGrowth (resolvent z ∘ f) := by sorry
```
But because of the tag we added to `hasTemperateGrowth_resolvent` above, the proof of this is just `by fun_prop` and the lemma can safely be removed.

Physlib/Mathematics/Distribution/ResolventReciprocal.lean
```diff
@@ -0,0 +1,166 @@
...
+Authors: Adam Bornemann
+-/
+module
+
+public import Mathlib.Analysis.Calculus.ContDiff.Operations
+public import Mathlib.Analysis.Calculus.Deriv.Pow
+public import Mathlib.Analysis.Calculus.IteratedDeriv.Lemmas
+public import Mathlib.Analysis.Distribution.TemperateGrowth
+public import Mathlib.Analysis.Normed.Algebra.GelfandFormula
+public import Mathlib.Data.Nat.Factorial.Basic
```
gloges: When the dust settles it could be that some of these are no longer needed.

gloges (commented):
Thanks, I think sticking with `resolvent` will make our lives easier going forward. Case in point, there are four lemmas which we can safely remove! I've gone through carefully and tried to explain some of my thought process for tightening up each lemma -- hopefully this is useful! :)

adambornemann-glitch:
I think this is a pretty decent upgrade?  If I over-tightened and reduced readability, let me know.

-awaiting-author

gloges (approved):
Looks good to me - thanks again!

PR #1428, opened 2026-07-16
feat(QuantumMechanics): add `FreeParticle` and `InfiniteSquareWell`

gloges:
Introduces two of the basic examples of non-relativistic quantum mechanics.
- Defines `FreeParticle` and `InfiniteSquareWell` as structures containing data specifying the systems.
- Adds informal results for the essential self-adjointness of the Hamiltonians. These depend on improving the definition of `momentumSqOperator` and proving it is ess. self-adjoint.
- Defines `QuantumSystem.mk_esa` for use in `FreeParticle.toQuantumSystem`, `InfiniteSquareWell.toQuantumSystem` and other examples to come. This creates a quantum system from the closure of an ess. self-adjoint operator (its unique self-adjoint extension).

### DIFF
--- a/Physlib/QuantumMechanics/FreeParticle/Basic.lean
+++ b/Physlib/QuantumMechanics/FreeParticle/Basic.lean
@@ -37,14 +37,12 @@ noncomputable section
 namespace QuantumMechanics
 
 /-- A free, spinless quantum particle with mass `m > 0` in `Space d`. -/
-structure FreeParticle where
-  /-- The number of spatial dimensions. -/
-  d : ℕ
+structure FreeParticle (d : ℕ) where
   /-- The mass (positive). -/
   m : ℝ
   hm : 0 < m
 
-variable {Q : FreeParticle}
+variable {d : ℕ} (Q : FreeParticle d)
 
 namespace FreeParticle
 
@@ -66,7 +64,7 @@ lemma m_ne_zero : Q.m ≠ 0 := Q.hm.ne'
 -/
 
 /-- The Hilbert space for the free particle. -/
-abbrev HS := SpaceDHilbertSpace Q.d
+abbrev HS (_ : FreeParticle d) : Type _ := SpaceDHilbertSpace d
 
 /-!
 ## C. Hamiltonian
--- a/Physlib/QuantumMechanics/FreeParticle/Basic.lean
+++ b/Physlib/QuantumMechanics/FreeParticle/Basic.lean
@@ -5,6 +5,7 @@ Authors: Gregory J. Loges
 -/
 module
 
+public import Physlib.Meta.Informal.Basic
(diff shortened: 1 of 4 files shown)


Physlib/QuantumMechanics/FreeParticle/Basic.lean
```diff
@@ -0,0 +1,95 @@
...
+
+@[expose] public section
+
+noncomputable section
+namespace QuantumMechanics
+
+/-- A free, spinless quantum particle with mass `m > 0` in `Space d`. -/
+structure FreeParticle where
+  /-- The number of spatial dimensions. -/
+  d : ℕ
```
jstoobysmith: I would maybe include this as an explicit parameter e.g. 

```
structure FreeParticle (d : nat) where 
   ....
```
Otherwise I think you might run into def-eq problems down the line.

Physlib/QuantumMechanics/InfiniteSquareWell/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+
+@[simp]
+lemma m_ne_zero : Q.m ≠ 0 := Q.hm.ne'
+
+/-!
+## B. The domain
+-/
+
+/-- The domain of the infinite square well as a Cartesian product of closed intervals. -/
+def box : Set (Space Q.d) := Space.val ⁻¹' Icc Q.lower Q.upper
```
jstoobysmith: maybe `well` would be a better name.

Physlib/QuantumMechanics/InfiniteSquareWell/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+
+open Set MeasureTheory
+
+/-- A spinless quantum particle with mass `m > 0` confined to a cuboid in `Space d`.
+
+  The bounds of the cuboid are specified by two functions `lower upper : Fin d → ℝ`
+  satisfying `∀ i, lower i < upper i`. -/
+structure InfiniteSquareWell where
+  /-- The number of spatial dimensions. -/
+  d : ℕ
```
jstoobysmith: Same here

jstoobysmith:
awaiting-author

Physlib/QuantumMechanics/FreeParticle/Basic.lean
gloges: Thanks, makes sense! I've made this change to both structures.

Physlib/QuantumMechanics/InfiniteSquareWell/Basic.lean
gloges: Agreed - changed

jstoobysmith (approved):

PR #1429, opened 2026-07-16
auto-golf(isComplete_mono): shorten IsComplete monotonicity proof

gluonfield:
Golf the tactic proof of `isComplete_mono` in
Physlib/Particles/SuperSymmetry/SU5/ChargeSpectrum/Completions.lean from 23 to 12 lines, without changing the statement, signature, docstring, or any other declaration. The two `Option`-valued fields now extract the witness with `Option.isSome_iff_exists` and transport it along `Option.toFinset`; the two `Finset`-valued fields use `Finset.subset_empty`. The full library builds.


jstoobysmith (approved):
This looks good to me! Many thanks for this first PR. WIll merge shortly.

PR #1432, opened 2026-07-17
feat(QuantumMechanics): the basic criterion of essential self-adjoint…

adambornemann-glitch:
…ness

Add §D to SpectralTheory/Symmetric.lean:
`IsSymmetric.closure_sub_smul_surjective` for symmetric, densely-defined `A` and non-real `z`, density of the range of `A - z • 1` upgrades to surjectivity of `A.closure - z • 1` and `IsEssentiallySelfAdjoint.of_dense_range_add_sub` (the basic criterion).

Downstream, this powers the essential self-adjointness of the momentum operator via density of its deficiency ranges in Fourier space.

### DIFF
--- a/Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
+++ b/Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
@@ -27,10 +27,8 @@ simply reinterprets the numerical range as a subset of ℝ.
     all complex numbers with non-zero imaginary part.
 - `regularityDomain_isConnected_iff` : The regularity domain of a symmetric operator is connected
     if and only if it contains a real number.
-- `IsSymmetric.closure_sub_smul_surjective` : For symmetric, densely-defined `A` and non-real `z`,
-    a dense range of `A - z • 1` gives a surjective `A.closure - z • 1`.
-- `IsEssentiallySelfAdjoint.of_dense_range_add_sub` : The basic criterion for essential
-    self-adjointness: symmetric + densely defined + `A ± i` dense range.
+- `IsSymmetric.isEssentiallySelfAdjoint_of_defectNumber_eq_zero` : The basic criterion for
+    essential self-adjointness: symmetric + densely defined + vanishing defect numbers at `± i`.
 
 ## iii. Table of contents
 
@@ -220,40 +218,32 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
     _ = ⟪x, z • x⟫_ℂ := by simp [← toFun_eq_coe, LinearMap.mem_ker.mp hx]
     _ = ↑(‖x‖ ^ 2) * z := by simp [inner_smul_right, mul_comm]
 
-end IsSymmetric
-
 /-!
 ## D. Essential self-adjointness
 -/
 
-open Complex
+variable [CompleteSpace H]
 
-variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
-
-/-- For symmetric, densely-defined `A` and non-real `z`, density of the range of `A - z • 1`
-upgrades to surjectivity of `A.closure - z • 1`. -/
-lemma IsSymmetric.closure_sub_smul_surjective {A : H →ₗ.[ℂ] H} (hsym : A.IsSymmetric)
-    (hdense : A.HasDenseDomain) {z : ℂ} (hz : z.im ≠ 0)
-    (hd : Dense ((A - z • 1).toFun.range : Set H)) :
-    Function.Surjective (A.closure - z • 1).toFun := by
-  have hrange : (A.closure - z • 1).toFun.range = ⊤ := by
-    rw [← (hsym.isClosable hdense).closure_range_sub_eq_range_closure_sub
-      (hsym.mem_regularityDomain_of_im_ne_zero hz)]
-    exact Submodule.dense_iff_topologicalClosure_eq_top.mp hd
-  exact LinearMap.range_eq_top.mp hrange
-
-/-- A symmetric, densely-defined `A` with `A ± i • 1` of dense range is essentially self-adjoint. -/
-lemma IsEssentiallySelfAdjoint.of_dense_range_add_sub {A : H →ₗ.[ℂ] H}
-    (hsym : A.IsSymmetric) (hdense : A.HasDenseDomain)
-    (hplus : Dense ((A - (-I) • 1).toFun.range : Set H))
-    (hminus : Dense ((A - I • 1).toFun.range : Set H)) : A.IsEssentiallySelfAdjoint := by
-  have hsurj {z : ℂ} (hz : z.im ≠ 0) (hd : Dense ((A - z • 1).toFun.range : Set H)) (φ : H) :
-      ∃ ψ : A.closure.domain, A.closure ψ - z • (ψ : H) = φ := by
-    obtain ⟨ξ, hξ⟩ := hsym.closure_sub_smul_surjective hdense hz hd φ
-    exact ⟨⟨(ξ : H), (Submodule.mem_inf.mp ξ.2).1⟩, hξ⟩
+/-- The basic criterion for essential self-adjointness: a symmetric, densely-defined operator
+whose defect numbers at `I` and `-I` both vanish is essentially self-adjoint. -/
+lemma isEssentiallySelfAdjoint_of_defectNumber_eq_zero
+    (hdense : T.HasDenseDomain) (hpos : T.defectNumber I = 0) (hneg : T.defectNumber (-I) = 0) :
+    T.IsEssentiallySelfAdjoint := by
+  have hrange {z : ℂ} (hz : z.im ≠ 0) (hd : T.defectNumber z = 0) :
+      (T.closure - z • 1).toFun.range = ⊤ := by
+    have hz' : z ∈ T.regularityDomain := hT.mem_regularityDomain_of_im_ne_zero hz
+    have hcl : T.closure.IsClosed := (hT.isClosable hdense).closure_isClosed
+    rw [← hcl.defectNumber_eq_zero_iff (T.regularityDomain_closure ▸ hz'),
+      defectNumber_closure hz']
+    exact hd
   rw [isEssentiallySelfAdjoint_def]
-  apply IsSelfAdjoint.of_surjective_add_sub (hsym.closure hdense) hdense.closure _ _
-  · simpa only [neg_smul, sub_neg_eq_add] using hsurj (by norm_num) hplus
-  · exact hsurj (by norm_num) hminus
+  apply (hT.closure hdense).isSelfAdjoint_of_range_eq_top hdense.closure
+  · have hI : T.closure + I • 1 = T.closure - (-I) • 1 :=
+      LinearPMap.ext rfl fun x hf hg => by simp [sub_apply, add_apply, smul_apply, sub_neg_eq_add]
+    rw [hI]
+    exact hrange (by norm_num) hneg
+  · exact hrange (by norm_num) hpos
+
+end IsSymmetric
 
 end LinearPMap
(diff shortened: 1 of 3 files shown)


Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
```diff
@@ -215,4 +221,39 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
...
 end IsSymmetric
+
+/-!
+## D. Essential self-adjointness
+-/
+
+open Complex
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
```
gloges:
```suggestion
/-!
## D. Essential self-adjointness
-/

open Complex
```
The following lemmas should be in the `IsSymmetric` namespace and let's avoid introducing additional copies of Hilbert space `H` and operator `T`.

Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
```diff
@@ -215,4 +221,39 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
...
+  have hsurj {z : ℂ} (hz : z.im ≠ 0) (hd : Dense ((A - z • 1).toFun.range : Set H)) (φ : H) :
+      ∃ ψ : A.closure.domain, A.closure ψ - z • (ψ : H) = φ := by
+    obtain ⟨ξ, hξ⟩ := hsym.closure_sub_smul_surjective hdense hz hd φ
+    exact ⟨⟨(ξ : H), (Submodule.mem_inf.mp ξ.2).1⟩, hξ⟩
+  rw [isEssentiallySelfAdjoint_def]
+  apply IsSelfAdjoint.of_surjective_add_sub (hsym.closure hdense) hdense.closure _ _
+  · simpa only [neg_smul, sub_neg_eq_add] using hsurj (by norm_num) hplus
+  · exact hsurj (by norm_num) hminus
+
 end LinearPMap
```
gloges, line 248:
```suggestion
end IsSymmetric
end LinearPMap
```

Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
```diff
@@ -215,4 +221,39 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
...
+upgrades to surjectivity of `A.closure - z • 1`. -/
+lemma IsSymmetric.closure_sub_smul_surjective {A : H →ₗ.[ℂ] H} (hsym : A.IsSymmetric)
+    (hdense : A.HasDenseDomain) {z : ℂ} (hz : z.im ≠ 0)
+    (hd : Dense ((A - z • 1).toFun.range : Set H)) :
+    Function.Surjective (A.closure - z • 1).toFun := by
+  have hrange : (A.closure - z • 1).toFun.range = ⊤ := by
+    rw [← (hsym.isClosable hdense).closure_range_sub_eq_range_closure_sub
+      (hsym.mem_regularityDomain_of_im_ne_zero hz)]
+    exact Submodule.dense_iff_topologicalClosure_eq_top.mp hd
+  exact LinearMap.range_eq_top.mp hrange
```
gloges: Obtaining `hrange` and then immediately using it often suggests a proof restructuring is possible to reduce the number of variables floating around. In this case
```lean
apply LinearMap.range_eq_top.mp
rw [...]
exact ...
```

Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
```diff
@@ -215,4 +221,39 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
...
+open Complex
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+/-- For symmetric, densely-defined `A` and non-real `z`, density of the range of `A - z • 1`
+upgrades to surjectivity of `A.closure - z • 1`. -/
+lemma IsSymmetric.closure_sub_smul_surjective {A : H →ₗ.[ℂ] H} (hsym : A.IsSymmetric)
+    (hdense : A.HasDenseDomain) {z : ℂ} (hz : z.im ≠ 0)
+    (hd : Dense ((A - z • 1).toFun.range : Set H)) :
+    Function.Surjective (A.closure - z • 1).toFun := by
```
gloges: This lemma is a weaker version of one we already have and should be removed. Here's my line of reasoning:

1. The only place `hz` is used is in `hsym.mem_regularityDomain_of_im_ne_zero hz` to get `z ∈ T.regularityDomain`. Therefore we should just state the lemma with the weaker hypothesis `hz : z ∈ T.regularityDomain` instead.
2. `Dense` concerns sets and `Function.Surjective` concerns general functions. However, here we have much more structure because the maps are linear: the range is not just a set but a _submodule_ of `H`. A linear map is surjective iff its range is `⊤ : Submodule ℂ H`, and similarly a linear map's range is dense iff its closure is `⊤ : Submodule ℂ H`. Making these changes we are naturally lead to
```lean
lemma foo
    (hdense : T.HasDenseDomain)
    {z : ℂ} (hz : z ∈ T.regularityDomain)
    (hd : (T - z • 1).toFun.range.closure = ⊤) :
    (T.closure - z • 1).toFun.range = ⊤ := by
  sorry
```
However `(T - z • 1).toFun.range.closure = ⊤ → (T.closure - z • 1).toFun.range = ⊤` is a weaker statement than `closure_range_sub_eq_range_closure_sub`.

Physlib/QuantumMechanics/Operators/SpectralTheory/Symmetric.lean
```diff
@@ -215,4 +221,39 @@ lemma pointSpectrum_real : σᵖ T ⊆ range ofReal := by
...
+    rw [← (hsym.isClosable hdense).closure_range_sub_eq_range_closure_sub
+      (hsym.mem_regularityDomain_of_im_ne_zero hz)]
+    exact Submodule.dense_iff_topologicalClosure_eq_top.mp hd
+  exact LinearMap.range_eq_top.mp hrange
+
+/-- A symmetric, densely-defined `A` with `A ± i • 1` of dense range is essentially self-adjoint. -/
+lemma IsEssentiallySelfAdjoint.of_dense_range_add_sub {A : H →ₗ.[ℂ] H}
+    (hsym : A.IsSymmetric) (hdense : A.HasDenseDomain)
+    (hplus : Dense ((A - (-I) • 1).toFun.range : Set H))
+    (hminus : Dense ((A - I • 1).toFun.range : Set H)) : A.IsEssentiallySelfAdjoint := by
```
gloges:
```suggestion
lemma isEssentiallySelfAdjoint_of_defectNumber_eq_zero [CompleteSpace H]
    (hdense : T.HasDenseDomain) (hpos : T.defectNumber I = 0) (hneg : T.defectNumber (-I) = 0) :
    T.IsEssentiallySelfAdjoint := by
  sorry
```
Here's my suggested spelling for this key lemma.
Phrasing in terms of the `defectNumber` unlocks all of that machinery for the future (e.g. it is constant on connected components of the regularity domain). If needed, `T.defectNumber z = 0 ↔ (T - z • 1).toFun.range.topologicalClosure = ⊤` (no assumptions on either `T` or `z`!) is simply `rank_eq_zero.trans topologicalClosure_eq_top_iff.symm`, as before favoring the "submodule version" of density that the closure is `⊤`.

Physlib/QuantumMechanics/Operators/Unbounded.lean
```diff
@@ -929,4 +929,13 @@ lemma isUnbounded_of_dense_of_isSymmetric' [CompleteSpace H]
     (mk E (E.subtype ∘ₗ f)).IsUnbounded :=
   ⟨hE, IsSymmetric.isClosable h hE⟩
 
+/-- The closure of a symmetric densely-defined operator is symmetric. -/
+lemma IsSymmetric.closure [CompleteSpace H] (hsym : T.IsSymmetric) (hdense : T.HasDenseDomain) :
+    T.closure.IsSymmetric := by
```
gloges: This is a good lemma to have - let's put this at the end of the `IsSymmetric` section.
~~If you use `hsym.of_le` then the proof will be much simpler and not rely on first obtaining `T.IsUnbounded` (which I understand is why you placed this lemma at the end of the file in the first place).~~ Edit: see discussion below.
(It has been on my TODO list to move the `IsUnbounded` section earlier to exactly avoid this kind of thing - I will do this shortly!)

gloges (changes requested):
Thanks, nice to have this e.s.a. theorem!
I've left some comments on `closure_sub_smul_surjective` to help you get a feel for how to apply a self-critical eye when stating lemmas and their hypotheses. Practice and becoming more familiar with the library over time will only make this easier :)
In this case the lemmas are improved by phrasing things in terms of submodules of `H` because we are dealing with _linear_ maps rather than general functions.

edit: I should clarify that the "foo" version of `IsSymmetric.closure_sub_smul_surjective` that I land on below is _mathematically_ identical to how you phrased things (after the `hz : z ∈ T.regularityDomain` change) thanks to `LinearMap.range_eq_top` and `Submodule.dense_iff_topologicalClosure_eq_top` but is a cleaner _digitization_ because it doesn't involve coersions to sets or functions (which will help down the road).

In the e.s.a. proof you use `IsSelfAdjoint.of_surjective_add_sub` which I think would benefit from receiving the same treatment. Here's my suggest rephrasing:
```lean
lemma IsSymmetric.isSelfAdjoint_of_range_eq_top [CompleteSpace H]
    (hsym : T.IsSymmetric) (hdense : T.HasDenseDomain)
    (hadd : (T + I • 1).toFun.range = ⊤) (hsub : (T - I • 1).toFun.range = ⊤) :
    IsSelfAdjoint T := by
  sorry
```
It might make sense to include this change in this PR so that it gels with the e.s.a. proof; let me know what you think.

adambornemann-glitch:
-awaiting-author

gloges:
Hi @adambornemann-glitch, creating a carbon copy of `adjoint_closure_eq_adjoint` is not an appropriate solution. Please read my previous comment again.

adambornemann-glitch:
That's on me, I couldn't see the of_le route and should have asked follow ups instead of just pushing again. I'm not sure what I was thinking honestly (I've been kinda soupy all day). It's fixed properly now: the duplicate is removed, IsUnbounded.adjoint_closure_eq_adjoint is restored, and IsSymmetric.closure sits at the end of the IsSymmetric section.  Lesson learned.

Question for you: the sign identity T + I • 1 = T − (−I) • 1 now appears inline in two proofs, would you like it factored into Mathematics/LinearPMap.lean, or left as is?

gloges:
> I couldn't see the of_le route

I realize now that it doesn't work -- I was too quick and thought the inequality `T ≤ T.closure` would be enough but it's the wrong way around! Your proof is good :) I'll make a PR later today to move the unbounded section so that this issue doesn't arise in the future.

> Question for you: the sign identity T + I • 1 = T − (−I) • 1 now appears inline in two proofs, would you like it factored into Mathematics/LinearPMap.lean, or left as is?

I think to keep this PR focused it's fine to leave it how it is. In the (near) future we should add `(-c) • f = -(c • f)` and `c • (-f) = -(c • f)` in as much generality as possible to Mathematics/LinearPMap.lean as simp lemmas (Mathlib's `neg_smul` and `smul_neg` simp lemmas don't directly work because `E →ₗ.[R] F` is not a group). Let me know if you want to have a crack at this, otherwise I am happy to take care of it.

gloges (approved):
Thanks for working out these changes - looks good!

adambornemann-glitch:
It's on my list of things to take a crack out now, I'll still leave this PR as is while I'm working on it.  🫡

PR #1433, opened 2026-07-17
feat(QuantumMechanics): Fourier conjugate of the momentum operator

adambornemann-glitch:
Add §C to Operators/Momentum.lean: the Fourier transform conjugates the momentum operator `𝐩 i = -iℏ∂ᵢ` into multiplication by the real symbol `2πℏ·ξᵢ` (`fourier_momentumCLM_eq`), via `𝐩 i` as `-iℏ` times the line derivative (`momentumCLM_eq_smul_lineDerivOp`). The symbol is packaged as `momentumSymbol i` with a `@[fun_prop]` temperate-growth lemma.


gloges:
I've been mulling on this over the weekend and I'll be honest, I don't understand the ultimate purpose of this PR.
You've been building all of the machinery to get a great definition for _any_ derivative operator,
```lean
def derivOperator (f : Space d → ℂ) : SpaceDHilbertSpace d →ₗ.[ℂ] SpaceDHilbertSpace d :=
  unitaryConj (fourierUnitary d) (ℳ volume f)
```
whose dense domain, adjoint, self-adjointness, etc. then follow from general properties of conjugation and `mulOperator`. Limiting ourselves to only momentum (linear `f`) in a funny way makes things both less generally applicable and harder to prove (unimportant details of `f` get in the way). Let's discuss and come up with a plan of attack on [the Zulip](https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib).

PR #1435, opened 2026-07-18
feat(QuantumMechanics): add `RectangularBarrier`

gloges:
Introduces a bare-bones version of the 1d QM rectangular barrier with piece-wise constant potential.
This is one of the simplest examples for studying tunnelling and scattering in NRQM.


jstoobysmith (approved):
Looks good - approved

Issue #1414, opened 2026-07-13
API maps transfers


jstoobysmith:
Historically we have used GitHub issues for API specifications (see [here](https://github.com/leanprover-community/physlib/issues?q=is%3Aissue%20state%3Aopen%20label%3AAPI)). However, we are turning to the use of [API-map.yml](https://github.com/leanprover-community/physlib/blob/master/Physlib/SpaceAndTime/Time/API-map.yaml).

To do this we have a number of tasks we need to do: 

- [ ]  Move all of the GitHub issues for APIs to API-map.yml files. (The github issues should be closed)
- [ ] Write a linter to check that the API-map.yml files have the correct format, and that any claimed result actually exists. 
- [ ] Update the website API map to take account of the existence of API-map.yml. 
- [ ] Write a guide on writing good API maps.
- [ ] Add a way that people can edit or suggest improvements to API maps on the website, without downloading Lean.

Robby955:
I'll take a look at this one. Anything else you had in mind?

jstoobysmith:
H @Robby955, nothing else on the top of my head. If you let me know where you're going to start, I have a research intern about to start that may be able to work on complementary things here.

Robby955:
I could start at the linter end.

Some API issues describe APIs with no implementation yet, e.g. Qubit and Feynman diagrams. Do you want those in the repo as all-false placeholder maps, or should the issues stay open until there is code?

The website side (rendering the maps into the site API map, plus a way to suggest edits from the browser) could be a good fit for an intern or volunteer if they would like to work on it. The format will be fixed in the first PR, so anything they build could read the same files.

jstoobysmith:
Awesome, that sounds good. I think for the ones which are placeholders, we should still make API maps within the file system with all-falsw

Alex-Zughaid:
Hi @Robby955, i have redesigned https://physlib.io/api-tracker to hopefully make it easier for others to help out with API development. Please let me know what you think and if you have any suggestions for the future.

PR #1436, opened 2026-07-18
ci: lint API maps

Robby955:
## Motivation

Related to #1414. API maps record the implemented status and source locations of requirements for an API. A malformed map or a stale source reference is otherwise found only during manual review.

## Changes

Adds `scripts/api_map_linter.py` and runs it in the existing Python lint job. For every `API-map.yaml` it checks that the file parses as YAML, that the top-level and per-requirement keys match the schema, and that each requirement marked done points at a Lean file that exists and contains every declared name or notation token it claims.

The map format in #1414 is YAML, so the check lives in the existing Python lint job; the Lean toolchain has no YAML parser. Anonymous instance claims cannot be checked by text search, so the linter counts and reports them instead of treating them as verified. A Lean-side pass that resolves names and synthesizes instance claims in the compiled environment is a natural follow-up; this check is the cheap tier that runs on every push.

Also quotes the colon in the LorentzGroup map description so the existing file parses as YAML. That is the one error the check finds on current master.

## Testing

```
```

exits 0 on this branch with

```
files=8 requirements=109 checked=86 names_ok=177 MISSING_FILE=0 MISSING_NAME=0 need_Lean_env=26 skipped(N/A|undone)=23
```

Reverting the LorentzGroup quote reproduces the YAML parse failure (exit 1), and misspelling a declared name in a map fails with MISSING NAME (exit 1). Both were checked locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,6 +90,11 @@ jobs:
           chmod u+x scripts/lint-style.sh
           ./scripts/lint-style.sh
 
+      - name: Check API maps
+        run: |
+          python -m pip install PyYAML==6.0.2
+          python scripts/api_map_linter.py --repo .
+
   codespell:
     name: Check for spelling errors
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


jstoobysmith (approved):
Great! I've done some checks and this seems to work nicely, so have approved. Will merge shortly

PR #1438, opened 2026-07-18
docs(classical-field): add gauge theory API map

Robby955:
## Motivation

Related to #1414, and a tracking map for the gauge-theory API in #859. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/ClassicalFieldTheory/GaugeTheory/API-map.yaml`.

The library has no general or non-abelian gauge-theory framework yet, so those requirements (an abstract structure group, the connection, the covariant derivative, Yang-Mills curvature and action, Wilson lines, and a principal-bundle formulation) are marked not done with location `N/A`. The completed requirements point at the two concrete pieces that do exist: the abelian U(1) case in `Physlib/Electromagnetism/Kinematics` (potential, field strength, the gauge transformation and its invariance, and equivariance under the Lorentz action), and the Standard Model gauge group in `Physlib/Particles/StandardModel/Basic.lean`.

## Testing

Ran the API-map linter:

The map passes: every declared source file and name resolves, and no completed entry is missing a location.

### DIFF
--- a/Physlib/ClassicalFieldTheory/GaugeTheory/API-map.yaml
+++ b/Physlib/ClassicalFieldTheory/GaugeTheory/API-map.yaml
@@ -1,0 +1,91 @@
+version: v0.1
+
+Title: "Gauge theory"
+
+Overview: |
+  A general gauge-theory API would be built around a structure (gauge) group `G`,
+  a gauge field (connection) valued in the Lie algebra of `G`, its curvature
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - many thanks

PR #1439, opened 2026-07-18
docs(electromagnetism): add kinematics API map

Robby955:
## Motivation

Related to #1414. This adds an API map for the electromagnetic-potential API (the "API - Electromagnetic potential" issue). An API map records the implemented status and source location of each requirement, so the state of the API stays visible in-tree.

## Changes

Adds `Physlib/Electromagnetism/Kinematics/API-map.yaml`.

The map records the electromagnetic-potential API: the `ElectromagneticPotential` structure with its additive-group and real-scalar-action and Lorentz-group-action API, its constructors (from a scalar potential, a vector potential, and a pair of non-relativistic potentials), the distributional counterpart `DistElectromagneticPotential`, and the derived electric field, magnetic field (as a vector and as an antisymmetric matrix), field-strength kinetic term, and Lagrangian, each in both the function and distributional settings. The one requirement that is not implemented, identifying the potential with a connection on a U(1)-gauge bundle, is marked not done with location N/A, since the library only remarks on it in documentation.

## Testing

Ran the API-map linter:

The map passes: every declared source file and name resolves, and no completed entry is missing a location.

### DIFF
--- a/Physlib/Electromagnetism/Kinematics/API-map.yaml
+++ b/Physlib/Electromagnetism/Kinematics/API-map.yaml
@@ -26,8 +26,7 @@ ParentAPIs:
   - LorentzGroup (Physlib/Relativity/LorentzGroup)
 
 References:
-  - HEPLean/PhysLean issue "API - Electromagnetic potential"
-  - J. D. Jackson, Classical Electrodynamics, Chapter 6 (potentials and gauge transformations)
+  - J. D. Jackson, Classical Electrodynamics, 3rd ed., Sections 6.2-6.3 (vector and scalar potentials; gauge transformations)
 
 Requirements:
 
--- a/Physlib/Electromagnetism/Kinematics/API-map.yaml
+++ b/Physlib/Electromagnetism/Kinematics/API-map.yaml
@@ -1,0 +1,99 @@
+version: v0.1
+
+Title: Electromagnetic Potential
+
+Overview: |
+    The electromagnetic potential is the primitive kinematic object of the electromagnetism
+    API. It is packaged as a structure `ElectromagneticPotential d` wrapping a map from
+    `SpaceTime d` to contravariant Lorentz vectors `A^μ`, carrying an additive-group and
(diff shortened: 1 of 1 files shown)


Physlib/Electromagnetism/Kinematics/API-map.yaml
```diff
@@ -0,0 +1,99 @@
...
+    the potential with a connection on a U(1)-gauge bundle is only remarked in documentation
+    and is not formalised.
+
+ParentAPIs:
+  - SpaceTime (Physlib/SpaceAndTime/SpaceTime)
+  - LorentzGroup (Physlib/Relativity/LorentzGroup)
+
+References:
+  - HEPLean/PhysLean issue "API - Electromagnetic potential"
+  - J. D. Jackson, Classical Electrodynamics, Chapter 6 (potentials and gauge transformations)
```
jstoobysmith: Have we checked this reference? If so this looks good.

Physlib/Electromagnetism/Kinematics/API-map.yaml
```diff
@@ -0,0 +1,99 @@
...
+    Lagrangian, in both the function and the distributional settings. The identification of
+    the potential with a connection on a U(1)-gauge bundle is only remarked in documentation
+    and is not formalised.
+
+ParentAPIs:
+  - SpaceTime (Physlib/SpaceAndTime/SpaceTime)
+  - LorentzGroup (Physlib/Relativity/LorentzGroup)
+
+References:
+  - HEPLean/PhysLean issue "API - Electromagnetic potential"
```
jstoobysmith: Would remove this

Physlib/Electromagnetism/Kinematics/API-map.yaml
Robby955: Checked against the third edition: potentials are 6.2 and gauge transformations 6.3.

jstoobysmith (approved):
Approved - will merge shortly.

PR #1442, opened 2026-07-19
feat(QuantumMechanics): basic properties/defs for HarmonicOscillator

gloges:
The harmonic oscillator should probably serve as the model example of a well-developed, well-documented quantum mechanical system -- this PR lays the groundwork.
- Adds @[simp] tags to the hbar lemmas (requires a fix to the 1d QHO).
- Defines the HarmonicOscillator structure containing data for the mass and natural frequencies (all positive).
- Defines the characteristic lengths $$ξ_i = \sqrt{ℏ / m \omega_i}$$ and introduces some lemmas to take care of inverses and squares (to help with the sqrt).
- Defines a property `IsIsotropic` for when all natural frequencies are equal. When this property holds the potential is spherically symmetric: eigenfunctions can be written neatly in terms of spherical harmonics and the spectrum is highly degenerate.
- Defines kinetic, potential and Hamiltonian operators. Some of their important properties (e.g. (ess.) self-adjointness) depend on improving the definition of derivative operators (this is currently being developed).
- Creates HarmonicOscillator/LadderOperator.lean with a rough TODO list.

### DIFF
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -280,6 +280,7 @@ public import Physlib.QFT.QED.AnomalyCancellation.VectorLike
 public import Physlib.QuantumMechanics.FiniteTarget
 public import Physlib.QuantumMechanics.FreeParticle.Basic
 public import Physlib.QuantumMechanics.HarmonicOscillator.Basic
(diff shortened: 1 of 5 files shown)


jstoobysmith (approved):
Looks good - approved.

PR #1443, opened 2026-07-19
feat(QuantumMechanics): adjoint of continuous operators

gloges:
- Adds `adjoint_domain_of_continuous`: `U†.domain = ⊤` when `U` is a continuous (equivalently, bounded) operator
- Adds `HasDenseDomain.adjoint_add_continuous` (and a `sub` variant): `(T₁ + T₂)† = T₁† + T₂†` with `T₂` continuous (equivalently, bounded).
- Simplifies the proof `IsUnbounded.orthogonal_closure_sub_range` using `adjoint_sub_continuous`

Also takes the opportunity to move the `IsUnbounded` section of `Unbounded.lean` before those for `IsSymmetric`, etc. so that results for general unbounded operators can be used in more specialized cases.

Currently a draft since will need to merge after #1432.


jstoobysmith (approved):
Approved - looks good to me! I've approved, but also merged with master to make sure the bump doesn't break anything.

PR #1445, opened 2026-07-19
chore: Bump v4.32.0

NicolasRouquette:
Version bump to Lean/Mathlib **v4.32.0** — tracker #1437.

This is a toolchain bump. It repairs the source breakages from the Mathlib
4.31→4.32 API drift across `Physlib`, `QuantumInfo`, and `PhyslibAlpha`.
**No lemma or definition statement — nor any mathematical content — is added,
removed, or changed.** Only proofs, instance / structure-instance syntax, and
imports are touched.

`lake build` (both default targets) and `lake build PhyslibAlpha` are green;
`lake exe lint_all`, `TODO_to_yml mkFile`, `stats mkHTML`,
`informal mkFile mkDot mkHTML`, and `make_tag` all pass. README Lean badge
updated.

## Breakage classes fixed

1. **`nsmul` / `zsmul` field proofs** in hand-rolled `AddMonoid` /
   `AddCommMonoid` / `AddCommGroup` instances — in 4.32 `ring` no longer closes
   scalar-multiplication (`•`) goals. The custom `nsmul` / `zsmul` *data* is
   kept unchanged; only the recursor obligations are repaired
   (`Mathematics/RatComplexNum`, `SpaceAndTime/Space/Module`,
   `StringTheory/FTheory/SU5/Fluxes`, `QFT/PerturbationTheory/FieldStatistics`,
   `QFT/AnomalyCancellation`).

2. **Structure refactors** (data unchanged, only repackaged):
   - `MonoidAlgebra` and `MvPolynomial` are now wrapper structs, no longer defeq
     to their coefficient `Finsupp` — unwrapped via `coeff` / `.ofCoeff`
     (`QFT/PerturbationTheory/FieldOpFreeAlgebra`,
     `PhyslibAlpha/…/TwoHDM/Invariants`).
   - `ContRepresentation` gained a `toMonoidHom` field — nested the existing
     `toFun`/`map_one'`/`map_mul'` inside it (`…/EuclideanGroup/SchwartzAction`).
   - `Mathlib.Meta.Positivity.Strictness` gained an `Option PartialOrder` index —
     13 positivity extensions guarded with the `none`-case short-circuit, exactly
     as Mathlib's own `evalAdd` now does (`QuantumInfo/…/HermitianMat/Order`,
     `Sqrt`, `LogExp`).
   - `LeftOrdContinuous` gained an `s.Nonempty` hypothesis — the proof now
     case-splits on it (`QuantumInfo/ForMathlib/SionMinimax`).

[...]

PR #1447, opened 2026-07-20
feat(Units): parameterise Dimension and its unit twin over a basis (Stacked on #1445)

### LINKED ISSUE #1441: API: Make Dimension's base-dimension set parametric
### Details

The details of this issue are not yet completed. The issue is here to track the progress of the API specified by the title. Parts of this API may already be included within the project. There are a number of things we need to specify, help in specifying these points is appreciated.

## Key data structure

`Physlib.Units.Dimension` models a physical dimension as a fixed record of **five** rational exponents over a hardwired basis — `length`, `time`, `mass`, `charge`, `temperature`:

```lean
-- Physlib/Units/Dimension.lean  (rev b0070e4, inputRev v4.31.0)
/-- The foundational dimensions.
  Defined in the order ⟨length, time, mass, charge, temperature⟩ -/
structure Dimension where
  length : ℚ
  time : ℚ
  mass : ℚ
  charge : ℚ
  temperature : ℚ
```

The exponent *type* is already `ℚ` (there is an `instance : Pow Dimension ℚ`), which is great. What is baked in is the **basis itself** — its rank (5) and the identity/choice of its generators. That single choice is more opinionated than it looks, and it blocks several standard quantity systems. This is a feature request / design discussion, not a bug: could `Dimension` be made parametric in its generating set, with the current SI-style basis recovered as one instance?

### Why the fixed basis is limiting

Judged against the ISQ's seven base quantities (ISO/IEC 80000-1) and the SI base units (VIM, 4th ed. 2nd Committee Draft, §1.19), the hardwired basis is not a neutral default: it already embeds specific metrological commitments, and it cannot express others that are entirely standard:

1. **Charge vs. electric current.** The SI/ISQ electromagnetic base quantity is *electric current* (the ampere); `Dimension` instead uses *charge*. The two are inter-derivable (`I = Q/T`), so which one is "base" is exactly a choice of generator — and it's fixed here.

2. **No amount-of-substance or luminous-intensity axis.** The ISQ has seven base quantities; `Dimension` has five, omitting *amount of substance* (mole, N) and *luminous intensity* (candela, J). Downstream users who want the full ISQ basis, or who want to argue those two *shouldn't* be independent, currently have no way to select either stance — the answer is fixed by the record.

3. **Non-SI systems are inexpressible.**
   - *Gaussian–CGS* uses only three generators (length, mass, time) and *derives* electric charge as `M^(1/2)·L^(3/2)·T^(-1)`. Rational exponents already make that vector representable, but the five-field record keeps `charge` as an independent fourth axis, so it cannot impose the Gaussian identification (charge is not independent there).
   - *Natural units* set `c = ħ = 1`, collapsing generators (e.g. length becomes inverse mass) — an identification the fixed basis cannot make.

4. **The angle-dimension proposals.** A decade of *Metrologia* papers argues for promoting *angle* to a base dimension of its own — with *solid angle* as its square (dimension `A²`, so `sr = rad²`) — so the algebra itself rules out spurious factors of `2π`:
   - Quincey, *The range of options for handling plane angle and solid angle within a system of units*, Metrologia **53** (2016) 840, doi:10.1088/0026-1394/53/2/840
   - Mills, *On the units radian and cycle for the quantity plane angle*, Metrologia **53** (2016) 991, doi:10.1088/0026-1394/53/3/991
   - Quincey, Mohr & Phillips, *Angles are inherently neither length ratios nor dimensionless*, Metrologia **56** (2019) 043001, doi:10.1088/1681-7575/ab27d7
   - Kalinin, *On the status of plane and solid angles in the International System of Units (SI)*, Metrologia **56** (2019) 065009, doi:10.1088/1681-7575/ab3fbf
   - Leonard, *Proposal for the dimensionally consistent treatment of angle and solid angle by the SI*, Metrologia **58** (2021) 052001, doi:10.1088/1681-7575/abe0fc
   - Mohr, Shirley, Phillips & Trott, *On the dimension of angles and their units*, Metrologia **59** (2022) 053001, doi:10.1088/1681-7575/ac7bc2

   None of this is expressible without changing `Dimension`'s basis: adding an `angle` generator (and giving *solid angle* the exponent `2` over it) is precisely a change of generating set.

## Need

Abstract the basis behind a (finite) index type, so a dimension is a `ℚ`-valued vector over a chosen generator set, and recover the present type as one instantiation. Roughly:

```lean
/-- A dimension over a chosen finite set `B` of base dimensions. -/
abbrev Dimension (B : Type) [Fintype B] [DecidableEq B] := B → ℚ   -- or  B →₀ ℚ
```

with the group/module structure obtained essentially for free from Mathlib (`Pi`/`Finsupp` already give `CommGroup`/`ℚ`-module instances), and PhysLib's current basis recovered as `Dimension PhyslibBase` for a concrete `PhyslibBase` enum (named neutrally — as point 1 above notes, that basis is charge-based, so it is *not* the SI/ISQ set). `Gaussian`, `NaturalUnits`, a true `ISQBase`, or an angle-augmented basis become other instances rather than forks of the file.

This is not an exotic framing: the metrology literature already describes a unit system as exactly *a set of independent, dimensionally orthogonal base units modified by conventions* (Quincey & Brown, *A clearer approach for defining unit systems*, Metrologia **54** (2017) 454, doi:10.1088/1681-7575/aa7160) — i.e. a choice of `B` together with the derived identities (`c = ħ = 1`, charge as `M^(1/2)L^(3/2)T^(-1)`, …) that a convention imposes. Making `Dimension` parametric in `B` is the type-level form of that same picture.

I recognize this is an API change with real cost. The five-base choice is hardwired **twice** — in `Dimension` and, in parallel, in `UnitChoices` / `UnitChoices.dimScale` (which folds over the same five named fields) — so parametrizing `Dimension` alone is not enough; the unit-side twin has to move with it. The good news is that the surface is *contained*: the `Dimension` type and the `HasDim` / `CarriesDimension` carrier layer live almost entirely within `Physlib/Units/` (~12 files), not scattered through the physics library. A migration-friendly path could be to introduce the parametric core alongside the current one — or make `Dimension` a thin `structure` wrapping `Dimension PhyslibBase`, keeping the named projections and `@[simp]` lemmas — so existing `Units/` code compiles unchanged while a new development can pick a different basis. I'm happy to help prototype if there's interest.

**Open questions for maintainers:**

- **One basis, or many?** `HasDim M` gives a single `Dimension`, so a development effectively shares one basis. Is the intended model a single default basis (`PhyslibBase`) for the library, with parametricity available to *separate* developments — or per-type / mixed bases (which then need change-of-basis maps to interoperate)?
- **Replace or extend?** Should the parametric type replace `Dimension`, or sit beside it (say, `Dimension'`) behind a deprecation window?
- **`B → ℚ` or `B →₀ ℚ`?** A total function over a `Fintype B` (simplest; matches the current fixed-rank record) or finitely-supported functions (allowing an open-ended basis)?
- **Conventions.** Systems like natural units impose identities among generators (`c = ħ = 1`). Are these best modelled by choosing a smaller `B` and deriving the rest, or by an explicit relation/quotient layer on top of a fixed `B`?

### Prototype

A self-contained sketch (compiles as-is against a current Mathlib) — the commutative-group structure really does come for free, and a different basis with rational exponents just works:

```lean
import Mathlib

/-- A dimension over a chosen generating set `B`: a `ℚ`-valued exponent vector,
written multiplicatively so that composing dimensions adds exponents. -/
abbrev Dimension (B : Type) := Multiplicative (B → ℚ)

namespace Dimension
variable {B : Type}

-- The commutative-group structure (`*`, `1`, `⁻¹`, `/`) is inherited for free:
example : CommGroup (Dimension B) := inferInstance

/-- Rational powers: scale the exponent vector (the `ℚ`-module action on `B → ℚ`). -/
instance : Pow (Dimension B) ℚ :=
  ⟨fun d q => Multiplicative.ofAdd (q • Multiplicative.toAdd d)⟩

/-- Base dimension `b`: exponent `1` at `b`, `0` elsewhere. -/
def base [DecidableEq B] (b : B) : Dimension B := Multiplicative.ofAdd (Pi.single b 1)

end Dimension

/-- The current five-generator (charge-based) basis — the drop-in default. -/
inductive PhyslibBase | length | time | mass | charge | temperature
  deriving DecidableEq
abbrev PhyslibDimension := Dimension PhyslibBase

/-- A different generating set: Gaussian–CGS has three generators, charge derived. -/
inductive GaussianBase | length | mass | time
  deriving DecidableEq

open GaussianBase in
/-- Gaussian charge `M^(1/2)·L^(3/2)·T^(-1)` — half-integer exponents, fine over `ℚ`. -/
def gaussianCharge : Dimension GaussianBase :=
  Dimension.base mass ^ (1/2 : ℚ) * Dimension.base length ^ (3/2 : ℚ)
    * Dimension.base time ^ (-1 : ℚ)
```

`Multiplicative (B → ℚ)` reuses Mathlib's `Pi` group/module instances, so only the `ℚ`-power (a one-liner over the module action) and the base-vector constructor need writing; back-compat can then wrap `Dimension PhyslibBase` in a `structure` carrying the named projections.

## Requirements

A tickbox list of the requirements of the API. This should include things which have already been done as well as things still to be worked on. Details on writing good system requirements can be found [here](https://www.ibm.com/docs/en/SSYQBZ_9.6.1/com.ibm.doors.requirements.doc/topics/get_it_right_the_first_time.pdf).

**Already satisfied by the current `Dimension` (to be preserved by the generic API):**

- [x] Dimension exponents shall be represented over `ℚ`, so fractional powers are representable (e.g. the Gaussian `charge = M^(1/2)·L^(3/2)·T^(-1)`). *(`structure Dimension` over `ℚ`; `instance : Pow Dimension ℚ`)*
- [x] Dimensions shall form a commutative group under multiplication — composition adds exponents, inversion negates them, and `1` is the dimensionless element. *(`instance : CommGroup Dimension`)*
- [x] Dimensional equality shall be extensional. *(`@[ext] lemma Dimension.ext`)*

**Core — parametrize the generating set:**

- [ ] `Dimension` shall be parametric in a finite index type `B` enumerating the base dimensions (the generating set), so that the rank and identity of the basis are inputs, not fixed — e.g. `Dimension (B : Type) [Fintype B] [DecidableEq B]`.
- [ ] For every finite `B`, the commutative-group and `ℚ`-power structure on `Dimension B` shall be derived generically (e.g. from `Pi`/`Finsupp`), not restated per basis.
- [ ] `Dimension B` shall have decidable equality whenever `B` has `DecidableEq` and `Fintype`.
- [ ] For each `b : B`, the API shall provide the corresponding unit base vector (the generic analog of `L𝓭`, `T𝓭`, …): exponent `1` at `b`, `0` elsewhere.
- [ ] The unit side of the bridge shall be parametrized in the **same** basis `B`: `UnitChoices` (currently five named unit fields) shall become a unit choice per `b : B`, and `dimScale : UnitChoices B → UnitChoices B → Dimension B →* ℝ≥0` shall fold over `B` generically — so `HasDim` / `CarriesDimension` / `HasDimension` carry the basis consistently. *(Parametrizing `Dimension` alone is insufficient; the same five bases are hardwired again in `UnitChoices`.)*

**Backward compatibility:**

- [ ] PhysLib's current basis shall be recoverable as a specific instantiation `Dimension PhyslibBase` over a concrete enumeration `PhyslibBase = {length, time, mass, charge, temperature}` (charge-based; a faithful SI/ISQ basis is a *separate* instance — see Coverage).
- [ ] For that SI instantiation, the existing named projections (`.length`, `.time`, `.mass`, `.charge`, `.temperature`) and their `@[simp]` lemmas shall remain available, so current downstream code compiles unchanged (e.g. via a thin `structure` wrapper or abbreviations over the parametric core).

**Coverage — the motivating quantity systems (each a verifiable instantiation):**

- [ ] The full ISQ seven-base-quantity system shall be expressible — adding *amount of substance* and *luminous intensity*, and taking *electric current* in place of charge, per ISO/IEC 80000-1.
- [ ] Gaussian–CGS shall be expressible: three generators `L, M, T`, with charge a *derived* `ℚ`-vector `M^(1/2)·L^(3/2)·T^(-1)` rather than an independent axis.
- [ ] A natural-unit system (`c = ħ = 1`, reduced generating set) shall be expressible.
- [ ] An angle-augmented basis shall be expressible: an added `angle` generator `A`, with *solid angle* the derived square `A²` (`sr = rad²`).

**Optional / stretch:**

- [ ] The API should provide a change-of-basis map for an embedding of bases `B ↪ B'` (e.g. `PhyslibBase ↪` an angle-augmented basis), so a dimension in one system can be re-expressed in an extending one. *(non-blocking; supports migration and cross-system comparison)*
- [ ] The design, the SI recovery, and at least one worked non-SI instantiation shall be documented (as an example or test).

### References

- P. Quincey, *The range of options for handling plane angle and solid angle within a system of units*, Metrologia **53** (2016) 840. doi:10.1088/0026-1394/53/2/840
- I. Mills, *On the units radian and cycle for the quantity plane angle*, Metrologia **53** (2016) 991. doi:10.1088/0026-1394/53/3/991
- P. Quincey, R. J. C. Brown, *A clearer approach for defining unit systems*, Metrologia **54** (2017) 454. doi:10.1088/1681-7575/aa7160
- P. Quincey, P. J. Mohr, W. D. Phillips, *Angles are inherently neither length ratios nor dimensionless*, Metrologia **56** (2019) 043001. doi:10.1088/1681-7575/ab27d7
- M. I. Kalinin, *On the status of plane and solid angles in the International System of Units (SI)*, Metrologia **56** (2019) 065009. doi:10.1088/1681-7575/ab3fbf
- B. P. Leonard, *Proposal for the dimensionally consistent treatment of angle and solid angle by the International System of Units (SI)*, Metrologia **58** (2021) 052001. doi:10.1088/1681-7575/abe0fc
- P. J. Mohr, E. L. Shirley, W. D. Phillips, M. Trott, *On the dimension of angles and their units*, Metrologia **59** (2022) 053001. doi:10.1088/1681-7575/ac7bc2
- ISO/IEC 80000-1 *Quantities and units — General*
- VIM, 4th ed. 2nd Committee Draft, July 31, 2023. https://www.bipm.org/documents/20126/115700832/VIM4_2CD_clean/c6d0dfb2-ddbf-059e-1f74-9b025c9c59d8

## Corresponding file system

The `Dimension` type is used **only** within `Physlib/Units/` — 11 files (the one match elsewhere is an unrelated doc-heading, "Dimension 7 plane", in `Particles/…/DimSevenPlane.lean`) — and the `HasDim` / `CarriesDimension` carrier layer appears in 10 of them. So the migration surface is a single subsystem, not the whole library.

**Core — the type**
- `Physlib/Units/Dimension.lean` — the `Dimension` record, its `CommGroup` / `Pow ℚ` structure, and the SI base vectors `L𝓭`, `T𝓭`, `M𝓭`, `C𝓭`, `Θ𝓭`.

**Unit/carrier bridge — the same five bases are hardwired again here**
- `Physlib/Units/Basic.lean` — `UnitChoices` (one unit field per base dimension; five fields), `UnitChoices.dimScale : UnitChoices → UnitChoices → Dimension →* ℝ≥0` (folds over the five named fields), and the carrier layer `HasDim` / `dim` / `CarriesDimension` / `HasDimension` / `Dimensionful`.

**Dimensioned-quantity machinery and worked instances (built on the above)**
- `Physlib/Units/UnitDependent.lean`, `Physlib/Units/Integral.lean`, `Physlib/Units/FDeriv.lean`, `Physlib/Units/Examples.lean`
- `Physlib/Units/WithDim/{Basic, Area, Energy, Momentum, Pressure, Speed}.lean` — worked dimensioned types.

NicolasRouquette:
Closes #1441 — makes `Dimension`'s base-dimension set parametric, and moves the unit-side twin (`UnitChoices` / `dimScale`) with it, as the issue requires.

Thanks to **Daniel Morrison** for a very useful and constructive design discussion on the Physlib Zulip that shaped this.

This builds on #1445 (the v4.32.0 bump); the substantive change is the final commit, `feat(Units): parameterise Dimension and its unit twin over a basis`.

## Requirements coverage (from the issue)

**Already satisfied — preserved by the generic API**
- [x] ℚ exponents / fractional powers — `exponent : B → ℚ`, `Pow (Dimension B) ℚ`
- [x] commutative group — `instance : CommGroup (Dimension B)` for every `B`
- [x] extensional equality — `@[ext] Dimension.ext`

**Core — parametrise the generating set**
- [x] `Dimension` parametric in a basis `B` — `structure Dimension (B)`
- [x] `CommGroup` / `ℚ`-power derived generically for every `B`
- [x] `DecidableEq (Dimension B)` for finite `B` — `instance [Fintype B] : DecidableEq (Dimension B)`
- [x] generic base vector per `b : B` — `Dimension.single b`, with `L𝓭 = single .length`, … (`*_eq_single`)
- [x] unit side parametrised in the **same** basis; `dimScale` folds over `B` — `UnitScale B` and `UnitScale.dimScale : UnitScale B → UnitScale B → Dimension B →* ℝ≥0` (`d ↦ ∏ b, (u₁ b / u₂ b) ^ d.exponent b`); `UnitChoices.toScale` exhibits the current five-field `UnitChoices` as the `PhyslibBase` instance

**Backward compatibility**
- [x] current basis recoverable — `Dimension PhyslibBase` over `PhyslibBase = {length, time, mass, charge, temperature}` (neutrally named; charge-based, i.e. *not* the SI/ISQ set)
- [x] named projections (`.length`, …) and `@[simp]` lemmas preserved — every existing `Units/` result builds unchanged (full build: 9238 jobs)

**Coverage — the motivating quantity systems** (`Physlib/Units/BasisExamples.lean`)
- [x] ISQ seven base quantities — current-based, with amount of substance and luminous intensity
- [x] Gaussian–CGS — three generators, charge derived `M^(1/2)·L^(3/2)·T^(-1)`
- [x] natural units — `c = ħ = 1`, a single generator (length = time = `mass⁻¹`)
- [x] angle-augmented — solid angle the derived square of plane angle (`sr = rad²`)

**Optional / stretch**
- [x] change-of-basis map for `B ↪ B'` — `Dimension.extend` (exponent-preserving on an embedding), with a `PhyslibBase ↪` angle-augmented demo
- [x] documented SI recovery + worked non-SI instantiations — the two example modules

## Design decisions on the issue's open questions

- **`B → ℚ` vs `B →₀ ℚ`** — total function (`structure Dimension (B) where exponent : B → ℚ`); `Fintype B` is required only where a fold is taken (`dimScale`, `DecidableEq`). This keeps the dimension-comparison cast (below) ergonomic.
- **Replace or extend** — the parametric type *replaces* `Dimension`; the current type is recovered as `Dimension PhyslibBase` with the projections re-exposed (the issue's "thin wrapper" path), so nothing downstream changes.
- **One basis or many** — a single default basis (`PhyslibBase`) for the `HasDim` / `CarriesDimension` carrier layer, with parametricity available to separate developments and to the other bases above. (So the unit-scaling carrier still shares one basis; per-type mixed bases are not introduced.)
- **Conventions (`c = ħ = 1`)** — modelled by choosing a smaller basis and deriving the rest (see `Natural`), the first option the issue lists.

## Dimension comparison

`WithDim` is parametric over the basis. A quantity of dimension `length` and a product of quantities of dimension `length/time` and `time` have *propositionally* but not *definitionally* equal dimensions (`rfl` and `decide` both fail — the equality is group cancellation over ℚ exponents). `WithDim.cast` bridges the two types, discharging the equality automatically; see `Physlib/Units/ParametricDimensionExamples.lean`.

## Checks

`lake build` (Physlib + QuantumInfo): 9238 jobs, 0 errors / 0 warnings / 0 `sorry`. `lint-style.py` and `lake exe lint_all` clean on all changed files. No new axioms beyond `propext` / `Classical.choice` / `Quot.sound`; no `sorryAx`.

---

*AI-assisted per the repository's [AI policy](AI-POLICY.md); commits carry the `Co-authored-by` trailer.*


NicolasRouquette:
Note on dependencies and review order.

This PR is stacked on #1445 (the v4.32.0 toolchain bump). Since it targets master, the first two commits — 48fad301 and 978ce094 — belong to #1445 and will drop out of this diff automatically once #1445 merges. The substantive change here is the final commit, 97c70b5c (feat(Units): parameterise Dimension and its unit twin over a basis); that's the one to review.

Because of that dependency:

- It's probably easiest to review/merge #1445 first — this PR then reduces to the single parametric commit with nothing further needed from me.
- If #1445 takes further review changes (new or amended commits), I'll rebase this branch onto the updated #1445 head so the two stay consistent. So please expect the occasional force-push here for that reason, rather than as a change to this PR's own content.

Since the original issue arose in the context of my metrology work which depends on Caltech's [TorchLean](https://github.com/lean-dojo/torchlean) that already migrated to 4.32.0, I chose to stack this PR on top of #1445.

NicolasRouquette:
-awaiting-author

Separate from the toolchain fix — a naming question raised in review. It came to me by email and I couldn't find it in the PR comments, so I'm recording it here for visibility:

> Is there a better name for this which doesn't refer to Physlib?

For context: this is `PhyslibBase`, the top-level inductive in `Physlib/Units/Dimension.lean` (its fully-qualified name is the bare `PhyslibBase`). In this PR `Dimension` becomes parameterised over a _basis_ `B` of base dimensions — `Dimension B` — and `PhyslibBase` is PhysLib's default basis: the five generators length, time, mass, charge, temperature. `Dimension PhyslibBase` reproduces the previous five-exponent `Dimension `exactly, so existing code is unaffected.

I named it after its provenance rather than generically because this basis is a specific, deliberate choice, not a neutral or canonical one. It is _charge_-based with five generators, so it is not the SI/ISQ base-quantity set (which takes electric current as base and adds amount of substance and luminous intensity); ISQ, Gaussian–CGS, natural units, and the angle-augmented system are other instantiations of `Dimension B` for different `B`. 

The previous unparameterised `Dimension` had this choice baked in implicitly, so the generic name was arguably misleading — it looked basis-neutral but wasn't. Tagging the basis with its origin documents that it is PhysLib's historical convention, one option among several.

That said, I'm very open to renaming it. My one constraint is that the name should still signal "a specific, non-ISQ choice of base quantities" rather than implying a canonical basis. If there's a preferred convention — or a good descriptive name for this charge-based, five-generator set — I'll gladly adopt it. Suggestions welcome.

NicolasRouquette:
Thanks @nateabr for the retitle (and the Stacked on #1445 note — that's the right base).

Want to confirm scope before I build anything further, since "and its unit twin" could mean two things.

What's in the PR today is _additive_: `Dimension B` is parameterised over a basis, and the unit side gets a basis-generic twin `UnitScale B` (`scale : B → ℝ≥0` + positivity, with `dimScale : UnitScale B → UnitScale B → Dimension B →* ℝ≥0` proved once via `Finset.prod`). 

The existing five-field `UnitChoices` is untouched; `UnitChoices.toScale` exhibits it as the `PhyslibBase` instance of the generic fold. `UnitScale` lives entirely in the new file; nothing downstream moves.

The open question is whether you also want to parameterise `UnitChoices` itself. I see four levels:

### A — Keep the bridge (status quo).

`UnitChoices` stays bespoke; `toScale` is the equivalence. Generic machinery is available for other bases (ISQ, Gaussian–CGS, natural, angle-augmented) without disturbing anything.

- Pro: smallest diff, cleanly stacked, zero risk to the 107 existing UnitChoices call sites. 
- Con: the ~12 hand-rolled UnitChoices.dimScale lemmas remain a parallel copy of the generic ones — the two sides are twinned but not unified.

### B — Consolidate the proofs, keep the API. 

Leave the `UnitChoices` structure and its public surface as-is, but re-derive `UnitChoices.dimScale` and its lemmas through `toScale/UnitScale.dimScale`, deleting the duplicated manual proofs.

- Pro: single source of truth for the scaling laws; blast radius is `Units/Basic.lean` only; downstream and `SI`/API unchanged. 
- Con: a little proof-plumbing; `UnitChoices` stays `PhyslibBase`-only (which is arguably correct — it's the default basis).

### C — Make `UnitChoices` fully basis-generic. 

Fold `UnitChoices` into `UnitScale PhyslibBase` and migrate the 107 references, so quantity code can be written over any basis.

- Pro: genuinely generic unit choices end-to-end. 
- Con: invasive across 12 files incl. `UnitDependent`; `UnitScale B`'s `scale : B → ℝ≥0` erases the `LengthUnit`/`MassUnit`/… _typed_-unit names (they carry no algebraic content, but they're real ergonomics/type-safety). Hard to review while stacked on the bump.

### D — Basis-generic and typed.

Attach to a basis `B` a family of unit types, one per base dimension, each carrying a positive-real magnitude — then let the typed layer project onto the magnitude layer already in the PR:

```lean
class UnitSystem (B : Type) where
  Unit    : B → Type                    -- the typed unit at each base
  mag     : {b : B} → Unit b → ℝ≥0      -- its magnitude
  mag_pos : ∀ {b} (u : Unit b), 0 < mag u

/-- A typed unit choice: pick a typed unit at every base dimension. -/
def TypedChoice (B : Type) [UnitSystem B] := (b : B) → UnitSystem.Unit b

/-- Forget the types down to the magnitude layer already in the PR. -/
noncomputable def TypedChoice.toScale {B : Type} [UnitSystem B]
    (u : TypedChoice B) : UnitScale B where
  scale     b := UnitSystem.mag (u b)
  scale_pos b := UnitSystem.mag_pos (u b)

/-- `PhyslibBase`'s instance recovers exactly today's typed names. -/
noncomputable instance : UnitSystem PhyslibBase where
  Unit
    | .length => LengthUnit | .time => TimeUnit | .mass => MassUnit
    | .charge => ChargeUnit | .temperature => TemperatureUnit
  mag {b} := match b with
    | .length | .time | .mass | .charge | .temperature =>
        fun u => ⟨u.val, u.val_pos.le⟩
  mag_pos {b} := match b with
    | .length | .time | .mass | .charge | .temperature =>
        fun u => NNReal.coe_pos.mp u.val_pos
```

Now `u .length : LengthUnit`, `LengthUnit.meters` still works, and you cannot put a `MassUnit` in the length slot — the dependent type forbids it, same guarantee as the original record.

- Pro: the only option that keeps _both_ properties that matter — full basis-genericity and the typed-unit names/safety. No proof duplication either: the scaling laws stay proved once on the magnitude layer `UnitScale B`, and the typed layer is a thin faithful wrapper that projects onto it via `toScale`.
- Pro (already validated): the sketch above isn't hypothetical -- it compiles as-is against this PR (`lake env lean` over `Physlib.Units.ParametricUnits`), the `PhyslibBase` instance and the `u .length : LengthUnit` type-safety check included.
- Con: the most machinery of the four — a new `UnitSystem` class plus a dependent `(b : B) → Unit b` encoding that's marginally less `simp`/elaboration-friendly than a flat record; and its payoff over `B` only materialises once a second typed basis actually exists.

Why D makes the most sense. The other three each give up something: 
- A leaves the two sides duplicated; 
- B unifies the proofs but stays `PhyslibBase`-only; 
- C reaches genericity but discards the typed-unit names that are the whole reason `UnitChoices` is nicer than a bare `B → ℝ≥0`.

D is the only one that is generic and type-safe, and it does so by reusing what the PR already has — the laws live once on `UnitScale B`, with the typed layer sitting cleanly on top.

But the blast radius is real, so I want your concurrence before expanding the diff. D can land two ways:

- D-additive (my preference for this PR) — add `UnitSystem`/`TypedChoice`/`toScale` alongside, keep `UnitChoices`, and prove `UnitChoices ≃ TypedChoice PhyslibBase`. Zero churn to the 107 call sites; reviewable stacked on #1445.

- D-migrate — redefine `UnitChoices := TypedChoice PhyslibBase` and migrate all 107 references. Type-safe, but as invasive as C; better as its own follow-up PR.

Does D (and specifically D-additive for this PR) match the direction you had in mind, or would you rather keep #1447 to A/B and take the typed-generic layer D separately?

nateabr:
Hey @NicolasRouquette, please give me a bit to go through the details of your PR and your comment and I will get back to you. Many apologies for the delay

nateabr:
The core abstraction you have done with parametric Dimension B, PhyslibBase, and backward compat layer looks clean and I'm happy with the math and the proofs. However, the three new files (BasisExamples, ParametricDimensionExamples, ParametricUnits) and the extend/single API have no current use at the moment and seem to be illustrative. My first thought would be to focus the PR to the core parts. 

However, even with regards to the core abstraction you have provided; I see the utility in having a parametric type that we could possibly instantiate with a different basis in the future. But purely from a maintenance perspective, this does add complexity to the core dimension type. Furthermore, if the motivation for this PR does lie in a downstream repo, would it make sense for this machinery to live in that repo and not Physlib?

That being said, I do believe I am unable to provide you with a constructive review at the moment, the above is more of my thoughts out loud. So I would say hold off on any changes and wait for a better look from a maintainer.

NicolasRouquette:
Thanks @nateabr for the feedback.

There were already several examples in the `Physlib` library per se, so it seemed to me sensible to add other examples for the parametric dimension and basis. However, we're getting to the point where it would make sense to move them out of the library:

```lean
public import Physlib.QuantumMechanics.HarmonicOscillator.OneDimension.Examples
public import Physlib.QuantumMechanics.Operators.Examples
public import Physlib.Units.BasisExamples
public import Physlib.Units.Examples
public import Physlib.Units.ParametricDimensionExamples
```

If someone can provide guidance where to move the unit-related examples, e.g., a `./examples/physlib` folder with a separate `PhyslibExamples` library in the lakefile, I'm OK in doing that refactoring in this PR.

jstoobysmith:
I think for the units examples we could put them in .Units/Examples for now. Will hopefully get chance to look at this tomorrow. But generally I think the concept of this is good, and should be included.

jstoobysmith:
Would it be possible to rebase this so only the relevant changes are shown? Also I think a PhyslibBase could have a better more physical name if still called that

NicolasRouquette:
Thanks @jstoobysmith. Done on the first point, and a pointer on the second.

Rebased to reduce the scope. #1445 landed on master (squashed as c217bddd), so I rebased this branch onto current master. The four v4.32.0 bump-chain commits dropped out — they're already in master — leaving a single feature commit. The PR now shows only the parametric change: 9 files under Physlib/Units/ (Dimension, Basic, UnitDependent, WithDim/Basic, Integral, the new ParametricUnits/BasisExamples/ParametricDimensionExamples, and the Physlib.lean imports), with no toolchain collateral. The feature patch itself is unchanged — only the base moved.

On the name. Agreed that `PhyslibBase` is worth reconsidering, and I'd raised the same question earlier here: [#1447 (comment)](https://github.com/leanprover-community/physlib/pull/1447#issuecomment-5025421830). Short version: it's the default basis B for the now-parameterised Dimension B — the charge-based five-generator set (length, time, mass, charge, temperature), which is deliberately not the SI/ISQ base-quantity set. I named it after its provenance to flag that it's a specific choice rather than a canonical one, but I'm very open to a more physical/descriptive name. My only constraint is that it should still signal "a specific, non-ISQ choice of base quantities" rather than implying a neutral basis. If you have a preferred convention or a name in mind for this charge-based five-generator set, I'll gladly adopt it.

Physlib/Units/BasisExamples.lean
jstoobysmith, line 1: Actually would just remove `SpecificBasis.lean` might be better, or seperate each to their own file.

Physlib/Units/Dimension.lean
```diff
@@ -21,177 +39,271 @@ open NNReal
...
 /-!
 
+## The PhysLib base
+
+-/
+
+/-- PhysLib's default basis of base dimensions — `length`, `time`, `mass`,
+  `charge`, `temperature`. Note this is *charge*-based, so it is not the SI/ISQ
+  base-quantity set; see `Dimension`. -/
+inductive PhyslibBase where
```
jstoobysmith: Could we even just remove this and just use the Basis in the current `BasisExamples.lean`?

Physlib/Units/BasisExamples.lean
```diff
@@ -0,0 +1,179 @@
...
+  amount of substance, luminous intensity. Unlike `PhyslibBase`, current is base and
+  charge is derived. -/
+inductive ISQBase
+  | length | time | mass | current | temperature | amount | luminousIntensity
+  deriving DecidableEq, Fintype
+
+namespace ISQ
+
+/-- The ISQ has seven base quantities. -/
+example : Fintype.card ISQBase = 7 := by decide
```
jstoobysmith: This could be an actual lemma (similar with the below).

jstoobysmith:
awaiting-author

Physlib/Units/ParametricDimensionExamples.lean
```diff
@@ -0,0 +1,100 @@
...
+-/
+
+@[expose] public section
+
+open Dimension
+
+namespace ParametricDimensionExamples
+
+/-!
+## Daniel Morrison's example on the standard basis
```
jstoobysmith: Would remove this doc-string

Physlib/Units/ParametricDimensionExamples.lean
```diff
@@ -0,0 +1,100 @@
...
+/-- A basis of information-theoretic base dimensions. -/
+inductive Info
+  /-- The information base dimension (bits). -/
+  | bit
+  /-- The symbol base dimension. -/
+  | symbol
+
+/-- The `bit` base dimension. -/
+def bitDim : Dimension Info := ⟨fun | .bit => 1 | .symbol => 0⟩
+/-- The `symbol` base dimension. -/
```
jstoobysmith, line 85: This should have a new line between the definitions

NicolasRouquette:
Thank you @jstoobysmith for the careful, constructive review — it genuinely sharpened both the scope and the design. Two of your observations reshaped the PR:

**1. Two concrete bases instead of illustrative examples.** You were right that the example modules carried no weight. I removed them and kept exactly two instantiations:
- `LTMCTDimensionBase` — the existing basis, renamed after its five base dimensions (**l**ength, **t**ime, **m**ass, **c**harge, **t**emperature) so the name announces a specific, deliberately non-ISQ choice rather than "the" basis; and
- `ISQDimensionBase` — the seven ISO/IEC 80000-1 (VIM 3) base quantities, as a reference.

`Dimension` is now trimmed to the generic algebra, with LTMCT's projections and generators in their own file.

**2. Truth-preserving maps between bases, enforced by construction.** This is where I'd argue from metrology. Relating two systems of quantities is *not* a relabelling of base symbols: base-vs-derived is a choice the system makes, not a property of the quantity. Electric charge is a base dimension in the LTMCT system but the *derived* `I·T` in the ISQ. The label injection I'd first sketched silently permitted exactly that category error — it could map the charge generator onto the current generator, equating two dimensionally distinct quantities.

So I made the admissible maps first-class, as bundled requirements on `Dimension`:
- `Dimension.Embedding` — an **injective** monoid homomorphism of dimensions (a faithful, dimension-preserving inclusion into a richer basis); and
- `Dimension.Projection` — a **surjective** monoid homomorphism (a truth-preserving *reduction* onto a coarser basis).

Requiring a `MonoidHom` is what enforces dimensional truth — the map must respect products, inverses and rational powers — so a relabelling that conflates inequivalent base quantities is simply not constructible as either. The LTMCT↔ISQ bridge is then the metrologically honest one: the embedding sends charge to the derived `I·T`; the projection reads current as charge/time and forgets amount-of-substance and luminous-intensity; and `projection ∘ embedding = id` states precisely that ISQ reduces onto LTMCT, but not the reverse without *adding* those base quantities back.

Green across the full build and the style/meta linters. Happy to iterate.

Physlib/Units/BasisExamples.lean
NicolasRouquette: Addressed in [d4c7056](https://github.com/leanprover-community/physlib/pull/1447/commits/d4c705659801b5615b5209f182940539b2255d49)

Physlib/Units/ParametricDimensionExamples.lean
NicolasRouquette: Done in [d4c7056](https://github.com/leanprover-community/physlib/pull/1447/commits/d4c705659801b5615b5209f182940539b2255d49)

Physlib/Units/ParametricDimensionExamples.lean
NicolasRouquette: Ran linters before pushing [d4c7056](https://github.com/leanprover-community/physlib/pull/1447/commits/d4c705659801b5615b5209f182940539b2255d49)

NicolasRouquette:
-awaiting-author

jstoobysmith (approved):
Approved - many thanks. Will merge shortly.

PR #1448, opened 2026-07-20
feat(QuantumMechanics): add algebraic properties for mulOperator

gloges:
Adds some basic results about multiplication operators and smul/neg/add/sub:
- `c • 𝓜 μ f ≤ 𝓜 μ (c • f)` with equality when `c ≠ 0`,
- `𝓜 μ (-f) = -𝓜 μ f`,
- `𝓜 μ f + 𝓜 μ g ≤ 𝓜 μ (f + g)` (and a `sub` variant).

### DIFF
--- a/Physlib/QuantumMechanics/HilbertSpaces/SpaceD/Basic.lean
+++ b/Physlib/QuantumMechanics/HilbertSpaces/SpaceD/Basic.lean
@@ -129,6 +129,9 @@ lemma MemHS.sub (hf : MemHS f μ) (hg : MemHS g μ) : MemHS (f - g) μ := MemLp.
 
 lemma MemHS.const_smul (c : ℂ) (hf : MemHS f μ) : MemHS (c • f) μ := MemLp.const_smul hf c
 
+lemma MemHS.const_smul_iff {c : ℂ} (hc : c ≠ 0) : MemHS (c • f) μ ↔ MemHS f μ :=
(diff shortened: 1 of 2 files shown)


nateabr:
This looks good to me. One comment is that it would be good to have more documentation than is available at the moment. Some doc strings for key lemmas, and a general elaboration on "why" behind the code would be great for readability and would improve the file greatly.

gloges:
Thanks, I'll aim to improve the documentation for this and other QM files in the near future.

nateabr:
Okayy sounds good, just one more question: 

For smul and composition, the PR (and existing code) follow a nice proof pattern. However, for add, we only get the inequality `𝓜 f + 𝓜 g ≤ 𝓜 (f + g)`. Would it be worth adding something like `mulOperator_add_eq` for the case where both domains are ⊤? Maybe a todo?

nateabr:
awaiting-author

gloges:
That's a good point. The condition `c ≠ 0` is particularly simple for `mulOperator_smul_eq` but for `add/sub` it seems much messier. I've added `mulOperator_add_eq` (and a `sub` variant) which uses the hypothesis `h : (𝓜 μ g).domain = ⊤` and a TODO to expand on this; I'm not sure at this point what conditions would be most useful to pin down.

nateabr:
Okayy looks good to me, will tag as reviewer approved :)

jstoobysmith (approved):

PR #1449, opened 2026-07-20
refactor(QuantumMechanics): rename SelfAdjoint to Symmetric

KrystianNowakowski:
`QuantumMechanics.OneDimension.UnboundedOperator.IsSelfAdjoint` in `Physlib/QuantumMechanics/Operators/OneDimension/Unbounded.lean` is defined as

`∀ ψ1 ψ2 : S, ⟪U ψ1, ι ψ2⟫_ℂ = ⟪ι ψ1, U ψ2⟫_ℂ`

which is the Hermicity — i.e. symmetry. It has no domain or density hypothesis, so it does not capture genuine self-adjointness, (`A† = A`, including equality of domains). For an unbounded operator on a proper dense core the two are not equivalent (The `DDimensions` operator API already draws this line correctly.)

Therefore, the PR renames to `IsSymmetric` and updates the three lemmas built on it, plus docstring/headings.

### DIFF
--- a/Physlib/QuantumMechanics/Operators/OneDimension/Momentum.lean
+++ b/Physlib/QuantumMechanics/Operators/OneDimension/Momentum.lean
@@ -117,11 +117,11 @@ lemma planeWaveFunctional_generalized_eigenvector_momentumOperatorUnbounded (k :
 
 /-!
 
-## The momentum operator is self adjoint
+## The momentum operator is symmetric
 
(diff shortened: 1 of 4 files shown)


gloges (approved):
Hi @KrystianNowakowski, thanks for making your first PR and correcting this error in nomenclature. These changes look good to me!

PR #1450, opened 2026-07-20
docs(fluid-dynamics): add fluid API map

Robby955:
## Motivation

Related to #1414, the tracking map for the fluid API in #887. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/FluidDynamics/API-map.yaml`.

The completed entries cover what exists: the mass flux density `momentumDensity` and the equation of continuity in its conservative, residual and smooth forms with the implication between them. The fluid data structure is recorded as not done, since `FluidState` currently provides density and velocity but the requirement also asks for pressure. The Euler equation, ideal fluid, entropy flux, Bernoulli equation, energy flux and flow out of a volume are not implemented and carry location `N/A`.

References cite Landau and Lifshitz, Fluid Mechanics, vol. 6, 2nd ed., Chapter 1.

## Testing

Ran the API-map linter:

```
```

Every declared source file and name resolves; no completed entry is missing a location.

### DIFF
--- a/Physlib/FluidDynamics/API-map.yaml
+++ b/Physlib/FluidDynamics/API-map.yaml
@@ -1,0 +1,84 @@
+version: v0.1
+
+Title: "Fluid"
+
+Overview: |
+  A fluid is described by the spatial distribution of its velocity, density, and
+  pressure fields. This is a core data structure of fluid dynamics from which the
+  continuity equation, the equations of motion, and thermodynamic balance laws are
+  built (Landau and Lifshitz, Fluid Mechanics, Chapter 1).
+
+  This map records the current state of Physlib/FluidDynamics.
+  Partial infrastructure exists. `FluidState` (density and velocity fields) is
+  defined in `FluidDynamics/FluidState.lean`, together with the field-type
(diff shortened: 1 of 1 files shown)


jstoobysmith:
This will likely need an update based on Physlib#1125, which I will merge now.

jstoobysmith:
Adding the tag 

awaiting-author 

for comment above.

Robby955:
Rebased onto master and updated the locations for the new structure. The Euler equation, the ideal fluid and the Bernoulli equation are marked done now, since #1125 formalized them. Flow out of a volume, the entropy flux density and the energy flux are still open. Entropy exists as a field of `ThermodynamicCauchyFlow`, but the flux density built from it does not, so I left that one open rather than pointing it at the field.

jstoobysmith (approved):
Approved - many thanks. Will merge shortly (just a reminder to comment with `-awaiting-author` when you've finished with changes, as it helps keep track of what needs reviewing :)).

PR #1456, opened 2026-07-22
docs(QuantumMechanics): add Hilbert space API map

gloges:
Begins an API map for Hilbert spaces on `Space d`.

This covers the L²-space `SpaceDHilbertSpace`, its important subspaces, and operators acting on it. About half of the included items are already completed, while the rest represent a basic "wish list" for some missing subspaces and operators which should serve useful in the future.


jstoobysmith (approved):
Looks good - approved

PR #1463, opened 2026-07-25
feat: Add down singlets

nateabr:
This PR adds the corresponding API for down-type quark singlets.


jstoobysmith (approved):
Looks good - approved

PR #1465, opened 2026-07-27
feat(Mathematics): `SchwartzMap` Gaussians

gloges:
Defines Gaussians as Schwartz maps `𝓢(E, 𝕜)` for any inner product space `E` and `RCLike 𝕜`.
Two technical lemmas are used to prove the decay property in the special case of the standard Gaussian with `𝕜 = ℝ`.
This is then composed with `RCLike.ofReal` and an invertible affine transformation to arrive at `exp (-2⁻¹ * ‖B⁻¹ (x - x₀)‖ ^ 2)` as a Schwartz map `𝓢(E, 𝕜)` with mean `x₀` and positive-definite covariance `BᵀB`.

In the future these can be used to construct states in the Schwartz submodule of `SpaceDHilbertSpace`, for example the energy eigenstates of the quantum harmonic oscillator, Gaussian wave packets, and trial wavefunctions for the variational method.

### DIFF
--- a/Physlib/Mathematics/InnerProductSpace/Gaussian.lean
+++ b/Physlib/Mathematics/InnerProductSpace/Gaussian.lean
@@ -17,10 +17,20 @@ public import Mathlib.Analysis.SpecialFunctions.ExpDeriv
 
 @[expose] public section
 
-open ContinuousLinearMap Real
+namespace InnerProductSpace
 
+noncomputable section
+
+open ContinuousLinearMap Filter RCLike Real SchwartzMap
+
+variable {D : Type*} [NormedAddCommGroup D] [InnerProductSpace ℝ D]
 variable {E : Type*} [NormedAddCommGroup E] [InnerProductSpace ℝ E]
-variable (x : E)
+variable (𝕜 : Type*) [RCLike 𝕜]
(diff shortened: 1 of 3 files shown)


Physlib/Mathematics/InnerProductSpace/Gaussian.lean
jstoobysmith, line 1: There are lots of things about Gaussians in HilbertSpace.OneDimension.Gaussians.lean. Should we make a TODO to merge that file with this one?

gloges: Good idea - I've added to the TODOs.

gloges:
With a fresh eye I think it is better to have `gaussian₀` and `stdGaussian` as abbreviations; going forward results can be proved just for the general `gaussian 𝕜 B x₀`. For the same reason I've added a note to avoid using `realStdGaussian` directly.

-awaiting-author

Cobord:
There are a bunch of lemmas about general gaussian in https://github.com/Cobord/ZetaThetaFunctions, the general QuadraticForm and chirped gaussian stuff should be able to be copied directly over. Chirp means you should be able to handle the phase for coupling to electromagnetic field without as much extras.

jstoobysmith (approved):
Approved - looks good.

Issue #939, opened 2026-02-13
API: TimeAndSpace


jstoobysmith:
## Key data structure

This is Euclidean SpaceTime. The key data structure should be a type `TimeAndSpace d` which is an abbreviation of `Time x Space d`. 

- [ ] is defined.  

## Need

This API is needed throughout PhysLean, for example in Euclidean field theory, but also areas such as electromagnetism and classical mechanics.

## Requirements

- [ ] The API shall contain isometry projections from `TimeAndSpace d`  to `Time` and `Space.
- [ ] The API shall contain an action of the Euclidean group on `TimeAndSpace d`. 
- [ ] The API shall contain an action of the  Euclidean group on Schwartz maps with targets `TimeAndSpace d`. 

## Corresponding file system

The corresponding file system is: 
https://github.com/lean-phys-community/PhysLean/tree/master/PhysLean/SpaceAndTime/TimeAndSpace

## References 

Specifics of the material in this section appears here: 
- https://github.com/mrdouglasny/OSforGFF/tree/main

## Parent APIs 

#854 #855 #940

jstoobysmith:
Closing as included in code.

Issue #911, opened 2026-01-16
API: Galilean group


jstoobysmith:
## Key data structure

The key data structure is the type corresponding to the Galilean group.

- [ ] is defined 

## Need

The Galilean group is used in newtonian and non-relativistic physics, it contains many of the usual transformations one may want to do on `Space` and `Time`. 

## Requirements

- [ ] The API shall contain an instance of a group on the type corresponding to the Galilean group. 
- [ ] The API shall contain an instance of a Lie group on the type corresponding to the Galilean group. 
- [ ] The API shall contain an inclusion from the group of rotations. 
- [ ] The API shall contain an inclusions from the space-translation group. 
- [ ] The API shall contain an inclusion from the time-translation group. 
- [ ] The API shall be general for any dimension. 

## Corresponding file system

A file system is not currently defined for the Galilean group.

jstoobysmith:
Closing as included in code.

PR #1466, opened 2026-07-27
docs(space-and-time): add Galilean group and TimeAndSpace API maps

Robby955:
## Motivation

Related to #1414, the tracking maps for the Galilean group API in #911 and the TimeAndSpace API in #939. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/SpaceAndTime/GalileanGroup/API-map.yaml` and `Physlib/SpaceAndTime/TimeAndSpace/API-map.yaml`.

For the Galilean group, the completed entries cover the group structure on `GalileanGroup d`, its action on `Time × Space d`, and the inclusions of the Euclidean, orthogonal, rotation, space-translation and time-translation subgroups, together with the parameterisation by arbitrary spatial dimension. One requirement from #911 is recorded as not met: there is no Lie group instance on `GalileanGroup`.

For TimeAndSpace, the entries cover the abbreviation itself, the continuous linear projections onto `Time` and `Space d` with their distance bounds, the Euclidean group action, and the induced action on Schwartz maps out of `TimeAndSpace d`. Two entries beyond the requirements in #939 record the commutation of time and space derivatives and the characterisation of functions whose time or space derivative vanishes, both of which the API already provides.

## Checks

`scripts/api_map_linter.py --repo .` passes on both files with no missing files and no missing names, resolving 23 and 27 declarations respectively against the Lean sources.

### DIFF
--- a/Physlib/SpaceAndTime/GalileanGroup/API-map.yaml
+++ b/Physlib/SpaceAndTime/GalileanGroup/API-map.yaml
@@ -1,0 +1,57 @@
+version: v0.1
+
+Title: Galilean group
+
+Overview: |
+    The key data structure is `GalileanGroup d`, a structure carrying a spatial
+    rotation, a boost velocity, a space translation and a time translation, for
+    an arbitrary spatial dimension `d`. It is the group of transformations of
+    non-relativistic space and time, acting on `Time × Space d` by rotating and
+    boosting, then translating. The API contains the group structure, the action,
+    and inclusions of the Euclidean, orthogonal, rotation, space-translation and
+    time-translation subgroups.
+
+ParentAPIs:
+  - "Space (Physlib/SpaceAndTime/Space)"
+  - "Time (Physlib/SpaceAndTime/Time)"
+  - "Euclidean group (Physlib/SpaceAndTime/Space/EuclideanGroup)"
+
+References: []
+
+Requirements:
+
+  - description: "The key data structure corresponding to the Galilean group is defined."
+    done: true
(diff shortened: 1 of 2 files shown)


Physlib/SpaceAndTime/GalileanGroup/API-map.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+Requirements:
+
+  - description: "The key data structure corresponding to the Galilean group is defined."
+    done: true
+    location: "Physlib/SpaceAndTime/GalileanGroup/Basic.lean (GalileanGroup, with fields rotation, velocity, spaceTranslation, timeTranslation)"
+
+  - description: "The API contains an instance of a group on the type corresponding to the Galilean group."
+    done: true
+    location: "Physlib/SpaceAndTime/GalileanGroup/Basic.lean (One, Mul, Inv, Group (GalileanGroup d), with simp lemmas for each field under one, mul and inv)"
```
jstoobysmith: I would expect some of these locations to fail against the linter we made, though they look good?

Robby955:
Good catch, and the answer is that they were not being checked. `classify_claim` treats anything that is not a bare identifier as an anonymous instance and defers it to a Lean-environment check, so the prose in those locations, for instance "with simp lemmas for each field under one, mul and inv", took the deferral path rather than the verified one.

I have replaced the prose with the declaration names. GalileanGroup now resolves 33 names against the source with 5 deferred, up from 23 with 6, and TimeAndSpace resolves 28 with none deferred, up from 27 with 2.

jstoobysmith (approved):
Ok, looks good - approved.

PR #1467, opened 2026-07-27
docs(space-and-time): add SpaceTime API map

Robby955:
## Motivation

Related to #1414, the tracking map for the SpaceTime API in #856. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/SpaceAndTime/SpaceTime/API-map.yaml`.

The completed entries cover the definition of `SpaceTime d` and its coordinates, the projections onto `Space d` and `Time` with the equivalence to `Time × Space d`, derivatives of functions and of distributions along a coordinate, the splitting of integrals over `SpaceTime` into integrals over time and space, time slicing, and the action of the Lorentz group. Two entries beyond the requirements record the measure-space structure used for integration and the time-slicing maps.

Two entries differ from the checklist on #856. Derivatives of functions to manifolds are recorded as done, since `manifoldDeriv` in `Derivatives.lean` provides exactly that. Derivatives of distributions to manifolds are recorded as not done, which matches the checklist.

## Checks

`scripts/api_map_linter.py --repo .` passes with no missing files and no missing names, resolving 27 declarations against the Lean sources.

### DIFF
--- a/Physlib/SpaceAndTime/SpaceTime/API-map.yaml
+++ b/Physlib/SpaceAndTime/SpaceTime/API-map.yaml
@@ -1,0 +1,62 @@
+version: v0.1
+
+Title: SpaceTime
+
+Overview: |
+    The key data structure is `SpaceTime d`, an abbreviation for `Lorentz.Vector d`,
+    Minkowski space-time with an arbitrary but fixed choice of units and origin, for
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - looks good to me. Thanks.

PR #1471, opened 2026-07-29
feat(Mathematics): ContDiff and HasTemperateGrowth for `physHermite`

gloges:
- Organizes the first half of `Physlib/Mathematics/SpecialFunctions/PhysHermite.lean` into sections and adds some basic documentation.
- Adds `ContDiff` and `HasTemperateGrowth` properties for the coercions to `ℝ → ℝ`; these will be needed when constructing the QHO eigenstates using `SchwartzMap.smulLeftCLM`.


jstoobysmith (approved):
All looks good to me - approved.

PR #1472, opened 2026-07-30
refactor: remove erw from wave equation proofs

Robertboy18:
I wanted to start with a small PR, so I cleaned up the three remaining uses of  `erw` in the wave equation file.

  For the first two, I used the existing `iteratedDeriv_succ` lemma directly.
  For the last one, I kept the expression in `planeWave` form so the spatial and
  time derivative lemmas work with ordinary `rw`. Everything passes and the linter too!

  Contributes to #385.


jstoobysmith (approved):
This looks great! Approved - many thanks for this first PR :). Will merge once all the linters finish

PR #1473, opened 2026-07-31
feat: API map index generator and guide

Robby955:
This adds a small index generator for the API maps: `lake exe api_map_index` reads the `API-map.yaml` files and writes `docs/APIMap.html` in the doc-site layout, so the existing maps are browsable in one place. A short guide for writing map entries is included at `docs/API_MAP_GUIDE.md`. It complements the per-area maps being added under #1414 and picks up new ones automatically. Nothing in the repo invokes the generator yet: the site build would add `lake exe api_map_index` next to its existing `lake exe stats mkHTML` step to publish the page.

### DIFF
--- a/scripts/APIMapIndex.lean
+++ b/scripts/APIMapIndex.lean
@@ -1,0 +1,302 @@
+/-
+Copyright (c) 2026 Robert Sneiderman. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Robert Sneiderman
+-/
+/-!
+# API map index
+
+This executable gathers every `API-map.yaml` in the library into a single
+YAML data file, `docs/_data/APIMap.yml`, for the website to render. It follows
+the pattern of `scripts/MetaPrograms/TODO_to_yml.lean`, which builds
+`docs/_data/TODO.yml` for the website's TODO list: the YAML is printed to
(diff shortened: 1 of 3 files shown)


scripts/APIMapIndex.lean
jstoobysmith, line 1: this all looks really good. But I think we should think about whether html is the correct output for this. 

Maybe if we output a json or a (large) yaml file containing all of the necessary information, on the website rep we can include a page which displays that json or yaml in nice ways. I've found this approach often works better (and we should likely do that for the stats page aswell). I think the TODO list is done in this was as an example. 

(I also think the html outputted would no longer be compatible with the format of the website).

Robby955: I'll take a look!

Robby955:
Reworked as suggested: the HTML output is gone, and the script now emits a single YAML data file, docs/_data/APIMap.yml, mirroring the repo's existing data-file exporter pattern with the same mkFile convention and uncommitted artifact. It carries each map's path, title, overview, parents, references and requirements with completion counts and GitHub links, so the website repo can render it the same way as the task list.

Robby955:
-awaiting-author

jstoobysmith (approved):
Great! I've approved this. Will merge shortly

PR #1474, opened 2026-07-31
EVMPR theorem twoState_helmholtzFreeEnergy_eq 

korolrom:
This is my first contribution to physlib. 
The result can be simplified further under hypothesis T \neq 0, but I left it as a more general result including zero temperature.

### DIFF
--- a/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
+++ b/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
@@ -116,7 +116,8 @@ lemma twoState_helmholtzFreeEnergy_eq (E₀ E₁ : ℝ) (T : Temperature) :
   have hE1 : -β T * E₁ = -x + (- C) := by
     simp [x, C]
     ring
-  rw [helmholtzFreeEnergy, twoState_partitionFunction_apply,show (T.val : ℝ) = T.toReal by rfl,hE0, hE1]
+  rw [helmholtzFreeEnergy, twoState_partitionFunction_apply,
+    show (T.val : ℝ) = T.toReal by rfl,hE0, hE1]
   have hfactor :
       Real.exp (x + (- C)) + Real.exp (-x + (- C)) = Real.exp (-C) * (2 * Real.cosh x) := by
     rw [Real.exp_add, Real.exp_add, Real.cosh_eq]
--- a/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
+++ b/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
@@ -104,8 +104,7 @@ informal_lemma twoState_entropy_eq where
   deps := [``twoState, ``thermodynamicEntropy]
 
 /-- A simplification of the `helmholtzFreeEnergy` of the two-state canonical ensemble. -/
-lemma twoState_helmholtzFreeEnergy_eq
-    (E₀ E₁ : ℝ) (T : Temperature) :
+lemma twoState_helmholtzFreeEnergy_eq (E₀ E₁ : ℝ) (T : Temperature) :
     (twoState E₀ E₁).helmholtzFreeEnergy T =
       (β T  * (E₀ + E₁) / 2 - Real.log
           (2 * Real.cosh (β T * (E₁ - E₀) / 2))) / β T  := by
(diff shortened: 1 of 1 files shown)


Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
```diff
@@ -104,8 +104,33 @@ informal_lemma twoState_entropy_eq where
   deps := [``twoState, ``thermodynamicEntropy]
 
 /-- A simplification of the `helmholtzFreeEnergy` of the two-state canonical ensemble. -/
-informal_lemma twoState_helmholtzFreeEnergy_eq where
-  tag := "EVMPR"
-  deps := [``twoState]
+lemma twoState_helmholtzFreeEnergy_eq
+    (E₀ E₁ : ℝ) (T : Temperature) :
```
jstoobysmith: Could probably collapse this to one line.

Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
```diff
@@ -104,8 +104,33 @@ informal_lemma twoState_entropy_eq where
...
+  set C := β T * (E₀ + E₁) / 2
+  have hE0 : -β T * E₀ = x +(- C) := by
+    simp [x, C]
+    ring
+  have hE1 : -β T * E₁ = -x + (- C) := by
+    simp [x, C]
+    ring
+  rw [helmholtzFreeEnergy, twoState_partitionFunction_apply]
+  rw [show (T.val : ℝ) = T.toReal by rfl]
+  rw [hE0, hE1]
```
jstoobysmith: May worth be putting this into a single large `rw [...]`

Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
```diff
@@ -104,8 +104,33 @@ informal_lemma twoState_entropy_eq where
...
+      Real.exp (x + (- C)) + Real.exp (-x + (- C)) =
+        Real.exp (-C) * (2 * Real.cosh x) := by
+    rw [Real.exp_add, Real.exp_add, Real.cosh_eq]
+    ring
+  rw [hfactor]
+  rw [Real.log_mul
+        (Real.exp_pos _).ne'
+        (by positivity : 2 * Real.cosh x ≠ 0)]
+  rw [Real.log_exp,Temperature.β_toReal]
+  simp only [div_eq_mul_inv, one_mul, inv_inv]
```
jstoobysmith: Could you combine the things in the `rw` with the things in the `simp` so e.g. 

```
simp only [Real.log_exp,Temperature.β_toReal, div_eq_mul_inv, one_mul, inv_inv]
```
(Will have to check if this actually works).

jstoobysmith (commented):
Hi @korolrom This is a great first PR. I've added some comments (some of them are a bit pedantic). If you could have a go at addressing them, this would be great. 

Let me know if you have any questions :).

jstoobysmith:
Also doing the T \neq 0 case would be a great future PR!

korolrom:
Thank you for the ultrafast review! @jstoobysmith 
I have condensed the lines you suggested as well as 1-2 more.

I am happy to also add the T \neq 0 case. Would that be another lemma? Feels like not "big" enough to warrant its own lemma :)

jstoobysmith:
This looks much better - and looks almost there! Some of the linters are now failing though. Would you mind fixing these (some instructions to running them locally [here](https://github.com/leanprover-community/physlib/blob/master/scripts/README.md), but can also just rely on github workflows. The ones I can see now are about line lengths. 

Regarding T \neq 0, if it is a large simplification then it probably does warrant its own lemma, but probably best to keep that to a separate Pull request (PR).

korolrom:
Ok, I split the line and I tried to run the linter locally but there are a lot of errors and warnings from 
`lake exe lint all`

I am ready to create another PR for the T\neq 0 result once this one is merged!

jstoobysmith (approved):
This looks good now. Yeah some of the warnings on all the linters are not vital. 

Many thanks for this first pr! Will merge shortly

jstoobysmith:
To be a bit more specific about the linters: `lake exe lint_all` has two types of linters: Those which must be done for the github ci and those which are optional (which are hopefully clearly labelled) 
Because they are optional, `lake exe lint_all` pulls up things which others haven't fixed in earlier PRs, and you shouldn't feel any expectation to fix these.

jstoobysmith:
Merged - looking forward to future PRs 🎉

Issue #457, opened 2025-04-07
Formalization: Derivatives and SpaceTime


jstoobysmith:
This issue originates from [this](https://leanprover.zulipchat.com/#narrow/channel/479953-PhysLean/topic/Maxwell's.20Equations/with/510454417) conversation.

- [ ] Define the epsilon tensor in general dimension. See [here](https://en.wikipedia.org/wiki/Levi-Civita_symbol).
- [ ] Redefine the `spaceCurl` in terms of this epsilon tensor.
- [x] Define a continuous linear equivalence between `SpaceTime` and `Time x Space`.
- [x] Define a map `timeSlice` taking `SpaceTime → M` to `Time  → Space  → M` with corresponding API.
- [x] Define `((realLorentzTensor d).tensorBasis _ (fun x => Fin.cast (by simp) μ))` formally as a unit vector in SpaceTime pointing the μ direction.
- [ ]  Change `SpaceTime.deriv` to an abbreviation of `fderiv` using the above spacetime derivative.
- [ ] Come up with a good way to go from `SpaceTime` to `Space` derivatives etc. (e.g. greek derivatives vs latin)
- [x] Use the above to rewrite Maxwell's equations in a better way.

For more details on individual points see the details of the above conversation.

PR #1476, opened 2026-08-01
feat(Mathematics): Levi-Civita symbol in general dimension

Robby955:
This adds the Levi-Civita symbol in general dimension, with the basic lemmas: antisymmetry under swaps, vanishing exactly on repeated indices, the value on the identity, and agreement with the rank-4 Lorentz tensor in the existing relativity file. A lemma expresses the components of curl as a contraction with the symbol, leaving the definition of curl unchanged. This covers the epsilon items of #457; the remaining boxes there stay open.

### DIFF
--- a/Physlib/Mathematics/LeviCivita/Basic.lean
+++ b/Physlib/Mathematics/LeviCivita/Basic.lean
@@ -30,6 +30,7 @@ permutation via `Matrix.det_permutation`.
 - `leviCivitaSymbol_id` : the normalization `ε_{0 1 ⋯ (d-1)} = 1`.
 - `leviCivitaSymbol_perm` : on a permutation `σ` the symbol is the sign of `σ`.
 - `leviCivitaSymbol_comp_swap` : antisymmetry under transposition of two indices.
+- `leviCivitaSymbol_swap_comp` : antisymmetry under transposition of two index values.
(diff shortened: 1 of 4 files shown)


Physlib/Mathematics/LeviCivita/Basic.lean
```diff
@@ -0,0 +1,139 @@
...
+## C. Antisymmetry
+
+-/
+
+/-- The Levi-Civita symbol is antisymmetric under transposition of two of its indices:
+precomposing with the swap of two distinct index positions negates it. -/
+lemma leviCivitaSymbol_comp_swap (g : ι → ι) {i j : ι} (hij : i ≠ j) :
+    leviCivitaSymbol (g ∘ Equiv.swap i j) = - leviCivitaSymbol g :=
+  generalizedKroneckerDelta_swap g id hij
+
```
nateabr, line 111: Maybe we should add a companion for this lemma above for post composition

nateabr (commented):
This looks good to me, only one minor comment.

nateabr:
awaiting-author

Robby955:
-awaiting-author

jstoobysmith (approved):
Approving this, and will merge shortly. Many thanks

PR #1477, opened 2026-08-01
fix(linters): re-enable defsWithUnderscore with narrow exemptions

Robby955:
This re-enables the defsWithUnderscore linter in both runner scripts. False positives are handled by four narrow type-level exemptions in a new Physlib/Meta/Linters module, wrapping the mathlib linter test since its own exemptions are hardcoded and cannot be extended downstream: informal declarations, parser and pretty-printer declarations, project-suffixed instance names, and numeric suffixes. The 23 genuine violations are renamed to the usual conventions, with companion lemmas and the two doc references updated; a repo-wide sweep including the API map files finds no stale names.


jstoobysmith (approved):
This is great! and very useful! I've approved this, will merge shortly.

PR #1479, opened 2026-08-01
auto-task(energyEigenstate_orthogonal): golf the tight-binding chain orthogonality proof

jakeuribe:
## Summary

Golfs the proof of `TightBindingChain.energyEigenstate_orthogonal` in
`Physlib/CondensedMatter/TightBindingChain/Basic.lean` from 57 lines to 32.
The statement is unchanged.

## What changed

- **Term lemma instead of a sum rewrite.** The old proof stated `hsum_eq`, a full
  equality of sums, and re-typed the whole summand inside
  `Fin.sum_univ_eq_sum_range (fun n => ...)`. It is replaced by a pointwise
  `hterm (n : ℕ) : conj (exp (I * k1 * n * a)) * exp (I * k2 * n * a) = ω ^ n`,
  which `simp only` applies under the binder, so the index-set conversion becomes
  the one-liner `Fin.sum_univ_eq_sum_range (ω ^ ·)`.
- **`geom_sum_eq` instead of `mul_geom_sum`.** The final three steps
  (`mul_geom_sum` → `rw` → `mul_eq_zero.mp ... |>.resolve_left ...`) collapse into
  `geom_sum_eq hω_ne_one, hω_pow, sub_self, zero_div`, folded into the closing `rw`.
- **`hω_pow` as a single `rw` chain.** The `calc` block with two `ring_nf` calls and
  two auxiliary `have`s becomes one `rw` using a `show ... by push_cast; ring`
  bridge and `T.quantaWaveNumber_exp_N` applied directly.
- **Divisibility instead of `nlinarith`.** `m = 0` was derived by bounding
  `T.N * m` and calling `nlinarith`. It is now
  `Int.eq_zero_of_abs_lt_dvd` on `(T.N : ℤ) ∣ (n2 : ℤ) - n1`, removing the
  `nlinarith` call and three cast `have`s.
- Smaller cleanups: `hω_ne_one` is a term-mode `fun ... => hne <| by ...`;
  `starRingEnd_apply`/`Complex.star_def` are dropped in favour of rewriting
  with `← Complex.exp_conj` directly; the hand-written `ofReal_*` `simp only`
  list becomes `push_cast`; `(T.N : ℂ) ≠ 0` is a term
  (`Nat.cast_ne_zero.mpr (NeZero.ne T.N)`) rather than a `simp` call.

The remaining comments were rewritten to sit above the three structural steps
(per-term root of unity → `ω ^ N = 1` → `ω ≠ 1` → geometric sum vanishes).

## Verification

`lake build` completes successfully (9264 jobs), with no new warnings, `sorry`s,
or axioms.

### DIFF
--- a/Physlib/CondensedMatter/TightBindingChain/Basic.lean
+++ b/Physlib/CondensedMatter/TightBindingChain/Basic.lean
@@ -418,62 +418,37 @@ N-th roots of unity exp(i(k₂-k₁)a), and the sum of all N-th roots of unity e
 lemma energyEigenstate_orthogonal :
     Pairwise fun k1 k2 => ⟪T.energyEigenstate k1, T.energyEigenstate k2⟫_ℂ = 0 := by
   intro k1 k2 hne
-  simp only [energyEigenstate, sum_inner]
-  simp_rw [inner_sum, inner_smul_left, inner_smul_right, localizedState_orthonormal_eq_ite]
-  simp only [mul_ite, mul_one, mul_zero, Finset.sum_ite_eq, Finset.mem_univ, ↓reduceIte]
   set ω := Complex.exp (Complex.I * (k2 - k1) * T.a) with hω_def
-  have hsum_eq : ∑ n : Fin T.N, (starRingEnd ℂ) (Complex.exp (Complex.I * k1 * n * T.a)) *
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - many thanks. APproved

PR #1489, opened 2026-08-04
refactor: Higgs GaugeGroupI action to a Representation

doxtor6:
## Summary

Resolves the TODO in `HiggsBoson/Basic.lean`: the action of `GaugeGroupI` on `HiggsVec` is
now a directly-defined `Representation ℂ GaugeGroupI HiggsVec` (`repGaugeGroupI`), and the
`SMul`/`MulAction`/`DistribMulAction`/`SMulCommClass` instance stack is removed, so the
representation is the single spelling of the action on `HiggsVec`.

- Lemmas restated in rep form: `repGaugeGroupI_apply` (and the `U1`/`SU2` variants),
  `repGaugeGroupI_inner`/`_norm`, `toRealGroupElem_apply_self`,
  `ofU1Subgroup_repGaugeGroupI_apply`, `repGaugeGroupI_phase_snd`; the orbit lemma becomes
  `exists_repGaugeGroupI_eq_iff_norm_eq` (stated with an explicit existential).
- New group-property lemmas (`repGaugeGroupI_mul_apply`, inverse cancellation simp lemmas,
  `repGaugeGroupI_inv_apply_eq_iff`) replace `mul_smul`/`inv_smul_smul` downstream.
- `EffectivePotential`: `IsInvariant` is stated via the representation; the `smul_comm` step
  is replaced by linearity (`LinearMap.map_smul_of_tower`).
- `TwoHDM`: `TwoHiggsDoublet` keeps its `MulAction` (its orbit API needs it), now defined
  componentwise through `repGaugeGroupI`; `GramMatrix` orbit lemmas ported to the rep form.

Carved out of #1415, where the agreed convention is that gauge actions on field target spaces
are `Representation`s rather than `MulAction` instances; this part is independent of the EFT
algebra work and useful on its own. The corresponding `EFTLagrangianExclDeriv` hookups stay
in #1415, since those files only exist there.

Built locally (the four changed files and their two downstream importers,
`HiggsBoson/Potential` and `TwoHDM/Potential`).

### DIFF
--- a/Physlib/Particles/BeyondTheStandardModel/TwoHDM/Basic.lean
+++ b/Physlib/Particles/BeyondTheStandardModel/TwoHDM/Basic.lean
@@ -59,21 +59,21 @@ lemma ext_of_fst_snd {H1 H2 : TwoHiggsDoublet}
 
 noncomputable instance : SMul StandardModel.GaugeGroupI TwoHiggsDoublet where
   smul g H :=
-    { Φ1 := g • H.Φ1
-      Φ2 := g • H.Φ2 }
+    { Φ1 := StandardModel.HiggsVec.repGaugeGroupI g H.Φ1
+      Φ2 := StandardModel.HiggsVec.repGaugeGroupI g H.Φ2 }
 
 @[simp]
 lemma gaugeGroupI_smul_fst (g : StandardModel.GaugeGroupI) (H : TwoHiggsDoublet) :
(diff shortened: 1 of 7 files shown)


jstoobysmith:
Looks good to me! @nateabr do you also want to take a look?

nateabr:
Yep looks great to me!

nateabr (approved):

PR #1490, opened 2026-08-06
feat(QuantumMechanics) : Prove potentialOperator_isSelfAdjoint

ereinhardt8:
This pull request provides a proof for the lemma `potentialOperator_isSelfAdjoint` in `Physlib/QuantumMechanics/RectangularBarrier/Basic.lean`, which was previously an informal_lemma. Also, it proves `potentialFunction_aestronglyMeasurable` because `potentialOperator_isSelfAdjoint` uses it. 

## Specific Changes

Two informal lemmas were formalized and proved:

- `RectangularBarrier.potentialFunction_aestronglyMeasurable` : formalize and prove that the potential function of a rectangular barrier is almost everywhere strongly measurable 
- `RectangularBarrier.potentialOperator_isSelfAdjoint` : prove that the potential function of a rectangular barrier is self adjoint using the above lemma 

## Reviewer Guide

I fixed any linting issues found in `Physlib/QuantumMechanics/RectangularBarrier/Basic.lean` after I ran `lake exe lint_all` and `./scripts/lint-style.sh` locally. 

The lemma `potentialFunction_aestronglyMeasurable` was only proved for the canonical `volume` measure. Is this an issue? 

## Additional Comments
I wrote the code, but used ChatGPT and Google Search to suggest approaches and next steps. 

Thank you!

### DIFF
--- a/Physlib/QuantumMechanics/RectangularBarrier/Basic.lean
+++ b/Physlib/QuantumMechanics/RectangularBarrier/Basic.lean
@@ -87,10 +87,15 @@ lemma potentialFunction_eq :
     Q.potentialFunction = fun x ↦ (Icc Q.lower Q.upper).indicator (fun _ ↦ Q.V₀) (x 0) := rfl
 
 /-- The piecewise-constant potential of the rectangular barrier is a.e. strongly measurable. -/
-informal_lemma potentialFunction_aestronglyMeasurable where
-  deps := [``RectangularBarrier]
-  tag := "QM-RB-aesm"
-  -- This relies on `Space.val` being measure-preserving.
+-- This relies on `Space.val` being measure-preserving.
+lemma potentialFunction_aestronglyMeasurable: AEStronglyMeasurable Q.potentialFunction volume := by
+  unfold potentialFunction
+  apply AEStronglyMeasurable.indicator
+  · fun_prop
+  · change (MeasurableSet ((Icc Q.lower Q.upper) ∘ (fun (x: Space 1) => x.val 0)))
+    have hi : MeasurableSet (Icc Q.lower Q.upper) := by measurability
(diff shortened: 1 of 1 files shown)


Physlib/QuantumMechanics/RectangularBarrier/Basic.lean
```diff
@@ -119,9 +124,13 @@ def kineticOperator : Q.HS →ₗ.[ℂ] Q.HS := (2 * Q.m)⁻¹ • momentumSqOpe
...
-informal_lemma potentialOperator_isSelfAdjoint where
-  deps := [``RectangularBarrier]
-  tag := "QM-RB-sa"
+lemma potentialOperator_isSelfAdjoint (Q : RectangularBarrier) :
+    IsSelfAdjoint Q.potentialOperator := by
+  unfold IsSelfAdjoint; unfold potentialOperator
+  rw [mulOperator_isSelfAdjoint_ofReal]
+  swap; ext x; simp only [Function.comp_apply, Complex.conj_ofReal]
+  have hQ := potentialFunction_aestronglyMeasurable
+  fun_prop
```
jstoobysmith, line 136: Hey @ereinhardt8 Many thanks for this first PR! Just one comment from me before this can be merged. It would be convention to not to have more then one tactic on a line e.g. `swap; ext x; simp [...]` should be three separate lines. Other then that, this PR looks great!

jstoobysmith:
awaiting-author

ereinhardt8:
Okay! I separated the lines that had multiple tactics. 
-awaiting-author

jstoobysmith (approved):
Great! Many thanks! I've approved, and when the linters finish, I will merge this. So nothing else you need to do here. 

Many thanks for this first PR! Hopefully the first of many.

PR #1491, opened 2026-08-07
docs(tensors): add Tensors API map

Robby955:
## Motivation

Related to #1414, the tracking maps for the tensor API in #908 and the Lorentz tensor API in #886. Both issues are still unfilled templates, so the map supplies the requirement list they ask for. An API map records the implemented status and source location of each requirement, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/Relativity/Tensors/API-map.yaml`, with 22 requirements, 16 of them done. The completed entries cover `TensorSpecies` and the types `Tensor`, `Pure` and `ComponentIdx`; the group action, index permutation, products, contraction of a dual pair, evaluation, and the unit and metric tensors with raising and lowering; the index-notation elaborator and `Tensorial`; conjugation; and the species `realLorentzTensor` and `complexLorentzTensor` with the semilinear map between them.

Six are recorded as not done: neither Lorentz species is extended to a `ConjTensorSpecies`, so reality and Hermiticity cannot yet be stated for them; the metrics and units of the real species are still built from `ContrMod` and `CoMod` rather than as intertwiners of `Vector` and `CoVector`; three commutation lemmas for `evalT` are absent; there is a Euclidean Levi-Civita symbol but no Euclidean tensor; `TensorSpecies` carries no condition relating the metrics to the basis vectors; and the tensor-level epsilon-epsilon identities `leviCivita_contract_self` and `leviCivita_contract_three` are stated but carry `sorry`, so they are recorded the way `solidSphere_inertiaTensor` is in the rigid body map. The componentwise identities `euclidLeviCivita_symbol_contract_zero`, `_one` and `_two` are proved and stay under the done entry. Nothing outside the new file changes, and no Lean source is touched.

## Checks

`python3 scripts/api_map_linter.py --repo .` passes with no missing files and no missing names, resolving 289 declarations against the Lean sources.

### DIFF
--- a/Physlib/Relativity/Tensors/API-map.yaml
+++ b/Physlib/Relativity/Tensors/API-map.yaml
@@ -1,0 +1,134 @@
+version: v0.1
+
+Title: Tensors
+
+Overview: |
+    The key data structure is `TensorSpecies`, which packages everything needed to
+    define a type of tensor with index notation: a symmetry group `G`, a type `C` of
+    index colors, a representation and a basis for each color, an involution `τ`
+    naming the dual color that an index may contract with, and the contraction, unit
+    and metric data for each dual pair. For `S : TensorSpecies` and a list of colors
+    `c : Fin n → C`, `S.Tensor c` is the type of tensors with those indices, realized
+    as a `PiTensorProduct`; `Pure S c` is the type of pure tensors; and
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):

Issue #882, opened 2026-01-05
API: Feynman diagrams


jstoobysmith:
## Key data structure 

The key data structure here is the general notation of a Feynman diagram. It is suspected that getting a good, and general definition for this data structure will be hard, but vital for the building of a good API. 

- [ ] is defined 

## Need

Feynman diagrams are one of the main calculation tools used in particle physics. With Feynman diagrams we will be able to formalise a lot more results within particle physics than is currently possible, and slowly make our way to research level results in this direction.

## Requirements

- [ ] The definition shall be general enough to accept any Feynman diagram from any theory.
- [ ] The API shall contain a definition to provably generate all Feynman diagrams up to a given order for a given amplitude.
- [ ] The API shall contain a computable definition of the symmetry factor of a Feynman diagram. 
- [ ] The API shall contain a definition permitting the conversion of a Feynman diagram into its underlying algebraic expression.

## Corresponding file system

There is a start to this in PhysLean, however this can likely be rewritten.

## Parent APIs

#881 #885

jstoobysmith:
Closing as included in code as API-map, this should be followed instead of this PR.

PR #1492, opened 2026-08-07
docs(qft): add Feynman diagram and topological field theory API maps

Robby955:
## Motivation

Related to #1414, the tracking maps for the Feynman diagram API in #882 and the topological field theory API in #877. Following the discussion in #1414, an API with no implementation yet still gets a map in the file system with every requirement marked not done; these are two of those.

## Changes

Adds `Physlib/QFT/PerturbationTheory/FeynmanDiagrams/API-map.yaml` and `Physlib/QFT/TopologicalFieldTheory/API-map.yaml`. Every requirement in both is recorded as not done with location `N/A`, 6 in the first and 17 in the second, so each map is specification only. The Feynman diagram directory holds `Basic.lean`, which states the intended design in its module documentation and declares nothing, so the map writes that design out as requirements and names the perturbation-theory APIs a diagram type would sit on. `Physlib/QFT/TopologicalFieldTheory` has no Lean file at all and now contains only the map; `check_file_imports` walks `.lean` files alone, so the import check is unaffected. Nothing outside the two new files changes, and no Lean source is touched.

`FeynmanDiagrams/Basic.lean` asks that you be contacted before anyone works in that directory. I have added only the map there and changed no Lean file, but say the word and I will drop that half and keep the topological field theory map alone.

## Checks

`python3 scripts/api_map_linter.py --repo .` passes with no missing files and no missing names. Both new maps report zero checked entries, since no requirement claims a location.

### DIFF
--- a/Physlib/QFT/PerturbationTheory/FeynmanDiagrams/API-map.yaml
+++ b/Physlib/QFT/PerturbationTheory/FeynmanDiagrams/API-map.yaml
@@ -1,0 +1,68 @@
+version: v0.1
+
+Title: "Feynman diagrams"
+
+Overview: |
+  The key data structure is a Feynman diagram, understood here as a partition of a
+  multiset of field operators into disjoint pairs. This is the multiset counterpart of
+  a Wick contraction, which is a finite set of disjoint pairs of positions in a list
+  of field operators; the intended relation is that the permissible Wick contractions,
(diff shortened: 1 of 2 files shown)


jstoobysmith (approved):
Approved - looks good to me.

PR #1499, opened 2026-08-11
docs(higgs): add Higgs field API map

Robby955:
## Motivation

Related to #1414 and the tracking issue #857, which is still an unfilled template, so the map supplies the requirement list it asks for. The Two Higgs Doublet Model map already lists Higgs Field (`Physlib/Particles/StandardModel/HiggsBoson`) as a parent API, so this map also fills a node that an existing map references. An API map records the implemented status and source location of each requirement, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/Particles/StandardModel/HiggsBoson/API-map.yaml`, with 21 requirements, 16 of them done. The completed entries cover `HiggsVec` with its basis, real-scalar coordinates, and the `repGaugeGroupI` representation with its invariance and orbit-by-norm theory; `HiggsBundle` and `HiggsField` with constant fields and component smoothness; the pointwise inner product and norm-squared with their algebra and smoothness; the `Potential` structure with its completing-the-square, discriminant, boundedness, and minimisation theory; and `EffectivePotential` with gauge invariance, factorisation through the norm, and the polynomial expansion in powers of the norm.

Five are recorded as not done: the stability-group lemmas and the global gauge action on `HiggsField` exist only as informal declarations, and the map cites them as such; the Higgs bundle is still the trivial bundle rather than an associated bundle; effective potentials are not yet connected to field-level potentials; and covariant derivatives with kinetic terms wait on a gauge-connection API. Nothing outside the new file changes, and no Lean source is touched.

## Checks

`python3 scripts/api_map_linter.py --repo .` passes with no missing files and no missing names, resolving 91 declarations in the new map against the Lean sources.

### DIFF
--- a/Physlib/Particles/StandardModel/HiggsBoson/API-map.yaml
+++ b/Physlib/Particles/StandardModel/HiggsBoson/API-map.yaml
@@ -1,0 +1,151 @@
+version: v0.1
+
+Title: Higgs Field
+
+Overview: |
+  The key data structure is `HiggsField`, the smooth sections of the Higgs bundle
+  `HiggsBundle` over `SpaceTime`. The bundle is currently the trivial bundle with
+  fiber `HiggsVec`, the two-dimensional complex vector space of pointwise Higgs
+  values. On `HiggsVec` the API provides an orthonormal basis, real-scalar
+  coordinates, the representation of the global Standard Model gauge group, its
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - looks good to me. Thanks

Issue #858, opened 2025-12-29
API: Standard Model gauge group


jstoobysmith:
# Standard model gauge group 

## Key data structure

The Standard Model gauge group is actually one of a number gauge groups corresponding to a quotient of SU(3)xSU(2)xU(1) by a finite group. 
The key data structure should take as an input a specification of which quotient, and correspond to said group.

- [ ] Is defined
	
## Need

The standard model gauge group is the most important groups in particle physics, and having a good API around it will open the door for lots of further results in this area.

## Requirements

The following are requirements of the API:

- [ ] the instance of a group
- [ ] the instance of a Lie group
- [ ] the center of the group
- [ ] the principal bundle
- [ ] associated bundles
- [ ] common representations, and properties thereof
- [ ] the electroweak subgroup
- [ ] connection to an API around the Lie algebra
	
## Corresponding File System 

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/Particles/StandardModel

## Contributing to this API

This API is open to contributions from anyone. If you believe that the requirements outlined above need modification, feel free to comment here. If you make a PR which is related to this API, please link it to this issue.

## Parent APIs

#859 #856

PR #1500, opened 2026-08-11
docs(standardmodel): add Standard Model gauge group API map

Robby955:
## Motivation

Related to #1414 and the tracking issue #858, which asks for a gauge group parameterised by a choice of finite quotient. That structure now exists in the tree (`GaugeGroupQuot` and `GaugeGroup`, from the gauge-group quotient API), so the map records the requirement list against it. The Higgs Field map lists the Standard Model gauge group (`Physlib/Particles/StandardModel`) as a parent API, so this also fills a referenced node.

## Changes

Adds `Physlib/Particles/StandardModel/API-map.yaml`, with 15 requirements, 10 of them done. The completed entries cover `GaugeGroupI` with its factor projections and star structure; the embeddings of the sixth, second and third roots of unity as central subgroups with membership, centrality and normality; the three quotient groups; the quotient-choice type `GaugeGroupQuot` with `GaugeGroup`, `subgroup` and `quotientMap` given uniformly in the choice, including the kernel characterisation of the quotient map; and the U(1) and fundamental SU(2) representations on the two-dimensional space of the Higgs sector.

Five are recorded as not done: the Lie group structure and the gauge bundle with gauge transformations exist only as informal declarations, and the map cites them as such; the center of the full group is bounded below by the central subgroups but not characterised exactly; and the unbroken gauge group from the Higgs field and the fermionic-content structure are both open tasks already marked in the source. Nothing outside the new file changes, and no Lean source is touched.

## Checks

`python3 scripts/api_map_linter.py --repo .` passes with no missing files and no missing names, resolving 62 declarations in the new map against the Lean sources.

### DIFF
--- a/Physlib/Particles/StandardModel/API-map.yaml
+++ b/Physlib/Particles/StandardModel/API-map.yaml
@@ -1,0 +1,113 @@
+version: v0.1
+
+Title: Standard Model gauge group
+
+Overview: |
+  The key data structures are `GaugeGroupI`, the full gauge group
+  SU(3) x SU(2) x U(1), and `GaugeGroup q` for `q : GaugeGroupQuot`, the gauge
+  group corresponding to a choice of quotient by a finite central subgroup
+  (ℤ₆, ℤ₂, ℤ₃, or none). The API provides the projections onto the three
+  factors, the star structure, the embeddings of the roots of unity as central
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved.

Issue #848, opened 2025-12-28
API: Finite dimensional Quantum mechanics system


jstoobysmith:
# Finite dimensional Hilbert Space 

## Key data structure

The key data structure is the finite dimensional Hilbert space, which takes as an input the dimension of the Hilbert space. This is for a general system and not for any specific Hamiltonian.

- [x] is defined  

## Need

Finite dimensional quantum mechanical systems are prevalent in lots of aspects of physics, including qubits in quantum information, and the Hilbert space of finite lattices. For example the tight binding chain.

## Requirements

The following are requirements of the API:

- [ ] ability to easily define notation for basis vectors
- [ ] statements about the dimensions 
- [ ] used in the definition of the qubit
- [ ] used in the definition of the tight binding chain 
- [ ] easy conversion between bras and kets
- [ ] tensor products 
- [ ] pure states 
	
## Corresponding File System

This file system exists, the link needs adding here. 
 
## Contributing to this API

This API is open to contributions from anyone. If you believe that the requirements outlined above need modification, feel free to comment here. If you make a PR which is related to this API, please link it to this issue.

PR #1501, opened 2026-08-11
docs(quantum-info): add qubit API map

Robby955:
## Motivation

Related to #1414, the tracking map for the qubit API in #850. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `QuantumInfo/States/Pure/API-map.yaml`, 25 requirements of which 11 are done. The done entries cover `Qubit` itself, the gates `X`, `Y`, `Z`, `H`, `S` and `T` with their relations, `controllize` and `CNOT`, states up to a global phase, POVM measurements, and the fidelity and trace distance, together with the finite dimensional constructions the qubit case instantiates: kets and bras, density matrices, unitary evolution and products. Those last are recorded as the ambient theory rather than as results about `Qubit`, since none of them mentions it.

Fourteen requirements are recorded as not done. Twelve have nothing behind them in the library yet: the Bloch vector and the Pauli decomposition of a state, the rotation gates and the decomposition of a general single-qubit unitary, the Born rule, the Bell and singlet-triplet bases, a closed form for the fidelity of two qubit states, the positive partial transpose criterion, registers of several qubits with universality of `H`, `T` and `CNOT`, and the link from kets on `Qubit` to the finite dimensional Hilbert space of a two element target. The other two are the Bloch sphere and the Bargmann invariant. Those are written and proved, in `BlochSphere.lean` and `BargmannInvariant.lean`, but `QuantumInfo.lean` imports neither, so nothing builds them and I did not want to record them as done; you may want to look at that separately from this map.

Two things I am unsure about. This is the first API map inside `QuantumInfo` rather than under `Physlib`, and the map is titled `Qubits` while it sits at `QuantumInfo/States/Pure`, whose largest module is the general `Braket.lean`. Tell me if you would rather it lived under `Physlib/QuantumMechanics`, or were split along the lines of #850 and #848.

## Checks

`python3 scripts/api_map_linter.py --repo .` passes with no missing files and no missing names, resolving 74 declarations against the Lean sources, and no Lean source is touched.

### DIFF
--- a/Physlib/QuantumMechanics/Qubit/API-map.yaml
+++ b/Physlib/QuantumMechanics/Qubit/API-map.yaml
@@ -7,19 +7,18 @@ Overview: |
     basis of a two-level quantum system. The content specific to qubits is the standard
     single-qubit gates, the controlled version of a unitary with a qubit as control
     register, and the Bloch sphere with its angular parameterization. Qubit states are
-    not built separately: a pure state of a qubit is a ket on `Qubit`, and a general
-    state is a density matrix on `Qubit`. Requirements below that are phrased for a
-    finite dimensional system record constructions defined for an arbitrary finite index
-    type, including the kets, bras and tensor products of this directory. They are
-    recorded here as the ambient theory in which the qubit case sits, obtained by taking
-    the index type to be `Qubit`, and are not claimed as results about `Qubit`. Two
-    modules of this directory, `BlochSphere.lean` and `BargmannInvariant.lean`, are not
+    not built separately: a pure state of a qubit is a ket on `Qubit`, defined for an
+    arbitrary finite index type in `QuantumInfo/States/Pure/Braket.lean`, and a general
+    state is a density matrix on `Qubit`. The ambient theory of a finite dimensional
+    system, in particular density matrices with their evolution, measurements and
+    distinguishability measures, is left to a separate API map. Two modules of the
+    source directory, `BlochSphere.lean` and `BargmannInvariant.lean`, are not
     imported by `QuantumInfo.lean` and so are not built with the library; the two
     requirements that point at them record that.
 
 ParentAPIs:
+  - "Kets and bras of a finite dimensional system (QuantumInfo/States/Pure)"
   - "Mixed states (QuantumInfo/States/Mixed)"
-  - "Unitary operators on quantum states (QuantumInfo/Operators)"
 
 References:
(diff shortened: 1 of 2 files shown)


Physlib/QuantumMechanics/Qubit/API-map.yaml
jstoobysmith, line 1: Could we move this to ./Physlib/QuantumMechanics/Qubit/API-map.yaml, as the long-term plan is to move the QuantumInfo stuff from ./QuantumInfo to ./Physlib eventually.

QuantumInfo/States/Pure/API-map.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+  - description: "The key data structure `Qubit`, the two element type labelling the computational basis of a two-level quantum system, is defined."
+    done: true
+    location: "QuantumInfo/States/Pure/Qubit.lean (Qubit)"
+
+  - description: "Kets and bras of a finite dimensional system are defined, carrying coercions to functions and to each other, and the API contains the bra-ket pairing, the normalization condition in both its componentwise and its inner product form, the computational basis states and the uniform superposition."
+    done: true
+    location: "QuantumInfo/States/Pure/Braket.lean (Ket, Bra, FunLike (Ket d) d ℂ, FunLike (Bra d) d ℂ, Coe (Ket d) (Bra d), Coe (Bra d) (Ket d), FunLike (Bra d) (Ket d) ℂ, dot, Ket.normalized, Ket.basis, uniform_superposition, Braket.dot_self_eq_one)"
+
+  - description: "Density matrices of a finite dimensional system are defined, carrying a coercion to Hermitian matrices and a `Prob` valued inner product of two states, and the API contains expectation values, the spectrum, the maximally mixed state, the embedding of pure states, and the purity together with its characterization of the pure states."
```
jstoobysmith: Maybe we should split this up a bit, as I think `Density matrices` should likely sit in another API map

Robby955:
Moved to Physlib/QuantumMechanics/Qubit/API-map.yaml, and split the content: the map now keeps the qubit-scoped and pure-state entries only. The density matrix, evolution, measurement and distinguishability entries are gone from it and can get their own map under #848.

jstoobysmith (approved):
Approved - looks good. Will merge shortly

PR #1510, opened 2026-08-11
Stone theorem for norm-continuous unitary one-parameter groups

TomOleDiem:
## Summary

This PR proves the bounded-generator form of Stone theorem for norm-continuous unitary one-parameter groups.

- `OneParameterSubgroup.apply_eq_exp_smul_deriv` proves the underlying Banach-algebra result for a continuous `AddChar`.
- `UnitaryOneParameterGroup` bundles an `AddChar` into the Mathlib unitary subgroup and coerces directly to bounded operators. Its canonical generator is self-adjoint, satisfies `U(t) = exp(-itA)`, and is unique. The converse construction and `stoneEquiv` give the correspondence.
- `FiniteTarget.unitaryTimeEvolution` specializes the general result to time translations. Its generator is `H/ℏ`, giving `U(t) = exp(-itH/ℏ)`.

The general unitary-group result is not specific to Hamiltonians: the same structure covers bounded generators of translations and rotations. The Hamiltonian interpretation therefore stays in `FiniteTarget`, while the reusable correspondence lives under `Mathematics/OneParameterSubgroups`.

This is the norm-continuous/bounded case. The strongly continuous theorem with unbounded generators requires spectral theory that is not yet available in Physlib.


jstoobysmith:
Many thanks for this PR. 
Just some quick comments which I think need addressing before merging or getting into the details: 
1. In Physlib we use `lemma` instead of `theorem`. 
2. Might be worth running `min_import` at the bottom of each file to reduce the number of imports. 
3. Worth checking for definitions which are simple rephrasing of things in mathlib e.g. `oneParameterNormedRing`. 
4. Worth checking how much of this one-parameter subgroup stuff can be taken from Mathlib, rather then needing to include it here.

jstoobysmith:
awaiting-author

jstoobysmith:
The final result here is really great work. Though, I do have a couple of concerns that still make it a bit hard to review in its current form:

1. Connection to existing quantum mechanics API: This doesn't currently link up with the self-adjoint operators that already exist in the codebase. I'd expect this result to build on/connect to those. Could you take a look at how they're structured and see if we can tie this in? This may also mean moving the file you added to the right place.

2. Framing as physics vs. math: Right now this reads as an API around mathematics rather than physics. I think there's a lot of useful physics we could surface here if we frame it a bit differently — could we look at adjusting the naming/structure/organization to make the physical meaning clearer?

TomOleDiem:
Thanks for your patience. I am still getting familiar with Lean and with Physlib’s file and folder structure.

I have reworked `Stone.lean` so the main result is now basis-independent and phrased directly in terms of physics: continuous unitary time evolution corresponds to a unique bounded self-adjoint Hamiltonian through

$$U(t)=\exp(-itH/\hbar).$$

It now works with bounded operators on a Hilbert space and is connected directly to `FiniteTarget.timeEvolution`. The general unbounded version would require spectral theory and strong operator continuity that are not yet available in Physlib.

I have left the two underlying one-parameter subgroup files under `Mathematics`. I’m happy to hear any other ideas for how this result could be integrated more naturally into Physlib.

TomOleDiem:
-awaiting-author

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
jstoobysmith, line 1: This looks much better! Two things I would do: 

1. Rename this file as `UnitaryTimeEvolution`. We typically name things after the definitions not the theorems. 
2. For the theorems in ./OneParameterSubgroup with are general, can they be reframed in this file as specifics-physics theorems related to objects within this file. For example, you have `value` both in here and in `Matrix`. Do we actually need it in matrix or any of the results there?

Physlib/QuantumMechanics/Stone.lean
```diff
@@ -0,0 +1,242 @@
...
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+omit [CompleteSpace H] in
+lemma real_smul_operator_eq_complex_smul (t : ℝ) (A : H →L[ℂ] H) :
+    t • A = (t : ℂ) • A := by
+  ext x
+  rfl
+
+/-- The operator implementing the time evolution at time `t`. -/
+def value (U : UnitaryTimeEvolution H) (t : ℝ) : H →L[ℂ] H := U (.ofAdd t)
```
jstoobysmith: You could make this a CoeFun instance which would help with the notation allowing you to just write `U t`

Physlib/QuantumMechanics/Stone.lean
```diff
@@ -0,0 +1,242 @@
...
+    toMonoidHom := E.codRestrict (unitary (H →L[ℂ] H)) fun t => by
+      apply NormedSpace.exp_mem_unitary_of_mem_skewAdjoint
+      rw [skewAdjoint.mem_iff, star_smul, Complex.star_def, Complex.conj_ofReal, hA]
+      module
+    continuous_toFun := Continuous.subtype_mk (by
+      change Continuous
+        (fun t : Multiplicative ℝ => NormedSpace.exp (((Multiplicative.toAdd t : ℝ) : ℂ) • A))
+      fun_prop) _ }
+
+@[simp] lemma value_ofHamiltonian {hamiltonian : H →L[ℂ] H}
```
jstoobysmith: Would typically write this as 

```
@[simp] 
lemma ...
```
etc.

jstoobysmith:
awaiting-author

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+@[expose] public section
+
+open Filter Topology
+
+noncomputable section
+
+namespace OneParameterSubgroup
+
+variable {E F : Type*} [NormedRing E] [FunLike F (Multiplicative ℝ) Eˣ]
+  [MonoidHomClass F (Multiplicative ℝ) Eˣ] [ContinuousMapClass F (Multiplicative ℝ) Eˣ]
```
gloges: Using `FunLike`, `Multiplicative` and `MonoidHomClass` together like this is a hack to say that `F : ℝ → Eˣ` is a homomorphism which sends addition to multiplication, but this is exactly the purpose of `AddChar ℝ E`!
The starting point should be something like
```suggestion
variable {E : Type*} [NormedRing E] (U : AddChar ℝ E)
```
and you can include `(h : Continuous U)` or `(h' : Differentiable ℝ U)` in lemmas that require it. This makes the `ambientValue` def and first few lemmas unnecessary.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+/-- `NormedSpace.exp`'s API wants a `ℚ`-algebra structure (for the `1/n!` coefficients), which
+isn't automatic from `NormedAlgebra ℝ E` alone; derive it once here rather than at each use site. -/
+local instance : NormedAlgebra ℚ E := .restrictScalars ℚ ℝ E
+
+/-- The unit of `E` obtained by scaling `1` by a nonzero real number. -/
+def scalarUnit (d : ℝ) (hd : d ≠ 0) : Eˣ where
+  val := d • 1
+  inv := d⁻¹ • 1
+  val_inv := by rw [smul_mul_smul_comm, mul_inv_cancel₀ hd, one_smul, one_mul]
+  inv_val := by rw [smul_mul_smul_comm, inv_mul_cancel₀ hd, one_smul, one_mul]
```
gloges: This is only used in one place below and is always coerced to `E`, so can be removed.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+  have hc : 0 < c := by dsimp [c]; positivity
+  have hevent : ∀ᶠ x : ℝ in 𝓝 0, ‖ambientValue U x - 1‖ < c := by
+    have hnorm : Continuous (fun x : ℝ => ‖ambientValue U x - 1‖) :=
+      continuous_norm.comp ((continuous_ambientValue U).sub continuous_const)
+    have hmem : Set.Iio c ∈ 𝓝 ‖ambientValue U 0 - 1‖ := by simpa using Iio_mem_nhds hc
+    exact hnorm.continuousAt hmem
+  rw [Metric.eventually_nhds_iff] at hevent
+  obtain ⟨r, hr, hrU⟩ := hevent
+  let d := r / 2
+  have hd : 0 < d := by dsimp [d]; positivity
```
gloges:
```suggestion
  have hd : 0 < d := by positivity
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+  let q : Eˣ := scalarUnit d hd.ne'
+  have hnear : ‖(∫ x in (0 : ℝ)..d, ambientValue U x) - (q : E)‖ < ‖(↑q⁻¹ : E)‖⁻¹ := by
+    have hqd : (q : E) = d • 1 := rfl
```
gloges:
```suggestion
  have hnear : ‖(∫ x in (0 : ℝ)..d, U x) - d • 1‖ < d := by
```
etc. (see above comment on `scalarUnit`)

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+  have heq : ambientValue U = fun s : ℝ => (F (s + d) - F s) * (↑v⁻¹ : E) := by
+    funext s
+    calc
+      ambientValue U s = (ambientValue U s * (v : E)) * (↑v⁻¹ : E) := by simp
+      _ = (F (s + d) - F s) * (↑v⁻¹ : E) := by rw [hv, htranslate]
+  rw [heq]
+  exact (hR.mul_const (↑v⁻¹ : E)).differentiableAt
+
+lemma exists_generator_of_nontrivial [Nontrivial E] (U : F) :
+    ∃ A : E, ∀ t : ℝ, (U (.ofAdd t) : E) = NormedSpace.exp (t • A) := by
```
gloges: Since `A` is "known" and simple to write down, I think this would be better phrased as
```suggestion
lemma apply_eq_exp_smul_deriv [Nontrivial E] (U : AddChar ℝ E) (t : ℝ) :
    U t = NormedSpace.exp (t • deriv U 0) := by
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,267 @@
...
+  rw [← he, Ring.inverse_unit] at hone
+  calc
+    ambientValue U t = (ambientValue U t * (↑e⁻¹ : E)) * (e : E) := by simp
+    _ = NormedSpace.exp (t • A) := by
+      rw [hone, one_mul, he]
+
+/-- Every continuous one-parameter subgroup of the unit group of a real Banach algebra is the
+exponential of a unique element of the algebra. -/
+lemma existsUnique_generator (U : F) :
+    ∃! A : E, ∀ t : ℝ, (U (.ofAdd t) : E) = NormedSpace.exp (t • A) := by
```
gloges: Similar to the previous lemma, perhaps this?
```suggestion
lemma generator_unique (U : AddChar ℝ E) (A : E) (h : ∀ t : ℝ, U t = NormedSpace.exp (t • A)) :
    A = deriv U 0 := by
```

Physlib/Mathematics/OneParameterSubgroups/Matrix.lean
gloges, line 1: It looks like everything in this file is just a special case of `OneParameterSubgroups/Basic.lean` with `E = Matrix (Fin n) (Fin n) ℂ`. Why is this needed?

Physlib/QuantumMechanics/Stone.lean
```diff
@@ -0,0 +1,242 @@
...
+open Constants
+
+noncomputable section
+
+namespace QuantumMechanics
+
+/-- A continuous unitary time evolution on a complex Hilbert space. -/
+abbrev UnitaryTimeEvolution (H : Type*) [NormedAddCommGroup H] [InnerProductSpace ℂ H]
+    [CompleteSpace H] :=
+  ContinuousMonoidHom (Multiplicative ℝ) (unitary (H →L[ℂ] H))
```
gloges: Same comments here about using `AddChar`. This automatically gives you the `CoeFun` @jstoobysmith suggests below that will make everything read more clearly.

Physlib/QuantumMechanics/Stone.lean
```diff
@@ -0,0 +1,242 @@
...
+
+namespace UnitaryTimeEvolution
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+omit [CompleteSpace H] in
+lemma real_smul_operator_eq_complex_smul (t : ℝ) (A : H →L[ℂ] H) :
+    t • A = (t : ℂ) • A := by
+  ext x
+  rfl
```
gloges: This is a duplicate of `Complex.coe_smul`.

gloges (commented):
Thanks for making headway on Stone's theorem!
There is a lot that needs to be cleaned up to best utilize the existing Mathlib machinery and reduce redundancy. The fundamental issue is that these results should be phrased in terms of `AddChar`, not the `MonoidHomClass/Multiplicative` combo currently being used.

TomOleDiem:
Thanks for all these comments. I have reworked the PR to incorporate the feedback:

- the general result now uses Mathlib’s `AddChar`;
- I removed the redundant matrix file and the definitions duplicating existing Mathlib machinery;
- the physics result now lives in `UnitaryTimeEvolution.lean`;
- `UnitaryTimeEvolution` directly exposes the group laws `U(0) = 1`, `U(s+t) = U(s)U(t)`, and `U(-t) = U(t)⁻¹`;
- the correspondence is phrased in terms of bounded self-adjoint Hamiltonians and integrated with `FiniteTarget.timeEvolution`.

Thanks again for the detailed guidance.

-awaiting-author

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+
+open Filter Topology
+
+noncomputable section
+
+namespace OneParameterSubgroup
+
+variable {E : Type*} [NormedRing E]
+
+variable [NormedAlgebra ℝ E] [CompleteSpace E]
```
gloges:
```suggestion
variable {E : Type*} [NormedRing E] [NormedAlgebra ℝ E] [CompleteSpace E]
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+        _ = c * d := by rw [sub_zero, abs_of_pos hd]
+    have hright : ‖(↑q⁻¹ : E)‖⁻¹ = d * ‖(1 : E)‖⁻¹ := by
+      have hqinv : (↑q⁻¹ : E) = d⁻¹ • (1 : E) := rfl
+      rw [hqinv, norm_smul, Real.norm_eq_abs, abs_inv, abs_of_pos hd]
+      field_simp
+    rw [hright]
+    refine hle.trans_lt ?_
+    dsimp [c]
+    nlinarith [inv_pos.mpr hone]
+  exact ⟨d, hd, (Units.ofNearby q _ hnear).isUnit⟩
```
gloges:
```suggestion
  refine ⟨d, hd, (Units.ofNearby q _ ?_).isUnit⟩
  calc
    _ = ‖(∫ x in 0..d, U x) - ∫ x in 0..d, 1‖ := by
      sorry
    _ = ‖∫ x in 0..d, (U x - 1)‖ := by
      sorry
    _ ≤ c * d := by
      sorry
    _ < ‖q.inv‖⁻¹ := by
      sorry
```
More of a stylistic suggestion, but I think this proof would be easier to follow (and maintain) if you use the value of `d` upfront and then `calc` your way through the calculation rather than the alternating `have/rw`.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+lemma differentiable [Nontrivial E] (U : AddChar ℝ E) (hU : Continuous U) :
+    Differentiable ℝ U := by
+  obtain ⟨d, _, hV⟩ := exists_isUnit_intervalIntegral U hU
+  let V : E := ∫ x in (0 : ℝ)..d, U x
+  let v : Eˣ := hV.unit
+  have hv : (v : E) = V := hV.unit_spec
+  let F : ℝ → E := fun t => ∫ x in (0 : ℝ)..t, U x
+  have hF (t : ℝ) : HasDerivAt F (U t) t :=
+    (hU.integral_hasStrictDerivAt 0 t).hasDerivAt
+  have htranslate (t : ℝ) : U t * V = F (t + d) - F t := by
```
gloges: I think this something like this should be pulled out as a stand-alone lemma:
```lean
lemma mul_intervalIntegral_eq_sub (U : AddChar ℝ E) (s t : ℝ) :
    U s * ∫ x in 0..t, U x = (∫ x in 0..(s+t), U x) - ∫ x in 0..s, U x := by
  &c.
```
or even
```lean
lemma mul_intervalIntegral_eq (U : AddChar ℝ E) (s t : ℝ) :
    U s * ∫ x in 0..t, U x = ∫ x in s..(s+t), U x := by
  &c.
```
I'm not sure which one would be better since they are one step away using `intervalIntegral.integral_add_adjacent_intervals`.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+      _ = (F (s + d) - F s) * (↑v⁻¹ : E) := by rw [hv, htranslate]
+  rw [heq]
+  exact (hR.mul_const (↑v⁻¹ : E)).differentiableAt
+
+lemma apply_eq_exp_smul_deriv (U : AddChar ℝ E) (hU : Continuous U) (t : ℝ) :
+    U t = NormedSpace.exp (t • deriv U 0) := by
+  cases subsingleton_or_nontrivial E with
+  | inl h =>
+      let : Subsingleton E := h
+      exact Subsingleton.elim _ _
```
gloges:
```suggestion
  | inl h => exact Subsingleton.elim _ _
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+      U t = (U t * (↑e⁻¹ : E)) * (e : E) := by simp
+      _ = NormedSpace.exp (t • A) := by
+        rw [hone, one_mul, he]
+
+/-- Any exponential generator of a one-parameter subgroup is its derivative at zero. -/
+lemma generator_unique (U : AddChar ℝ E) (A : E)
+    (h : ∀ t : ℝ, U t = NormedSpace.exp (t • A)) : A = deriv U 0 := by
+  have hfun : U = fun t : ℝ => NormedSpace.exp (t • A) := funext h
+  rw [hfun]
+  exact (by simpa using (hasDerivAt_exp_smul_const A (0 : ℝ)).deriv.symm)
```
gloges:
```suggestion
  simp [funext h, (hasDerivAt_exp_smul_const A (0 : ℝ)).deriv]
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+  have heq : U = fun s : ℝ => (F (s + d) - F s) * (↑v⁻¹ : E) := by
+    funext s
+    calc
+      U s = (U s * (v : E)) * (↑v⁻¹ : E) := by simp
+      _ = (F (s + d) - F s) * (↑v⁻¹ : E) := by rw [hv, htranslate]
+  rw [heq]
+  exact (hR.mul_const (↑v⁻¹ : E)).differentiableAt
+
+lemma apply_eq_exp_smul_deriv (U : AddChar ℝ E) (hU : Continuous U) (t : ℝ) :
+    U t = NormedSpace.exp (t • deriv U 0) := by
```
gloges, line 138: Same comment here about the alternating `have/rw` in the `inr h` block. I suggest reserving `have`s for key facts and/or restructuring the proof using `calc/suffices`. Maybe something like
```lean
  | inr h =>
    let G : ℝ → E := fun t => U t * NormedSpace.exp (t • (-deriv U 0))
    have hG : Differentiable ℝ G := sorry
    suffices deriv G = 0 by
      sorry
    &c.
```
to start?

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+  cases subsingleton_or_nontrivial E with
+  | inl h =>
+      let : Subsingleton E := h
+      exact Subsingleton.elim _ _
+  | inr h =>
+    let : Nontrivial E := h
```
gloges:
```suggestion
```
This line can be deleted with no change! It's just giving `h` another name.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,214 @@
...
+      rw [mul_assoc, hcomm.eq]
+      noncomm_ring
+    have hconst (t : ℝ) : G t = G 0 := is_const_of_deriv_eq_zero hG hGder t 0
+    have hone : U t * NormedSpace.exp (t • (-A)) = 1 := by
+      simpa [G] using hconst t
+    have hneg : t • (-A) = -(t • A) := by module
+    have hinv : NormedSpace.exp (t • (-A)) = Ring.inverse (NormedSpace.exp (t • A)) := by
+      rw [hneg]
+      exact (Ring.inverse_exp (t • A)).symm
+    rw [hinv] at hone
```
gloges: e.g. here's a way to begin reducing the number of stray `have`s:
```suggestion
    have hone : U t * Ring.inverse (NormedSpace.exp (t • A)) = 1 := by
      simpa [G, ← Ring.inverse_exp] using hconst t
```

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
```diff
@@ -0,0 +1,258 @@
...
+namespace UnitaryTimeEvolution
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+instance : CoeFun (UnitaryTimeEvolution H) fun _ => ℝ → unitary (H →L[ℂ] H) :=
+  ⟨fun U => U.toAddChar⟩
+
+/-- Time evolution at time zero is the identity operator. -/
+lemma apply_zero (U : UnitaryTimeEvolution H) : U 0 = 1 :=
+  U.toAddChar.map_zero_eq_one
```
gloges:
```suggestion
/-- Time evolution at time zero is the identity operator. -/
@[simp]
lemma apply_zero (U : UnitaryTimeEvolution H) : U 0 = 1 :=
  U.toAddChar.map_zero_eq_one
```
This is a nice example of a `simp` lemma.

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
```diff
@@ -0,0 +1,258 @@
...
+  U.toAddChar.map_add_eq_mul s t
+
+/-- Backward time evolution is the inverse of forward time evolution. -/
+lemma apply_neg (U : UnitaryTimeEvolution H) (t : ℝ) : U (-t) = (U t)⁻¹ := by
+  exact U.toAddChar.map_neg_eq_inv t
+
+/-- At time zero, evolution leaves every state unchanged. -/
+lemma apply_zero_state (U : UnitaryTimeEvolution H) (state : H) :
+    (U 0 : H →L[ℂ] H) state = state := by
+  simp
```
gloges: I don't think this is needed since its essence is that `U 0` is the identity operator, but this is already covered by `apply_zero` and `(1 : H →L[ℂ] H) state = state` by definition. The hint to me to scrutinize this lemma further is that the proof is just "`by simp`".

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
```diff
@@ -0,0 +1,258 @@
...
+
+/-- Evolving a state for `t` and then for `s` is evolution for the total time `s + t`. -/
+lemma apply_add_state (U : UnitaryTimeEvolution H) (s t : ℝ) (state : H) :
+    (U (s + t) : H →L[ℂ] H) state =
+      (U s : H →L[ℂ] H) ((U t : H →L[ℂ] H) state) := by
+  rw [U.apply_add]
+  rfl
```
gloges: Same here.

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
gloges, line 1: I don't think we should limit ourselves to only time evolution/Hamiltonians since everything in this file applies equally well, for example, to translation/momenta.

Also, throughout the file you are casting `U t` to type `H →L[ℂ] H`. This suggests to me that the `CoeFun` needs to be changed or the definition adjusted to something like
```lean
structure UnitaryFoo (H : Type*) [NormedAddCommGroup H] [InnerProductSpace ℂ H]
    [CompleteSpace H] where
  /-- The additive character sending time to its unitary evolution operator. -/
  toAddChar : AddChar ℝ (H →L[ℂ] H)
  /-- foo -/
  selfAdjoint : ∀ t, IsSelfAdjoint (toAddChar t)
  /-- Time evolution is continuous in the operator norm. -/
  continuous : Continuous toAddChar
```

Physlib/QuantumMechanics/UnitaryTimeEvolution.lean
```diff
@@ -0,0 +1,258 @@
...
+
+@[ext]
+lemma ext {U V : UnitaryTimeEvolution H} (h : ∀ t, U t = V t) : U = V := by
+  cases U
+  cases V
+  congr
+  exact AddChar.ext _ _ h
+
+lemma star_apply_neg (U : UnitaryTimeEvolution H) (t : ℝ) :
+    star (U (-t) : H →L[ℂ] H) = U t := by
```
gloges:
```suggestion
@[simp]
lemma adjoint_eq (U : UnitaryTimeEvolution H) (t : ℝ) :
    ContinuousLinearMap.adjoint (U t) = U (-t) := by
```
`star := adjoint` for these maps. By moving the negative to the other side I think this makes for a nice `simp` lemma.

gloges (commented):
Thanks, things look much better now with `AddChar`!
I've made another pass through making comments/suggests - at this point those for `OneParameterSubgroups` are more stylistic and should help to improve the proofs (this will make life easier in the future).

TomOleDiem:
Thanks for all the detailed comments. I made a final structural pass through the PR and reworked the result around the more general object.

The central definition is now `UnitaryOneParameterGroup`, under `Mathematics/OneParameterSubgroups`. It stores an `AddChar` into the Mathlib unitary subgroup, while its coercion returns the underlying continuous linear operator, so expressions read simply as `U t` without repeated casts. I kept unitarity as the intrinsic condition here, since the operators `U(t)` are generally not self-adjoint.

The long `existsUnique_hamiltonian` proof has been removed. The API now defines the canonical self-adjoint `generator`, proves `U(t) = exp(-itA)`, proves uniqueness, constructs a group from any bounded self-adjoint operator, and packages the correspondence as `stoneEquiv`.

Hamiltonians are now treated as the time-translation specialization rather than as part of the general theorem. `FiniteTarget.unitaryTimeEvolution` is generated by `H/ℏ`; this is exposed directly by `unitaryTimeEvolution_generator`, and the existing formula `U(t) = exp(-itH/ℏ)` follows from the general result.

I also incorporated the proof-structure suggestions in `OneParameterSubgroups/Basic`: the integral calculation is a `calc`, translation of the interval integral is a separate lemma, and the generator and uniqueness arguments have fewer intermediate facts. The redundant matrix specialization remains removed.

This seems to give a cleaner separation between the reusable Stone correspondence and its physical interpretation as time evolution. I am happy to adjust the placement or naming if you would prefer a different organization.

-awaiting-author

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+
+/-- `NormedSpace.exp`'s API wants a `ℚ`-algebra structure (for the `1/n!` coefficients), which
+isn't automatic from `NormedAlgebra ℝ E` alone; derive it once here rather than at each use site. -/
+local instance : NormedAlgebra ℚ E := .restrictScalars ℚ ℝ E
+
+lemma exists_isUnit_intervalIntegral [Nontrivial E] (U : AddChar ℝ E) (hU : Continuous U) :
+    ∃ d : ℝ, 0 < d ∧ IsUnit (∫ x in (0 : ℝ)..d, U x) := by
+  let c : ℝ := ‖(1 : E)‖⁻¹ / 2
+  have hone : 0 < ‖(1 : E)‖ := norm_pos_iff.mpr one_ne_zero
+  have hc : 0 < c := by dsimp [c]; positivity
```
gloges:
```suggestion
  have hc : 0 < c := by positivity
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+  have hevent : ∀ᶠ x : ℝ in 𝓝 0, ‖U x - 1‖ < c := by
+    have hnorm : Continuous (fun x : ℝ => ‖U x - 1‖) :=
+      continuous_norm.comp (hU.sub continuous_const)
+    have hmem : Set.Iio c ∈ 𝓝 ‖U 0 - 1‖ := by simpa using Iio_mem_nhds hc
+    exact hnorm.continuousAt hmem
+  rw [Metric.eventually_nhds_iff] at hevent
+  obtain ⟨r, hr, hrU⟩ := hevent
+  let d := r / 2
+  have hd : 0 < d := by dsimp [d]; positivity
```
gloges:
```suggestion
  have hd : 0 < d := by positivity
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+  let q : Eˣ := {
+    val := d • 1
+    inv := d⁻¹ • 1
+    val_inv := by rw [smul_mul_smul_comm, mul_inv_cancel₀ hd.ne', one_smul, one_mul]
+    inv_val := by rw [smul_mul_smul_comm, inv_mul_cancel₀ hd.ne', one_smul, one_mul] }
+  refine ⟨d, hd, (Units.ofNearby q _ ?_).isUnit⟩
+  calc
+    ‖(∫ x in (0 : ℝ)..d, U x) - (q : E)‖ =
+        ‖(∫ x in (0 : ℝ)..d, U x) - d • (1 : E)‖ := rfl
+    _ = ‖(∫ x in (0 : ℝ)..d, U x) - ∫ _x in (0 : ℝ)..d, (1 : E)‖ := by
```
gloges, line 75:
```suggestion
  calc
    _ = ‖(∫ x in (0 : ℝ)..d, U x) - d • (1 : E)‖ := rfl
    _ = ‖(∫ x in (0 : ℝ)..d, U x) - ∫ _x in (0 : ℝ)..d, (1 : E)‖ := by
```

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+    _ ≤ c * d := by
+      calc
+        _ ≤ c * |d - 0| := intervalIntegral.norm_integral_le_of_norm_le_const (fun x hx => by
+          apply le_of_lt
+          apply hrU
+          rw [Real.dist_0_eq_abs]
+          rw [Set.uIoc_of_le hd.le] at hx
+          rw [abs_of_nonneg hx.1.le]
+          exact hx.2.trans_lt (by dsimp [d]; linarith))
+        _ = c * d := by rw [sub_zero, abs_of_pos hd]
```
gloges:
```suggestion
    _ ≤ c * |d - 0| := sorry
    _ = c * d := by rw [sub_zero, abs_of_pos hd]
```
No need to nest `calc` blocks like this - add intermediate steps if they are useful!

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+    _ < ‖(↑q⁻¹ : E)‖⁻¹ := by
+      have hright : ‖(↑q⁻¹ : E)‖⁻¹ = d * ‖(1 : E)‖⁻¹ := by
+        change ‖d⁻¹ • (1 : E)‖⁻¹ = _
+        rw [norm_smul, Real.norm_eq_abs, abs_inv, abs_of_pos hd]
+        field_simp
+      rw [hright]
+      dsimp [c]
+      nlinarith [inv_pos.mpr hone]
```
gloges:
```suggestion
    _ < d * ‖(1 : E)‖⁻¹ := by dsimp [c]; nlinarith [inv_pos.mpr hone]
    _ = ‖q.inv‖⁻¹ := by simp [q, norm_smul, mul_comm, abs_of_pos hd]
```
Same here.

Physlib/Mathematics/OneParameterSubgroups/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+
+/-- Translating a one-parameter subgroup translates its interval integral. -/
+lemma mul_intervalIntegral_eq_sub (U : AddChar ℝ E) (hU : Continuous U) (s t : ℝ) :
+    U s * ∫ x in (0 : ℝ)..t, U x =
+      (∫ x in (0 : ℝ)..(s + t), U x) - ∫ x in (0 : ℝ)..s, U x := by
+  let L : E →L[ℝ] E :=
+    (LinearMap.mulLeft ℝ (U s)).mkContinuous ‖U s‖ (fun x => norm_mul_le _ _)
+  calc
+    U s * ∫ x in (0 : ℝ)..t, U x = L (∫ x in (0 : ℝ)..t, U x) := rfl
+    _ = ∫ x in (0 : ℝ)..t, L (U x) :=
```
gloges, line 101:
```suggestion
  let L : E →L[ℝ] E := (LinearMap.mulLeft ℝ (U s)).mkContinuous ‖U s‖ (fun x => norm_mul_le _ _)
  calc
    _ = L (∫ x in (0 : ℝ)..t, U x) := rfl
    _ = ∫ x in (0 : ℝ)..t, L (U x) :=
```
Best to keep the `=/≤/≥` [aligned](https://leanprover-community.github.io/contribute/style.html#calculations) in `calc`.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,202 @@
...
+namespace UnitaryOneParameterGroup
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+instance : CoeFun (UnitaryOneParameterGroup H) fun _ => ℝ → (H →L[ℂ] H) :=
+  ⟨fun U t => U.toAddChar t⟩
+
+/-- The operator-valued additive character underlying a unitary one-parameter group. -/
+def operatorAddChar (U : UnitaryOneParameterGroup H) : AddChar ℝ (H →L[ℂ] H) :=
+  (unitary (H →L[ℂ] H)).subtype.compAddChar U.toAddChar
```
gloges: Again, that you find it necessary to define `operatorAddChar` suggests that the definition of `UnitaryOneParameterGroup` is not yet "correct" (in the sense of best serving its use cases).

I've tinkered a bit with this and feel strongly that the starting point should be
```suggestion
/-- A norm-continuous unitary one-parameter group on a complex Hilbert space. -/
structure UnitaryOneParameterGroup (H : Type*) [NormedAddCommGroup H] [InnerProductSpace ℂ H]
    [CompleteSpace H] where
  /-- The additive character from the real parameter to unitary operators. -/
  toAddChar : AddChar ℝ (H →L[ℂ] H)
  /-- foo -/
  mem_unitary : ∀ t, toAddChar t ∈ unitary (H →L[ℂ] H)
  /-- The group is continuous in the operator norm. -/
  continuous : Continuous toAddChar

namespace UnitaryOneParameterGroup

variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]

attribute [coe] UnitaryOneParameterGroup.toAddChar

instance : CoeFun (UnitaryOneParameterGroup H) fun _ => ℝ → (H →L[ℂ] H) := ⟨fun U => U.toAddChar⟩
```
(I think early I mistakenly wrote something like this with self-adjoint in place of unitary - sorry for the confusion! If something doesn't make sense please feel free to ask.) See below for an example of how the unitary property can be used in proofs.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,202 @@
...
+  congr
+  exact AddChar.ext _ _ fun t => Subtype.ext (h t)
+
+@[simp]
+lemma adjoint_eq (U : UnitaryOneParameterGroup H) (t : ℝ) :
+    ContinuousLinearMap.adjoint (U t) = U (-t) := by
+  change star (U t) = U (-t)
+  apply left_inv_eq_right_inv (U.toAddChar t).property.1
+  have h := congrArg Subtype.val (U.toAddChar.map_add_eq_mul t (-t))
+  simpa using h.symm
```
gloges:
```suggestion
@[simp]
lemma adjoint_eq (U : UnitaryOneParameterGroup H) (t : ℝ) : (U t).adjoint = U (-t) := by
  apply left_inv_eq_right_inv <| star_mul_self_of_mem <| U.mem_unitary t
  simp [← AddChar.map_add_eq_mul]
```
An example of using `UnitaryOneParameterGroup.mem_unitary`.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,202 @@
...
+
+/-- The bounded self-adjoint generator of `U`, in the convention `U(t) = exp (-itA)`. -/
+noncomputable def generator (U : UnitaryOneParameterGroup H) : H →L[ℂ] H :=
+  Complex.I • deriv U.operatorAddChar 0
```
gloges: This is a good idea! You can probably just write `deriv U 0`, making use of the coercions.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,202 @@
...
+
+lemma deriv_star_eq_neg (U : UnitaryOneParameterGroup H) :
+    star (deriv U.operatorAddChar 0) = -deriv U.operatorAddChar 0 := by
```
gloges:
```suggestion
lemma adjoint_deriv_eq_neg (U : UnitaryOneParameterGroup H) :
    (deriv U 0).adjoint = -deriv U 0 := by
```

gloges (commented):
A few minor suggestions for `OneParameterSubgroups/Basic.lean`.
Even more than before I think that the definition of `UnitaryOneParameterGroup` is not yet optimal - I've played around a bit with it and have left my suggestion below.
Also, a general comment is that many of the `dsimp` and `change` lines can simply be removed.

TomOleDiem:
Thanks for your suggestions!
I think I have incorporated them as much as possible.
"U.toAddChar" Im still using in one or two places. I wasn't able to get completely rid of them.
-awaiting-author

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+
+noncomputable section
+
+/-- A norm-continuous unitary one-parameter group on a complex Hilbert space. -/
+structure UnitaryOneParameterGroup (H : Type*) [NormedAddCommGroup H] [InnerProductSpace ℂ H]
+    [CompleteSpace H] where
+  /-- The additive character from the real parameter to unitary operators. -/
+  toAddChar : AddChar ℝ (H →L[ℂ] H)
+  /-- foo -/
+  mem_unitary : ∀ t, toAddChar t ∈ unitary (H →L[ℂ] H)
```
gloges: We'll need something more descriptive here.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+lemma ext {U V : UnitaryOneParameterGroup H} (h : ∀ t, U t = V t) : U = V := by
+  cases U
+  cases V
+  congr
+  exact AddChar.ext _ _ h
+
+lemma apply_eq_exp_deriv (U : UnitaryOneParameterGroup H) (t : ℝ) :
+    U t = NormedSpace.exp ((t : ℂ) • deriv U 0) := by
+  simpa only [Complex.coe_smul] using
+    OneParameterSubgroup.apply_eq_exp_smul_deriv U.toAddChar U.continuous t
```
gloges:
```suggestion
lemma apply_eq_exp_deriv (U : UnitaryOneParameterGroup H) (t : ℝ) :
    U t = NormedSpace.exp ((t : ℂ) • deriv U 0) :=
  OneParameterSubgroup.apply_eq_exp_smul_deriv U.toAddChar U.continuous t
```
The `simpa` can be removed!

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+
+lemma deriv_unique (U : UnitaryOneParameterGroup H) (A : H →L[ℂ] H)
+    (h : ∀ t : ℝ, U t = NormedSpace.exp ((t : ℂ) • A)) :
+    A = deriv U 0 := by
+  exact OneParameterSubgroup.generator_unique U.toAddChar A (fun t => by
+    simpa only [Complex.coe_smul] using h t)
```
gloges:
```suggestion
    A = deriv U 0 :=
  OneParameterSubgroup.generator_unique U.toAddChar A h
```
Similarly here.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+  cases V
+  congr
+  exact AddChar.ext _ _ h
+
+lemma apply_eq_exp_deriv (U : UnitaryOneParameterGroup H) (t : ℝ) :
+    U t = NormedSpace.exp ((t : ℂ) • deriv U 0) := by
+  simpa only [Complex.coe_smul] using
+    OneParameterSubgroup.apply_eq_exp_smul_deriv U.toAddChar U.continuous t
+
+lemma deriv_unique (U : UnitaryOneParameterGroup H) (A : H →L[ℂ] H)
```
gloges:
```suggestion
lemma generator_unique (U : UnitaryOneParameterGroup H) (A : H →L[ℂ] H)
```
This lemma is about uniqueness of the generator, not uniqueness of the derivative.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+
+/-- The bounded self-adjoint generator of `U`, in the convention `U(t) = exp (-itA)`. -/
+noncomputable def generator (U : UnitaryOneParameterGroup H) : H →L[ℂ] H :=
+  Complex.I • deriv U 0
+
+lemma adjoint_deriv_eq_neg (U : UnitaryOneParameterGroup H) :
+    (deriv U 0).adjoint = -deriv U 0 := by
+  let A : H →L[ℂ] H := deriv U 0
+  have hA (t : ℝ) : U t = NormedSpace.exp ((t : ℂ) • A) := by
+    simpa [A] using U.apply_eq_exp_deriv t
```
gloges: The simp here can be removed here as well:
```lean
  have hA (t : ℝ) : U t = NormedSpace.exp ((t : ℂ) • A) := U.apply_eq_exp_deriv t
```
Also `hA` is only used in one place, so might as well just use `U.apply_eq_exp_deriv` there (see below).

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+  have hrep : ∀ t : ℝ, U t =
+      NormedSpace.exp ((t : ℂ) • (-star A)) := by
+    intro t
+    calc
+      U t = star (U (-t)) := by
+        change U t = ContinuousLinearMap.adjoint (U (-t))
+        simpa only [neg_neg] using (U.adjoint_eq (-t)).symm
+      _ = star (NormedSpace.exp (((-t : ℝ) : ℂ) • A)) := by rw [hA (-t)]
```
gloges:
```suggestion
      _ = star (NormedSpace.exp (((-t : ℝ) : ℂ) • A)) := by rw [U.apply_eq_exp_deriv (-t)]
```

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+      _ = NormedSpace.exp (star (((-t : ℝ) : ℂ) • A)) := NormedSpace.star_exp _
+      _ = NormedSpace.exp ((t : ℂ) • (-star A)) := by
+        congr 1
+        simp [star_smul]
+  have hgen : A = -star A :=
+    (U.deriv_unique (-star A) hrep).symm
+  change star A = -A
+  exact (neg_eq_iff_eq_neg.mpr hgen).symm
```
gloges:
```suggestion
lemma adjoint_deriv_eq_neg (U : UnitaryOneParameterGroup H) :
    (deriv U 0).adjoint = -deriv U 0 := by
  let A : H →L[ℂ] H := deriv U 0
  refine (neg_eq_iff_eq_neg.mpr ?_).symm
  refine (U.generator_unique (-A.adjoint) fun t ↦ ?_).symm
  calc
    _ = (U (-t)).adjoint := by simp
    _ = (NormedSpace.exp (-t • A)).adjoint := by simp [A, U.apply_eq_exp_deriv]
    _ = NormedSpace.exp (-t • A).adjoint := NormedSpace.star_exp _
    _ = NormedSpace.exp (t • -A.adjoint) := by simp [← ContinuousLinearMap.star_eq_adjoint]
```
Just a stylistic suggestion, but by (1) working "backwards" from the goal (2) sticking to `adjoint` for `ContinuousLinearMap` and (3) making use of `simp` lemmas these proofs can be much easier to follow and maintain.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+    congr 1
+    rw [smul_smul]
+    congr 1
+    ring
+  have hderiv : (-Complex.I) • A = deriv U 0 :=
+    U.deriv_unique ((-Complex.I) • A) hrep
+  calc
+    A = Complex.I • ((-Complex.I) • A) := by rw [smul_smul]; simp
+    _ = Complex.I • deriv U 0 := by rw [hderiv]
+    _ = U.generator := rfl
```
gloges, line 109: The mathematical content of this lemma is identical to `generator_unique` (`deriv_unique`) above. We should probably only have one of these.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+    continuous := ?_
+  }
+  · intro t
+    dsimp [U]
+    apply NormedSpace.exp_mem_unitary_of_mem_skewAdjoint
+    rw [skewAdjoint.mem_iff]
+    simp [hB]
+  · dsimp [U]
+    change Continuous (fun t : ℝ => NormedSpace.exp ((t : ℂ) • B))
+    fun_prop
```
gloges:
```suggestion
def ofSelfAdjoint {A : H →L[ℂ] H} (hA : IsSelfAdjoint A) : UnitaryOneParameterGroup H where
  toAddChar := {
    toFun t := NormedSpace.exp ((t : ℂ) • (-Complex.I • A))
    map_zero_eq_one' := by simp
    map_add_eq_mul' s t := by
      have hcomm : Commute ((s : ℂ) • (-Complex.I • A)) ((t : ℂ) • (-Complex.I • A)) :=
        ((Commute.refl (-Complex.I • A)).smul_left _).smul_right _
      rw [← NormedSpace.exp_add_of_commute hcomm]
      simp [add_smul, add_comm]
    }
  mem_unitary t := by
    apply NormedSpace.exp_mem_unitary_of_mem_skewAdjoint
    simp [skewAdjoint.mem_iff, hA.star_eq]
  continuous := by
    change Continuous (fun t : ℝ => NormedSpace.exp ((t : ℂ) • (-Complex.I • A)))
    fun_prop
```
Just a reorganization of your def/proofs to improve readability. Some things to note: variables `t` `s` left of the `:=` and some `dsimp/change` lines aren't needed

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+
+@[simp]
+lemma ofSelfAdjoint_apply {A : H →L[ℂ] H} (hA : IsSelfAdjoint A) (t : ℝ) :
+    ofSelfAdjoint hA t = NormedSpace.exp ((-(t : ℂ) * Complex.I) • A) := by
+  simp only [ofSelfAdjoint]
+  change NormedSpace.exp ((t : ℂ) • ((-Complex.I) • A)) = _
+  rw [smul_smul]
+  congr 1
+  ring_nf
```
gloges, line 157:
```suggestion
  change NormedSpace.exp ((t : ℂ) • ((-Complex.I) • A)) = _
  rw [smul_smul]
  ring_nf
```
Some lines aren't needed!

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+/-- Stone's correspondence between norm-continuous unitary one-parameter groups and bounded
+self-adjoint generators. -/
+noncomputable def stoneEquiv :
+    UnitaryOneParameterGroup H ≃ {A : H →L[ℂ] H // IsSelfAdjoint A} where
+  toFun U := ⟨U.generator, U.generator_selfAdjoint⟩
+  invFun A := ofSelfAdjoint A.2
+  left_inv U := by
+    apply ext
+    intro t
+    rw [ofSelfAdjoint_apply, U.apply_eq_exp_generator]
```
gloges: The mathematical content of this `left_inv` proof should be its own (simp) lemma:
```lean
@[simp]
lemma ofSelfAdjoint_generator (U : UnitaryOneParameterGroup H) :
    ofSelfAdjoint U.generator_isSelfAdjoint = U := by
  sorry
```
With this in place a proof is just `left_inv _ := by simp`.

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+  right_inv A := by
+    apply Subtype.ext
+    exact ((ofSelfAdjoint A.2).generator_unique A.1 (ofSelfAdjoint_apply A.2)).symm
```
gloges: The mathematical content of this `right_inv` proof should be its own (simp) lemma:
```lean
@[simp]
lemma generator_ofSelfAdjoint {A : H →L[ℂ] H} (hA : IsSelfAdjoint A) :
    (ofSelfAdjoint hA).generator = A :=
  sorry
```
After all, this is the defining property of `ofSelfAdjoint`!
With this in place a proof is just `right_inv _ := by simp`

Physlib/Mathematics/OneParameterSubgroups/Unitary.lean
```diff
@@ -0,0 +1,199 @@
...
+      _ = NormedSpace.exp ((t : ℂ) • (-star A)) := by
+        congr 1
+        simp [star_smul]
+  have hgen : A = -star A :=
+    (U.deriv_unique (-star A) hrep).symm
+  change star A = -A
+  exact (neg_eq_iff_eq_neg.mpr hgen).symm
+
+/-- The generator of a unitary one-parameter group is self-adjoint. -/
+lemma generator_selfAdjoint (U : UnitaryOneParameterGroup H) : IsSelfAdjoint U.generator := by
```
gloges:
```suggestion
lemma generator_isSelfAdjoint (U : UnitaryOneParameterGroup H) : IsSelfAdjoint U.generator := by
```

gloges (commented):
Thanks, I think we are converging on the best form of these results.
There's a pair of lemmas both expressing the uniqueness of the generator - I would suggest keeping the one in terms of `.generator`, but up to you which one is more useful. There are also two lemmas that should be extracted from the `stoneEquiv` proofs expressing the defining properties of `ofSelfAdjoint`.
Other than that my comments/suggestions are more stylistic and I hope useful to demonstrate ways to improve your lean as you continue to learn.

TomOleDiem:
I think I have incorporated your comments as much as possible.
The `ofSelfAdjoint` restructure I wasn't able to get to work. Kept the original version.

gloges:
Hi @TomOleDiem, a few of my previous comments are unresolved - can you address these, ask for help, or comment why you don't think changes are needed?
For the `adjoint_deriv_eq_neg` proof can we at least align the `=/≤/≥` in the `calc` block (c.f. [this comment](https://github.com/leanprover-community/physlib/pull/1510#discussion_r3788043589))?

> The ofSelfAdjoint restructure I wasn't able to get to work.

Please feel free to ask for help on the Zulip if you run into difficulties. I don't think the definition as it stands is up to par - there's no reason that the `toAddChar`, `mem_unitary` and `continuous` terms can't be cut and pasted into the constructor rather than being scattered around it:
```lean
  ...
  exact {
    toAddChar = {
      toFun t := ...
      ... }
    mem_unitary t := by
      ...
    continuous := by
      ....
  }
```
If you want to take a crack at tidying up further all the better, but no worries if you want to leave it at that :) (It may help to have that local NormedSpace instance like in the other file.)

TomOleDiem:
Addressed all three remaining suggestions:

- `adjoint_deriv_eq_neg`: rewrote as a backward calc proof working with `.adjoint` directly, no more `star`/`change`.
- `ofSelfAdjoint`: folded `toAddChar`/`mem_unitary`/`continuous` into a single constructor instead of separate `let` + `refine`.
- `ofSelfAdjoint_apply`: dropped the redundant `congr 1`.

-awaiting-author

gloges (approved):
Looks good - approved and will merge shortly!
For the future having much smaller PRs will help accelerate the review process and thoroughly scrutinizing your code (removing duplication, removing unnecessary definitions, etc.) will avoid putting the burden of cleaning up LLM-generated code away from the reviewer.

PR #1513, opened 2026-08-13
docs(cosmology): add FLRW API map

Robby955:
## Motivation

Related to #1414. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree. Cosmology has no map yet.

## Changes

Adds `Physlib/Cosmology/FLRW/API-map.yaml`, 7 requirements of which 5 are done. The done entries cover the spatial geometry with its metric profile function and flat limits, the first and second order Friedmann equations, the Hubble parameter, and the deceleration parameter with the relation ∂ₜH = -H²(1+q). The two open entries record the `FLRW` type, which is not yet a concrete structure, and the matter content block (continuity equation, equation of state, density scaling), which the source lists as open items. The six modules that contain no declarations yet are summarized in the Overview rather than given per-file rows.

Checked with `scripts/api_map_linter.py`: all location files and declaration names resolve.

### DIFF
--- a/Physlib/Cosmology/FLRW/API-map.yaml
+++ b/Physlib/Cosmology/FLRW/API-map.yaml
@@ -1,0 +1,98 @@
+version: v0.1
+
+Title: FLRW cosmology
+
+Overview: |
+    A Friedmann-Lemaître-Robertson-Walker (FLRW) spacetime models a homogeneous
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - looks good to me. Though PR is marked as draft so won't merge until marked as ready.

PR #1515, opened 2026-08-13
docs(operators): add quantum operators API map

### LINKED ISSUE #881: API: Operator Algebra
The details of this issue are not yet completed. The issue is here to track the progress of the API specified by the title. Parts of this API may already be included within the project. There are a number of things we need to specify, help in specifying these points is appreciated.

## Key data structure 

What is the key data structure the API is built around.

## Need

Why is this API needed, after it is completed, what will it make easier or allow us to do.

## Requirements

A tickbox list of the requirements of the API. This should include things which have already been done as well as things still to be worked on.

## Corresponding file system

If it exists, the part of the PhysLean which corresponds to this API.

Robby955:
## Motivation

Closes #881. Related to #1414. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree.

## Changes

Adds `Physlib/QuantumMechanics/Operators/API-map.yaml`, 17 requirements of which 12 are done, covering the 21 modules in the directory.

The done entries cover the unbounded-operator framework (dense domain, symmetry, essential self-adjointness, the adjoint calculus, the self-adjointness criteria and unitary conjugation), the spectrum with its resolvent, regularity domain and defect numbers, the numerical range with Toeplitz-Hausdorff, reality of the spectrum of a self-adjoint operator, spectral measures as projection-valued measures, multiplication operators, position and momentum with the radial multipliers, the canonical commutation relation, angular momentum and the rotation algebra, expectation value through the Robertson and Robertson-Schrodinger uncertainty bounds, and the separate one-dimensional development.

The five open entries are the spectral theorem with its functional calculus and Stone's theorem, self-adjointness of momentum on H1 and of momentum squared on H2, angular momentum as an operator on the Hilbert space with its spectrum and ladder operators, the concrete position-momentum uncertainty relation, and the example operators that separate symmetric from self-adjoint. Each of those is either recorded as an open item in the source or absent from it; no module in the directory carries a proof gap.

Checked with `scripts/api_map_linter.py`: 12 checked, 234 names ok, no missing files or names.

### DIFF
--- a/Physlib/QuantumMechanics/Operators/API-map.yaml
+++ b/Physlib/QuantumMechanics/Operators/API-map.yaml
@@ -1,0 +1,273 @@
+version: v0.1
+
+Title: Operator algebra
+
+Overview: |
+    An observable of a quantum system is a self-adjoint operator on the Hilbert space of
+    states, and its measured values are the points of the spectrum of that operator. The
+    observables of interest are almost never defined on the whole space. Position
+    multiplies a wavefunction by a coordinate and momentum differentiates it, and both
+    send square-integrable functions out of the space unless the wavefunction decays or
+    is smooth enough. The object to work with is therefore an operator defined on a dense
+    subspace, a partially defined linear map, together with the conditions that make it a
+    legitimate observable, namely a dense domain, symmetry of the inner-product pairing
+    on that domain, and self-adjointness, which is strictly stronger than symmetry once
+    the domain is a proper subspace.
+
+    This API builds that framework and then the standard observables of non-relativistic
+    quantum mechanics on `Space d`. Multiplication by a real function of position gives
+    the potential energy and, componentwise, the position operator; powers of the radius
+    give the multipliers used for central potentials. Differentiation gives the momentum
+    operator -iℏ∂ᵢ. The two do not commute, and the canonical commutation relation
+    [xᵢ, pⱼ] = iℏ δᵢⱼ is proved on Schwartz maps, alongside the commutators of the
+    angular momenta 𝐋ᵢⱼ = xᵢpⱼ - xⱼpᵢ, which close into the Lie algebra of rotations and
+    commute with 𝐋², with the momentum squared, and with every regularized power of the
+    radius.
+
+    The statistics of an observable T in a state ψ are the expectation value
+    ⟨T⟩ = re ⟪ψ, T ψ⟫, the variance ‖T ψ - ⟨T⟩ψ‖², and the standard deviation. The API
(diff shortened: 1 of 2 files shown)


Physlib/QuantumMechanics/Operators/API-map.yaml
```diff
@@ -0,0 +1,273 @@
...
+    location: |
+      Physlib/QuantumMechanics/Operators/OneDimension/Unbounded.lean (UnboundedOperator, UnboundedOperator.ofSelfCLM, UnboundedOperator.IsGeneralizedEigenvector, isGeneralizedEigenvector_ofSelfCLM_iff, UnboundedOperator.IsSymmetric); Physlib/QuantumMechanics/Operators/OneDimension/Position.lean (positionOperator, positionOperatorSchwartz, positionOperatorUnbounded, positionStates_generalized_eigenvector_positionOperatorUnbounded, positionOperatorUnbounded_isSymmetric); Physlib/QuantumMechanics/Operators/OneDimension/Momentum.lean (momentumOperator, momentumOperatorSchwartz, momentumOperatorUnbounded, planeWaveFunctional_generalized_eigenvector_momentumOperatorUnbounded, momentumOperatorUnbounded_isSymmetric); Physlib/QuantumMechanics/Operators/OneDimension/Parity.lean (parityOperator, parityOperatorSchwartz, parityOperatorUnbounded, parityOperatorSchwartz_parityOperatorSchwartz, parityOperatorUnbounded_isSymmetric); Physlib/QuantumMechanics/Operators/OneDimension/Commutation.lean (positionOperatorSchwartz_commutation_momentumOperatorSchwartz, positionOperatorSchwartz_momentumOperatorSchwartz_eq, momentumOperatorSchwartz_positionOperatorSchwartz_eq)
```
gloges: I don't think this should be included since the intention is to remove the 1d Hilbert spaces/operators once everything has been moved over to the d-dimensional analogues (see various TODOs in these 1d files).

gloges (commented):
Thanks for getting this API map up and started! A general comment is that all of the descriptions are very verbose compared to all of the other API maps and each addresses many things. For example, the first item covers `HasDenseDomain`, `IsUnbounded`, `IsSelfAdjoint`, `IsEssentiallySelfAdjoint`, closures, adjoints, ... .
To help with this perhaps it makes sense to split of all of the spectral theory content into its own API map, since that is logically separate from operators' algebraic properties.

Robby955:
Thanks! I'll take a look through that and refactor

gloges (approved):
Looks good - thanks!

PR #1516, opened 2026-08-13
docs(units): add units and dimensions API map

Robby955:
## Motivation

Related to #1414. An API map records the implemented status and source location of each requirement for an API, so the gap between a planned API and the current library stays visible in-tree. Units has no map yet, and it is a dependency of most of the library.

## Changes

Adds `Physlib/Units/API-map.yaml`, 16 requirements of which 13 are done, covering the 21 modules in the directory.

The done entries cover the dimensional algebra over an arbitrary base-dimension set, the default five base dimensions and the seven ISQ base quantities with the embedding and reduction between them, unit choices with the rescaling factor and its transitivity, dimension-carrying types, the parametric unit-system layer with the SI choice over the ISQ quantities, unit-dependent types, dimensional correctness, the dimension-tagged value type, the named derived quantities (area, speed, energy, pressure, momentum) with their concrete units, the dimensional behaviour of the derivative and the integral, and the worked examples.

The three open entries are the concrete unit layer (a computable SI choice, and a rescaling operation on the current, amount and luminous-intensity unit types so that decimal multiples can be named), the remaining algebraic, order and topological instances on the tagged type, and mass and velocity, which the source announces in module documentation but which carry no declarations. The first two are recorded as open items in the source; no module under `Physlib/Units` carries a proof gap.

Checked with `scripts/api_map_linter.py`: 13 checked, 236 names ok, no missing files or names.

### DIFF
--- a/Physlib/Units/API-map.yaml
+++ b/Physlib/Units/API-map.yaml
@@ -5,12 +5,14 @@ Title: Units
 Overview: |
     Every physical quantity has a dimension: a rational power of each base dimension,
     recording how the numerical value of the quantity changes when the unit of each base
-    quantity is changed. The default set of base dimensions is length, time, mass, charge
-    and temperature, written L𝓭, T𝓭, M𝓭, C𝓭 and Θ𝓭. Dimensions multiply by adding
+    quantity is changed. The LTMCT set has five base dimensions: length, time, mass,
+    charge and temperature, written L𝓭, T𝓭, M𝓭, C𝓭 and Θ𝓭. Dimensions multiply by adding
     exponents and invert by negating them, so they form a commutative group with a
-    rational power operation. The International System of Quantities takes seven base
-    quantities instead, in which electric charge is derived as current times time; both
-    sets are available and are related by a faithful inclusion and a reduction.
+    rational power operation. The International System of Quantities uses a different set
+    of seven base quantities, in which electric charge is derived as current times time;
+    both sets are available and are related by a faithful inclusion and a reduction.
+    LTMCT is the set over which the typed unit layer (LTMCTUnitChoices and its SI choice)
+    and the named derived quantities are built; the ISQ set carries SIUnitChoices.
 
     A unit system fixes one concrete unit for each base quantity. Changing from one system
     to another multiplies a quantity of dimension d by a positive factor: the product,
@@ -18,8 +20,8 @@ Overview: |
     assigns to that base quantity. That factor is multiplicative in the dimension, equals
     one when the two systems agree, and composes along a chain, so the value of a quantity
     in one system determines its value in every other. The statement is proved once for a
-    general set of base dimensions and specialises to the default five and to the seven
-    ISQ base quantities without being restated.
+    general set of base dimensions and specialises to the LTMCT set and to the ISQ set
+    without being restated.
 
     A value carrying a dimension inherits the additive, order and scalar structure of its
     underlying type, while products and quotients add and subtract the exponents. Area,
@@ -66,7 +68,7 @@ Requirements:
     done: true
     location: "Physlib/Units/Dimension.lean (Dimension.extend, Dimension.extend_exponent_apply, Dimension.extendHom, Dimension.Embedding, Dimension.Projection, Dimension.Embedding.ofBasis)"
 
-  - description: "The default set of five base dimensions is defined with the exponent projection of each."
+  - description: "The LTMCT set of five base dimensions is defined with the exponent projection of each."
     done: true
     location: "Physlib/Units/LTMCTDimensionBase.lean (LTMCTDimensionBase, Dimension.length, Dimension.time, Dimension.mass, Dimension.charge, Dimension.temperature)"
 
@@ -94,7 +96,7 @@ Requirements:
     done: true
     location: "Physlib/Units/ISQDimensionBase.lean (ISQDimensionBase.single_amount_ne_one, ISQDimensionBase.single_luminousIntensity_ne_one)"
 
-  - description: "A faithful embedding carries the default five base dimensions into the ISQ set, sending the charge generator to the derived ISQ charge."
+  - description: "A faithful embedding carries the LTMCT base dimensions into the ISQ set, sending the charge generator to the derived ISQ charge."
     done: true
     location: "Physlib/Units/ISQBridge.lean (Dimension.toISQFun, Dimension.toISQHom, Dimension.toISQHom_apply, Dimension.toISQHom_injective, Dimension.ltmctToISQ, Dimension.toISQHom_C𝓭)"
 
@@ -106,7 +108,7 @@ Requirements:
     done: true
     location: "Physlib/Units/ISQBridge.lean (Dimension.fromISQHom_comp_toISQHom, Dimension.isqToLTMCT_comp_ltmctToISQ)"
 
-  - description: "A unit choice fixes one unit for each of the five default base quantities, with the SI choice of metre, second, kilogram, coulomb and kelvin."
+  - description: "A unit choice fixes one unit for each of the five LTMCT base quantities, with the SI choice of metre, second, kilogram, coulomb and kelvin."
     done: true
     location: "Physlib/Units/Basic.lean (LTMCTUnitChoices, LTMCTUnitChoices.SI, LTMCTUnitChoices.SI_length, LTMCTUnitChoices.SI_time, LTMCTUnitChoices.SI_mass, LTMCTUnitChoices.SI_charge, LTMCTUnitChoices.SI_temperature)"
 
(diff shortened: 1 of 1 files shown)


jstoobysmith:
I think the same comment here applies to @gloges's comment on your other API file PR yesterday. These seem a lot more verbose than usual. Could we refactor this one accordingly. 

awaiting-author

Robby955:
I refactored to one concept per requirement. I can also split WithDim into its own map as well if that would help. Let me know what you all think @gloges @jstoobysmith

Robby955:
-awaiting-author

jstoobysmith (approved):
Looks good! APproved

jstoobysmith:
@Robby955 Up to you, I've approve this for now. But if you think WIthDim deserves its own API map, that seems a good idea. Let me know!

awaiting-author

NicolasRouquette:
Thanks for this — the map is accurate and I'd like it to land. One wording request, on a word I put there myself.

The map calls `LTMCTDimensionBase` **the default** basis in eight places (Overview ×2, plus the requirement lines for the five base dimensions, the ISQ embedding, the five-field unit choice, `equivUnitSystem`, the computable-SI open item, and `WithDim.scaleUnit_val`). That is faithful to the source: the word is mine, from #1447 — `Physlib/Units/Dimension.lean:25,45`, `LTMCTDimensionBase.lean:11,13,31`, `ISQDimensionBase.lean:21`, `ISQBridge.lean:12`, `SIUnitChoices.lean:14`, `UnitSystem.lean:76,105,141,190,192`. 

I now think it was the wrong word, and an API map is the last place it should be fixed in, so I'd rather retire it here and follow up in the source. Four reasons:

**1. Nothing in the API is defaulted to it.** 

`Dimension` is declared `structure Dimension (B : Type)` (`Dimension.lean:46`) — no `optParam`, no default instance, no elaboration path that supplies `LTMCTDimensionBase` when you omit it. Every use site writes its basis. So "default" names no behaviour a reader can invoke or rely on. Every other description in this map cashes out in declarations the linter checks; this one cashes out in nothing.

**2. In software the word is normative, and the norm it states is the one #1447 removed.**

"Default" reads as *what you get if you don't choose, and what you should use unless you have a reason not to*. Parameterising `Dimension` over `B` was precisely about no basis holding that position. The Overview shows the strain: the change-of-unit theorem "is proved once for a general set of base dimensions and specialises to the default five and to the seven ISQ base quantities without being restated" — the theorem treats the two symmetrically, the sentence ranks them. An
API map is what a contributor reads to decide where new material belongs, so the ranking has a concrete cost: it says to state new results over `LTMCTDimensionBase`, which is how the five-generator hardwiring comes back.

**3. Metrologically it points at the wrong set.** 

If any base-quantity set has a standing claim, it is the ISQ seven of ISO 80000-1 — the standard this map cites in References. LTMCT is charge-based and *deliberately* not that set (`LTMCTDimensionBase.lean:23-25`). Calling the non-standard set "the default" and then introducing the standard one with "takes seven base quantities **instead**" inverts their standing for a reader who meets the map before the code. Also, the VIM defines a base quantity as one of a conventionally chosen subset of *a given system of quantities* (JCGM 200:2012, 1.4) — relative to a system, never absolutely. "Default base dimension" is not a notion the vocabulary has.

The map already carries its own counterexample: the last requirement records a basis of bits and symbols, "a set of base dimensions with no physical unit system". That is not a deviation from a default; it is another instantiation, on equal footing.

**4. The vocabulary decides who feels addressed, and PhysLib's reach extends past physics.**

Electrical, chemistry, photometry, and the information-theoretic use this map already documents (bits and symbols) are all within reach of the dimensional algebra as it stands — and the layer that serves several of them is built and proved(`ISQDimensionBase`, `ISQBridge`, `SIUnitChoices`). To a reader from any of those fields, "the default set of base dimensions" reads as *the one you already know*, and the one they know is ISO 80000-1's seven: L, M, T, I, Θ, N, J. LTMCT is a different set, and the differences are ones each audience would rather see up front than meet later:

* **electrical** — the base quantity is charge, not current, so exponents differ from the handbook: capacitance is `M⁻¹L⁻²T²C²` over LTMCT against `M⁻¹L⁻²T⁴I²` in ISO 80000-6, the time exponent 2 against 4. (In a library downstream of PhysLib I maintain on this basis, there is a small rendering layer that re-expresses charge exponents in ISQ current before a dimension is cited against 80000-6. It is cheap, but it exists because the basis is a choice.)
* **chemistry** — LTMCT has no generator for amount of substance, so molar mass and mass, or molar concentration and number density, are not dimensionally separated.
* **photometry** — no generator for luminous intensity, so photometric and radiometric quantities cannot be told apart by dimension.

Each is a defensible convention and PhysLib is entitled to it; the last two are the ordinary choices in much of theoretical physics. The point is only that they are conventions, and stated as a default they are conventions the reader cannot see. Naming the sets instead — "the LTMCT set", "the ISQ set" — costs nothing and lets a chemist or an electrical engineer recognise at once which layer fits their work, and that `ISQDimensionBase` is there *for* them rather than an alternative off the main path. This map records that PhysLib proves amount of substance and luminous intensity independent in ISQ (`ISQDimensionBase.single_amount_ne_one`, `single_luminousIntensity_ne_one`) — precisely the content those readers are looking for. Naming is what leads them to it.

**Suggested wording** — name the set instead of ranking it:

* Overview: 
  "The default set of base dimensions is length, time, mass, charge and temperature, written …" →
  "PhysLib's five-generator LTMCT set is length, time, mass, charge and temperature, written …"

* Overview: 
  "specialises to the default five and to the seven ISQ base quantities" → 
  "specialises to the LTMCT set and to the ISQ set"

* Requirements: 
  "the default five base dimensions" / "the default base dimensions" / "the five
  default base quantities" →
  "the LTMCT base dimensions" (or "the LTMCT set").

Where LTMCT's role in the library is worth recording, record the fact rather than the rank — e.g. "LTMCT is the set over which the typed unit layer (`LTMCTUnitChoices` and its SI choice) and the named derived quantities are built; the ISQ set carries `SIUnitChoices`." That is checkable against declarations, and it leaves a reader free to choose a basis on the merits.

I would avoid "canonical", "primary" or "standard" as substitutes: the first two carry the same ranking in a quieter voice, and the third is false of LTMCT in the ISO sense.

None of this touches the content of the map, which I think is right. Happy to send the matching docstring pass upstream as a separate PR so this one stays a docs-only addition — or to fold the source edits in here if you would rather the vocabulary move in a single change.

Robby955:
Thanks @NicolasRouquette much appreciated and I agree, I have pushed an update.

`Dimension` is declared `structure Dimension (B : Type)` with no `optParam` and no default instance, so every use site writes its basis. "Default" named a behaviour a reader cannot invoke. Every other description in the map cashes out in declarations the linter checks; that one cashed out in nothing.

Pushed in 203bd05, using your wording:

- Overview: "The default set of base dimensions is ..." becomes "The LTMCT set has five base dimensions: length, time, mass, charge and temperature, written L𝓭, T𝓭, M𝓭, C𝓭 and Θ𝓭."
- Overview: "The International System of Quantities takes seven base quantities **instead**" becomes "uses a different set of seven base quantities". You flagged that clause, and "instead" was doing the same ranking work in a quieter voice.
- Overview: "specialises to the default five and to the seven ISQ base quantities" becomes "specialises to the LTMCT set and to the ISQ set", so the sentence treats the two sets the way the theorem does.
- Added your factual replacement so the information is not just dropped: "LTMCT is the set over which the typed unit layer (`LTMCTUnitChoices` and its SI choice) and the named derived quantities are built; the ISQ set carries `SIUnitChoices`."
- The five requirement lines now read "the LTMCT set", "the LTMCT base dimensions", "the five LTMCT base quantities".

No "default" left in the file, and no "canonical", "primary" or "standard" swapped in. `scripts/api_map_linter.py`: 63 checked, 236 names ok, no missing files or names.

 M⁻¹L⁻²T²C² against M⁻¹L⁻²T⁴I² for capacitance reads as a typo until you know the basis is charge-based, which is a thing a reader should meet before citing a dimension against 80000-6 rather than after.

And yes sure if you wouldn't mind, you can open as a PR.

This PR stays docs-only, so the source change gets reviewed on its own terms. Thirteen sites across six files still say "default basis": `Dimension.lean:25,45`, `LTMCTDimensionBase.lean:11,13,31`, `ISQDimensionBase.lean:21`, `ISQBridge.lean:12`, `SIUnitChoices.lean:14`, `UnitSystem.lean:76,105,141,190,192`. 

**`WithDim` stays in this map.** Fourteen of the entries are `WithDim`, but the tagged-type story is continuous with the dimension algebra above it, and splitting now would put a cast, a scaling law and five named quantities in a file that has to restate the algebra to make sense. I will split it out when it grows past that @jstoobysmith 

Let me know if I missed anything of course,

Robby955:
-awaiting-author

NicolasRouquette:
Looks good! One small suggestion: explain the acronyms for the dimension bases where they are first introduced in the doc.

**LTMCT** — the initials of length, time, mass, charge and temperature, which nothing states. The Overview lists the five in that order immediately after the name, so it is recoverable, though the two T's are time and temperature:

> The LTMCT set has five base dimensions — length, time, mass, charge and temperature, whose initials give the name — written L𝓭, T𝓭, M𝓭, C𝓭 and Θ𝓭.

**ISQ** — used 34 times; "International System of Quantities" appears twice, and the two are never joined. `ISQDimensionBase.lean` does join them ("The International System of Quantities (ISQ), specified in ISO/IEC 80000-1…"), so the map is the only place a reader meets the bare acronym:

> The International System of Quantities (ISQ) uses a different set of seven base quantities…

Both names then appear bare in the requirement lines, which is where a reader who greps for a declaration is likely to meet them first.

Happy to carry the same gloss in the docstring pass instead, if you would rather keep the map lean — just say which, so the two do not drift.

jstoobysmith:
Willl put 

awaiting-author 

on till @NicolasRouquette's comments above get addressed (just to help keep track of PRs)

Robby955:
Thanks for the bump @jstoobysmith totally missed this one!

@NicolasRouquette, Both acronyms now get their gloss at first use in the map. The LTMCT sentence reads "length, time, mass, charge and temperature, whose initials give the name", and the first "International System of Quantities" now carries "(ISQ)", so the bare uses further down all resolve to something already expanded.

On which file should carry it, I think the map, since that is where a reader meets the acronyms first. The docstring pass does not need to duplicate the gloss, so there is nothing to keep in sync.

Robby955:
-awaiting-author

jstoobysmith:
Thanks @Robby955 will merge now :).

PR #1519, opened 2026-08-14
feat(SpaceAndTime): add reference frames

RaunakChhatwal:
A point in physical space and a list of coordinates are different kinds of data: coordinates require a chosen origin and basis. Existing uses of `Space` can blur this distinction by treating affine points as though they belonged to a coordinate vector space.

This PR introduces time-dependent reference frames, each carrying an affine origin and coordinate basis. Vectors expressed in a frame use coordinates relative to its basis and can be converted into geometric displacements. The API also describes inertial, orthonormal, and metric-conserving frames, recovering the usual Euclidean norm and inner-product formulas for orthonormal coordinates.

### DIFF
--- a/Physlib/SpaceAndTime/ReferenceFrame.lean
+++ b/Physlib/SpaceAndTime/ReferenceFrame.lean
@@ -44,7 +44,7 @@ A reference frame can be pictured as a coordinate grid carried through time. It 
 motion is described, not an additional physical object moving with the particles.
 -/
 
--- TODO: add timeOrigin to enable time translations
+TODO "Add `timeOrigin` to enable time translations."
 /-- A time-indexed choice of affine origin and displacement basis in `d`-dimensional space. -/
 structure ReferenceFrame (d : ℕ) where
   /-- The point assigned coordinate zero at each time. -/
@@ -175,7 +175,7 @@ def contDispEquiv (t : Time) : frame.Vector ≃L[ℝ] EuclideanSpace ℝ (Fin d)
   (componentContLinearEquiv frame).trans (frame.basis t).equivFun.toContinuousLinearEquiv.symm
 
 /-- The physical norm on frame vectors, pulled back from geometric displacement space. -/
-instance [Fact frame.IsMetricConserved] : NormedAddCommGroup frame.Vector :=
+instance [_h : Fact frame.IsMetricConserved] : NormedAddCommGroup frame.Vector :=
   let normedSpace :=
     NormedAddCommGroup.induced _ _ (dispEquiv 0).toLinearMap (dispEquiv 0).injective
   let metricSpace :=
@@ -211,3 +211,7 @@ lemma inner_euclidean_if_orthonormal [h : Fact frame.Orthonormal] :
         (basis.repr.symm <| WithLp.toLp 2 w.components) := by rfl
     _ = inner ℝ (WithLp.toLp 2 v.components) _ := by rw [basis.repr.symm.inner_map_map]
     _ = _ := by rw [PiLp.inner_apply]; simp_rw [Real.inner_apply]
+
+end ClassicalMechanics.ReferenceFrame.Vector
+
+end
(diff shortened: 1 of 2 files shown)


Physlib/SpaceAndTime/ReferenceFrame.lean
```diff
@@ -0,0 +1,213 @@
...
+variable {d : ℕ}
+
+/-!
+## A. Reference frames
+
+A reference frame can be pictured as a coordinate grid carried through time. It is part of how
+motion is described, not an additional physical object moving with the particles.
+-/
+
+-- TODO: add timeOrigin to enable time translations
```
jstoobysmith: If you write this as: 

TODO "add timeOrigin ..." 

It will get picked up by our TODO list :).

Physlib/SpaceAndTime/ReferenceFrame.lean
```diff
@@ -0,0 +1,213 @@
...
+## C. Vectors in a reference frame
+
+`frame.Vector` stores the coordinates of a displacement; it is not a second kind of geometric
+displacement. Its dependence on `frame` records which basis gives those coordinates their meaning.
+The map `dispEquiv` converts the coordinates back into the displacement they represent at a
+given time.
+-/
+
+/-- The `d` real components used to represent a displacement in `frame`. -/
+structure Vector (frame : ReferenceFrame d) where
```
jstoobysmith, line 139: `Displacement` might be a better name.

Physlib/SpaceAndTime/ReferenceFrame.lean
```diff
@@ -0,0 +1,213 @@
...
+/-- In an orthonormal frame, the inner product is the sum of component products. -/
+lemma inner_euclidean_if_orthonormal [h : Fact frame.Orthonormal] :
+    ∀ v w : frame.Vector, inner ℝ v w = ∑ i, v.components i * w.components i := by
+  intro v w
+  let basis := (frame.basis 0).toOrthonormalBasis (h.out 0)
+  calc
+    _ = inner ℝ (basis.repr.symm <| WithLp.toLp 2 v.components)
+        (basis.repr.symm <| WithLp.toLp 2 w.components) := by rfl
+    _ = inner ℝ (WithLp.toLp 2 v.components) _ := by rw [basis.repr.symm.inner_map_map]
+    _ = _ := by rw [PiLp.inner_apply]; simp_rw [Real.inner_apply]
```
jstoobysmith, line 215: can we end the sections and namespaces opened.

jstoobysmith (commented):
Looks great - a couple of comments for small things.

Physlib/SpaceAndTime/ReferenceFrame.lean
RaunakChhatwal: `Vector` is supposed to be much more expressive. For example, if a particle stores its position relative to the frame's origin as Time -> frame.Vector, its velocity and acceleration are just derivatives of position function. Therefore frame.Vector should also express velocity, acceleration, even forces, momentum, etc.

jstoobysmith: I think this may lead to problems down the line, since e.g. displacement and velocity transform differently under a change in frame, so I would naively put them into different types. Using different types also means you can't cast between things incorrectly, e.g. set a force equal to a distance.

It is also possible to define derivatives so they land in different spaces, the derivative from `Time -> frame.Vector` shouldn't really land in `frame.Vector` (the dimensions don't agree).

RaunakChhatwal: The problem of conflating different quantities is deep, but keeping them different types conflicts with the primary purpose of reference frames: a coordinate system in which calculations use coordinate components. Newton's laws are in terms of motion relative to a frame of reference, not frame-independent motion.

jstoobysmith: Ok, good point. Can we modify the doc-string to say something along these lines then.

jstoobysmith (approved):
Approved this! Assuming it passes the linters, I will merge. Many thanks for this first PR - hope it is the first of many :).

RaunakChhatwal:
Thanks!

PR #1521, opened 2026-08-14
chore: Bump 4.33

NicolasRouquette:
# bump to v4.33.0

Closes the build portion of #1514.

## What's in here

- `lean-toolchain` → `leanprover/lean4:v4.33.0`; `lakefile.toml` mathlib + doc-gen4 revs → `v4.33.0`; `lake-manifest.json` regenerated from a clean `.lake` via `lake update`; README Lean badge updated.
- Proof/definition repairs across ~70 files so that `lake build` (default targets `Physlib` + `QuantumInfo`) completes with zero errors. No declaration was deleted or weakened, no `sorry`/`axiom` introduced; with a handful of exceptions noted below, every original proof is kept verbatim.

## Why so many `set_option backward.isDefEq.respectTransparency false in` lines

The overwhelming majority of the breakage (~90% of ~380 errors) has a single root cause: in Lean v4.33, `isDefEq` now [**respects the ambient transparency level**](https://lean-lang.org/doc/reference/latest/releases/v4.33.0/#The-Lean-Language-Reference--Release-Notes--Lean-4___33___0-_LPAR_2026-08-10_RPAR_--Highlights--Breaking-Changes--Transparency). In v4.32, when `rw`/`simp`/unification matched a pattern against a goal, it would unfold plain (semireducible) `def`s even in contexts restricted to `reducible`/`instances` transparency. In v4.33 such defs are an opaque wall, and the compiler provides `backward.isDefEq.respectTransparency false` as the per-declaration escape hatch that restores the old behavior — the same option this repository already used at 30+ sites before this bump, and which mathlib v4.33 itself uses to guard e.g. its `mfderiv` proofs.

Physlib is disproportionately affected because its house style builds abstraction barriers out of plain `def` type synonyms and proofs freely cross them:

- `ACCSystemCharges.Charges := Fin n → ℚ` — anomaly-cancellation proofs state goals at `Charges` and rewrite with pointwise lemmas that only match at `Fin n → ℚ`;
- `Space d` / `Vector d` / `CoVector d` vs their `Fin 1 ⊕ Fin d → ℝ` unfoldings; `HSOp ℋ` vs `L ℋ`; `Prob`, `HermitianMat`, `Distribution` subtype carriers;
- raw `![![…]]` lambdas used where `Matrix ι ι _` is expected;
- one-field Prop defs such as `IsReindexing`, where anonymous-constructor proofs elaborate at the unfolded `… ∧ …` type and are consumed at the folded type (surfacing as *kernel* errors on `_proof_i` auxiliaries);
- the `deriving Fintype` handler's generated rewrites failing through `↑enumList` coercions (surfacing as "declaration has metavariables").

Under v4.32, unification silently bridged each crossing; under v4.33 each one fails — as a refused `rw` motive ("motive is not type correct" / "Did not find an occurrence"), a `simp` that "made no progress", a stuck `rfl`, an LCNF-signature failure, or a kernel-level type mismatch. One semantic change, many symptom shapes; the new diagnostic says it directly: *"The target expression is not type-correct under the `implicit` transparency level… usually caused by unfolding of semireducible definitions in prior tactic steps."*

We deliberately applied the option **per declaration** rather than file-wide (or making the synonyms `abbrev`): everywhere the option is absent, v4.33's stricter, faster matching stays active, the original proofs remain byte-for-byte unchanged for review, and no global unfolding behavior changes for downstream consumers. Migrating the affected proof styles off the compatibility option (e.g. barrier-respecting API lemmas) is left as a separate refactor.

## Fixes that are not the `set_option` idiom

- **NNReal/Subtype mk-coercion proofs** (`Physlib/Units/*Unit.lean`, `Units/Basic.lean`): goals restated at ℝ via definitional `show` (matching the unnormalized literal shapes), closed with `push_cast`/`norm_num`/`inv_div`/`div_mul_div_comm`; an ℝ≥0-level `div_mul_div` cocycle lemma added per unit module; proof-irrelevant `⟨r,_⟩ = ⟨r,_⟩` residues closed with `rfl`.
- **`deriving Fintype` on the base-dimension enums** (`LTMCTDimensionBase`, `ISQDimensionBase`): the derive handler generates an ill-typed `Finset` under v4.33; instances written out manually (`elems := {…}; complete := fun x => by cases x <;> decide`).
- **`Matrix.IsUpperTriangular`**: mathlib v4.33 now provides an identical, strictly more general declaration (`[Zero R]` vs local `[CommRing R]`), so the local `abbrev` in `Mathematics/SchurTriangulation.lean` is removed and downstream uses resolve to mathlib's; the deprecated `det_of_upperTriangular` call is renamed accordingly.
- **mathlib normal-form / rename drift**: `setOf → Set.ofPred` refactor (explicit proofs replace derailed `grind`/`tauto` in `QuantumInfo/ForMathlib/SionMinimax.lean`), `chartAt_self_eq` no longer default-simp (added where `fun_prop` stalled on self-model charts), `Equiv.apply_eq_iff_eq_symm_apply` → `Equiv.eq_symm_apply`, `Set.restrict` → `Set.domRestrict`, `zpow_ofNat` now default-simp (a `simp`-normalized hypothesis carries `x ^ (3:ℕ)` while a `simp only`-pinned goal keeps `x ^ (3:ℤ)` — display-identical, `rw`-mismatched; align the `simp only` set), and a few staged `simp only` splits where lemma-firing order now matters.

## Checklist from #1514

- [x] Update mathlib rev in lakefile.toml
- [x] Update doc-gen4 rev in lakefile.toml
- [x] Remove `.lake`, run `lake update`
- [x] `lean-toolchain` updates correctly (v4.33.0)
- [x] Update the Lean version badge in README.md
- [x] `lake build` with no errors (9327 jobs)
- [x] CI linters all pass: `check_file_imports`, `check_dup_tags`, `sorry_lint`, `runPhyslibLinters`
- [x] `lake exe lint_all` runs; **caveat**: its advisory transitive-import step (ImportGraph `redundantImports`, not run in CI) now flags 57 files because mathlib v4.33's import-graph reshuffle made many direct imports graph-reachable. Three were genuinely redundant and are removed here (`SU5/Quanta/IsViable.lean`, `Space/Origin.lean`, `Space/Module.lean` — build stays green). The rest are left untouched: under the module system, graph reachability ≠ visibility (e.g. removing the two flagged imports of `Time/Derivatives.lean` breaks elaboration — the chain crosses non-public edges), so a blanket cleanup needs a module-aware linter first.
- [x] `lake exe TODO_to_yml mkFile`
- [x] `lake exe stats mkHTML`
- [x] `lake exe informal mkFile mkDot mkHTML`
- [x] `lake exe make_tag`
- [x] Generated files removed


nateabr (commented):
I have reviewed this and looks good to me

jstoobysmith:
PhyslibAlpha doesn't build here. You've got two options for this: 
- comment out the scripts that don't build from PhyslibAlpha.lean 
- Fix them (if you think that is easy). 
The latter is better, but part of the 'deal' for PhyslibAlpha is that we don't promise to maintain the code in there, so the former is also valid.

NicolasRouquette:
I fixed PhyslibAlpha so that it builds squeaky clean with 4.33.0 as well.

NicolasRouquette:
@jstoobysmith — answering your [question on Zulip](https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/Bumps.20of.20PhysLean.2E/near/616768854) here, so it stays with the PR.

Yes, all five run without errors. I re-ran them just now at the head commit `f4fa33a3`, since the first pass was at `fd877270`, before the PhyslibAlpha repair. Lean v4.33.0, fresh `lake exe cache get`, `lake build` 9327 jobs and `lake build PhyslibAlpha` 8825 jobs, both successful.

| script | exit | output |
| --- | --- | --- |
| `lake exe lint_all` | 0 | steps 2–6 pass: build successful, "All files are imported correctly into Physlib.lean", no duplicate tags, sorry/pseudo attribution passes, `-- Linting passed for Physlib` and `-- Linting passed for QuantumInfo`. Steps 1 and 7 below. |
| `lake exe TODO_to_yml mkFile` | 0 | `TODOList file made.` |
| `lake exe stats mkHTML` | 0 | 584 files, 157104 lines, 2803 definitions, 7896 lemmas; `HTML file made.` |
| `lake exe informal mkFile mkDot mkHTML` | 0 | `Markdown file made.` / `DOT file made.` / `HTML file made.` |
| `lake exe make_tag` | 0 | prints a fresh tag |

The five generated files were removed afterwards and `git status` is clean.

Three things worth flagging, none of them about the bump itself.

**`lint_all`'s exit code is not a signal.** It ends in `pure 0` whatever its seven steps print, so the output is the check, not the status. Steps 2–6 are the ones CI already runs, and both "Lean based style linters" jobs are green on `f4fa33a3` — `build.yml` for Physlib + QuantumInfo, `alphaBuild.yml` for PhyslibAlpha. The two steps CI does not run are 1 and 7.

**Step 1** (`style_lint`, the one it tells you GitHub does not check) prints 226 hits — double blank lines, ` )`, `⟨ `, and similar — concentrated in `ClassicalMechanics/Lagrangian/TotalDerivativeEquivalence.lean` and `QuantumInfo/Capacity/Capacity.lean`. Not one of them is on a line this PR adds; I checked every `file:line` against the PR diff. They are pre-existing, so I left them alone rather than mixing a style sweep into a bump.

**Step 7** (`redundant_imports`) flags 57 files, which is the caveat already in the PR description: mathlib v4.33's import-graph reshuffle made many direct imports graph-reachable. Three were genuinely redundant and are removed here. The rest stay, deliberately — under the module system, graph reachability is not visibility, so the linter's advice is not always safe to take: dropping the two imports it flags in `SpaceAndTime/Time/Derivatives.lean` breaks elaboration, because the chain crosses non-public edges. Acting on the other 54 wants a module-aware linter first.

Happy to paste any of the full logs if you would rather read them than re-run.

jstoobysmith (approved):
Approved - will merge now!

PR #1522, opened 2026-08-15
Ladder Operators & Harmonic Oscillator

TomOleDiem:
unpolished code for PhyslibAlpha


jstoobysmith:
Hey @TomOleDiem this looks great! Though still needs to pass the automatic linters before we can merge it. There should be instructions in the above automatic message about how to do this locally

jstoobysmith (approved):
Approved - and will merge now.

PR #1523, opened 2026-08-16
feat(QuantumMechanics): add QHO eigenstates

gloges:
Introduces eigenfunctions/states for the `d`-dimensional harmonic oscillator in terms of Hermite polynomials and `SchwartzMap` Gaussians.
Two temperate growth lemmas are included to help prove `eigenfunction_apply`, one for `Space.coord` and the other for finite products.


Physlib/Mathematics/HasTemperateGrowth.lean
```diff
@@ -0,0 +1,34 @@
...
+-/
+@[expose] public section
+
+namespace Function.HasTemperateGrowth
+
+open Finset
+
+/-- The finite product of functions of temperate growth is again of temperate growth. -/
+@[to_fun (attr := fun_prop)]
+lemma prod {ι : Type*} [DecidableEq ι] {s : Finset ι} {E F : Type*} [NormedAddCommGroup E]
```
nateabr: I think having DecidableEq might be unnecessary as this is due to your proof strategy and not the statement itself. This would make every downstream item also need decidableEq - maybe we should use classical inside the proof body.

Physlib/QuantumMechanics/HarmonicOscillator/Eigenstates.lean
```diff
@@ -0,0 +1,134 @@
...
+  intro n n' h
+  obtain ⟨h, i, hi⟩ := Pi.lt_def.mp h
+  exact sum_lt_sum (fun i _ ↦ by simp [h i]) ⟨i, mem_univ i, by simp [hi]⟩
+
+/-!
+### A.2. Eigenfunctions
+-/
+
+/-- The `i`th normalization constant for `Q.eigenfunction n`, `1 / √(2 ^ nᵢ * nᵢ! * √π * ξᵢ)`. -/
+def coeff (i : Fin d) : ℝ := 1 / √(2 ^ n i * (n i)! * √π * Q.ξ i)
```
nateabr: Maybe this can be called eigencoeff, as there are alot of coeff's defined in places, might help having specific names for each one.

nateabr (commented):
This looks great - just two very minor comments

nateabr:
awaiting-author

Physlib/QuantumMechanics/HarmonicOscillator/Eigenstates.lean
gloges: Makes sense - renamed to `eigenCoeff`.

Physlib/Mathematics/HasTemperateGrowth.lean
gloges: I've removed the `DecidableEq ι`; do you know if there are any cons to using `classical` logic? My impression is that this is a "detail" of the mathematical foundations that doesn't matter so much for physlib...

gloges:
Thanks for the comments - I've improved the constant's name and removed `DecidableEq ι` from the temperate growth lemma.

Physlib/Mathematics/HasTemperateGrowth.lean
nateabr: Mhm not sure if I totally understand your question. But there should be no 'cons' as such, but more so classical might be the right choice from a design standpoint. When DecidableEq is in the signature, every downstream caller must also carry that constraint, even if their lemma has nothing to do with deciding equality. Using classical inside the proof keeps the constraint local. 

For a lemma in this case, the end result is the same, your proposition has been proven. I think where DecidableEq comes in when the statement itself needs it - e.g. a def with an if that actually computes data.

I might be repeating myself but you are correct I think this is a nitpick as the end result is the same and the mathematical detail doe not really matter, but might as well not carry a constraint downstream if not needed :)

jstoobysmith (approved):
Will merge now

PR #1525, opened 2026-08-16
docs(QuantumMechanics): add finite-dimensional quantum systems API map

TomOleDiem:
Roadmap for finite-dimensional quantum mechanics based on the C*-algebra of linear maps `𝒜(H) := H →L[ℂ] H`. Suitable for integrating QuantumInfo.

Planning only, no Lean source changes.


Physlib/QuantumMechanics/HilbertSpaces/FiniteTarget/API-map.yaml
```diff
@@ -0,0 +1,197 @@
...
+    algebraic test for complete positivity: a map is CP exactly when its Choi matrix is
+    positive.
+
+ParentAPIs:
+  - "Development in this direction: quantum systems (Physlib/QuantumMechanics/QuantumSystem)"
+  - "Development in this direction: finite target quantum mechanics (Physlib/QuantumMechanics/FiniteTarget.lean)"
+  - "Development in this direction: Hermitian matrices (QuantumInfo/ForMathlib/HermitianMat)"
+  - "Development in this direction: mixed and pure states (QuantumInfo/States)"
+  - "Development in this direction: POVMs (QuantumInfo/Measurements/POVM.lean)"
+  - "Development in this direction: unitary operators on quantum states (QuantumInfo/Operators/Unitary.lean)"
```
jstoobysmith, line 85: These should probably not go here until these APIs are actually built (/moved to Physlib). Maybe we could put them in the above summary instead.

Physlib/QuantumMechanics/HilbertSpaces/FiniteTarget/API-map.yaml
```diff
@@ -0,0 +1,197 @@
...
+
+  - description: |
+      The positive cone `𝒜(H)₊ := {A | ∀ ψ, 0 ≤ ⟨ψ, A ψ⟩}` shall be defined: closed under
+      addition, nonnegative scalar multiples and conjugation `A ↦ B† A B`.
+      The order `A ≤ B := 0 ≤ B - A` on `Observable H` shall be defined here.
+    done: false
+    location: N/A
+
+  - description: |
+      Density operators `State H := {ρ : Observable H // 0 ≤ ρ ∧ tr ρ = 1}` shall be defined,
```
jstoobysmith, line 132: Why `State` and not `MixedState`?

jstoobysmith (commented):
Generally looks good to me! Some comments. Also: How much of this could be generalized to infinite dimensional systems too?

Physlib/QuantumMechanics/HilbertSpaces/FiniteTarget/API-map.yaml
TomOleDiem: Density Operators represent all States.
They form a convex set whose extreme points are called pure.
If one wants to one can define mixed states as not pure.
People often get this mixed up (xD).

jstoobysmith:
Putting

awaiting-author 

tag as some comments above still to address, but also I guess changes may be made based on 
https://leanprover.zulipchat.com/#narrow/channel/479953-Physlib/topic/Integrating.20QuantumInfo/with/616822131 ?

PR #1553, opened 2026-08-20
docs(reference-frame): reference frame API map

Robby955:
Adds an API map for reference frames, covering `Physlib/SpaceAndTime/ReferenceFrame.lean`.

Thirteen requirements are marked done and point at declarations in that file: the `ReferenceFrame`
structure and the construction from reference-point trajectories, the metric-conservation and
orthonormality conditions with the implication between them and the `Fact` instances that carry it
to typeclass inference, the inertiality condition with its origin velocity, and the `Vector` carrier
with its component equivalences, module and topological structure, displacement identification,
pulled-back norm and inner product, and the Euclidean component formulas in an orthonormal frame.

Five requirements are marked not done: a time origin so that time translations can act on frames
(the source already marks this one as pending), the defining property of the origin velocity and of a frame
built from reference points (both are currently opaque, the velocity being a `Classical.choose` with
no exported spec), the relative motion of two inertial frames with the induced transformation law on
frame vectors, the derivative of a trajectory in a frame as the time derivative of its coordinate
components, and the statement that a frame with constant axes whose origin moves uniformly relative
to an inertial frame is itself inertial.

The linter is clean: 13 checked, 21 names verified, no missing files or names. The eight instance
claims are reported as needing the Lean environment, which is the same treatment the other maps get.

On placement. The source here is a single file, `SpaceAndTime/ReferenceFrame.lean`, rather than a
directory, so the map goes in a new `SpaceAndTime/ReferenceFrame/` directory holding only the map.
Three merged maps already sit in directories with no Lean files of their own (GaugeTheory, Qubit,
TopologicalFieldTheory), and this path is what makes `APIMapIndex.lean` derive the module context
`Physlib.SpaceAndTime.ReferenceFrame`, which is the file's actual module name. The tidier end state
would be to move the source to `ReferenceFrame/Basic.lean` so it matches Space, Time, SpaceTime and
TimeAndSpace, but that edits already-merged source in a docs-only change, so I have left it. Happy
to do either.

One note on the overview. The source docstring says space has no automatic zero point. `Space d`
does in fact carry a distinguished zero from `Space/Origin.lean`, so I have phrased it as space
having a zero that an observer is free not to use, which is what the frame structure actually
expresses.

### DIFF
--- a/Physlib/SpaceAndTime/ReferenceFrame/API-map.yaml
+++ b/Physlib/SpaceAndTime/ReferenceFrame/API-map.yaml
@@ -1,0 +1,107 @@
+version: v0.1
+
+Title: Reference frame
+
+Overview: |
+    An observer describing motion needs a coordinate grid, and the grid is a
+    choice rather than a feature of space. Space is affine: two points determine
+    a displacement, but no point is singled out as zero, and a displacement has
+    numerical components only once axes have been picked. A reference frame is
+    that pair of choices, an origin and a set of axes, carried along through
+    time.
+
+    An inertial frame is one whose grid is not being pushed around. Its axes
+    neither rotate nor change scale as time passes, and its origin drifts in a
+    straight line at constant speed. This is a restriction on the observer, not
+    on whatever is being observed, and it is the setting in which the familiar
+    Newtonian statements about free motion hold.
+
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved.

PR #1554, opened 2026-08-20
docs(temperature): temperature API map

Robby955:
Adds an API map for temperature, covering `Physlib/Thermodynamics/Temperature/Basic.lean` and
`Physlib/Thermodynamics/Temperature/TemperatureUnits.lean`.

The Units map already lists Temperature as one of its parent APIs, so this fills in the map that
entry points at.

Twelve requirements are marked done, across 49 named declarations: the `Temperature` structure with
its coercions, topology, zero and extensionality; the inverse temperature with its defining formula,
the construction going the other way, and the round trips in both directions; positivity; continuity
in the inverse temperature and differentiability of the real coordinate on the positive reals; the
large inverse temperature limit; the constructions from a nonnegative real and from a real with a
nonnegativity proof; the derivative of the inverse temperature with respect to the temperature and
the chain rule between the two variables; and on the units side the `TemperatureUnit` structure,
the ratio of two units with its algebraic properties, rescaling with its composition law, and kelvin
with the units built from it.

Three requirements are marked not done: the temperature manifold that the units file's own
documentation refers to but does not yet define, conversion of a temperature value between two
units, and scales whose zero is offset from absolute zero such as ordinary Celsius and Fahrenheit,
as distinct from the absolute Fahrenheit-sized degree that is already there.

The linter is clean: 12 checked, 49 names verified, no missing files or names.

Two things worth flagging. The differentiability requirement is deliberately worded around the real
coordinate rather than the temperature itself, because `ofβ_differentiableOn` is stated for
`fun x : ℝ => ((ofβ (Real.toNNReal x)).val : ℝ)` and `Temperature` carries no normed structure.
And the overview notes that taking temperature nonnegative excludes negative absolute temperatures,
which are physically real in systems with a bounded energy spectrum, since the module scopes itself
to the usual undergraduate notion.

### DIFF
--- a/Physlib/Thermodynamics/Temperature/API-map.yaml
+++ b/Physlib/Thermodynamics/Temperature/API-map.yaml
@@ -3,33 +3,36 @@ version: v0.1
 Title: Temperature
 
 Overview: |
-    Temperature measured from absolute zero is a nonnegative quantity: there is a
-    coldest state, and no scale used here places anything below it. A choice of
-    scale is a choice of how large one degree is, and nothing more, so two such
-    choices differ by a positive ratio and any statement of physics must be
-    unchanged when that ratio is applied. Kelvin is taken as the reference
-    scale and the others are built from it by rescaling.
+    Temperature here is measured from absolute zero and is taken to be
+    nonnegative: no scale used here places a state below absolute zero, which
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Looks good - approved

PR #1556, opened 2026-08-21
The static theorem for reversible dynamics on Hilbert spaces

TomOleDiem:
## Summary

Adds the static classification of reversible dynamics on the bounded-operator algebra of a complex Hilbert space. Every star-automorphism is implemented by unitary conjugation, with the expected scalar-phase ambiguity.

## API structure

- `OperatorAlgebra.Basic` contains only the generic operator-algebraic vocabulary.
- `OperatorAlgebra.HilbertSpace` introduces the physics-facing class `OperatorAlgebra.HilbertSpace H`, bundling the standard complex Hilbert-space assumptions behind one uniform hypothesis.
- The same file defines `B(H)`, gives it a direct canonical `OperatorAlgebra` instance, and defines `Representation A H`. The redundant `HilbertSpaceOperatorAlgebra` alias has been removed.
- A low-priority compatibility instance lets existing Mathlib code with separate normed-group, inner-product, and completeness assumptions use the bundled interface.
- `OperatorAlgebra.Dynamics.Automorphism` contains the Hilbert-space automorphism results and now consistently assumes `[HilbertSpace H]`.

### DIFF
--- a/Physlib/QuantumMechanics/OperatorAlgebra/Automorphism.lean
+++ b/Physlib/QuantumMechanics/OperatorAlgebra/Automorphism.lean
@@ -1,0 +1,92 @@
+/-
+Copyright (c) 2026 Tom Ole Diem. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Tom Ole Diem
+-/
+module
+
+public import Mathlib.Analysis.CStarAlgebra.ContinuousLinearMap
+public import Mathlib.Analysis.Normed.Operator.ContinuousAlgEquiv
+public import Mathlib.Analysis.CStarAlgebra.Hom
+
+/-!
+# Automorphisms of the bounded operators
+
+Reversible transformations of a quantum system act on its observable algebra by
+⋆-automorphisms.
+
+For a complex Hilbert space `H`, every ⋆-automorphism of `B(H)` is implemented by
+unitary conjugation:
+  `A ↦ U A U⋆`.
+
+Two unitaries implement the same transformation exactly when they differ by a
+scalar phase. Consequently,
+  `Aut⋆(B(H)) ≅ U(H) / U(1) ≅ PU(H)`,
+the projective unitary group.
+
+For Hamiltonian dynamics, this projective ambiguity corresponds to
+the freedom to shift a Hamiltonian by a scalar multiple of the identity.
+-/
+
+@[expose] public section
+
+namespace OperatorAlgebra
+
+variable {H : Type*}
+    [NormedAddCommGroup H]
+    [InnerProductSpace ℂ H]
+    [CompleteSpace H]
+
+/--
+Every ⋆-automorphism of `B(H)` is implemented by unitary conjugation.
+-/
+lemma conjStarAlgAut_surjective :
+    Function.Surjective
+      (Unitary.conjStarAlgAut ℂ (H →L[ℂ] H)) := by
+  intro φ
+  obtain ⟨U, hU⟩ :=
+    φ.eq_linearIsometryEquivConjStarAlgEquiv
+      (NonUnitalStarAlgHom.isometry φ φ.injective).continuous
+  refine ⟨Unitary.linearIsometryEquiv.symm U, ?_⟩
+  rw [Unitary.conjStarAlgAut_symm_unitaryLinearIsometryEquiv]
+  exact hU.symm
+
+
+/--
+Two unitaries implement the same ⋆-automorphism of `B(H)` exactly when
+they differ by a scalar phase.
+-/
+lemma conjStarAlgAut_eq_iff
+    (u v : unitary (H →L[ℂ] H)) :
+    Unitary.conjStarAlgAut ℂ (H →L[ℂ] H) u =
+        Unitary.conjStarAlgAut ℂ (H →L[ℂ] H) v ↔
+      ∃ c : unitary ℂ, u = c • v :=
+  Unitary.conjStarAlgAut_ext_iff' u v
+
+
+/--
+The projective unitary group of `H`.
+This is the unitary group modulo the scalar phases which act trivially
+by conjugation.
+-/
+noncomputable def ProjectiveUnitary (H : Type*)
+    [NormedAddCommGroup H]
+    [InnerProductSpace ℂ H]
+    [CompleteSpace H] :=
+  unitary (H →L[ℂ] H) ⧸
+    MonoidHom.ker (Unitary.conjStarAlgAut ℂ (H →L[ℂ] H))
+
+noncomputable instance :
+    Group (ProjectiveUnitary H) :=
+  QuotientGroup.Quotient.group _
+
+/--
+Reversible transformations of `B(H)` are precisely projective unitaries.
+-/
+noncomputable def projectiveUnitaryEquivStarAlgAut :
+    ProjectiveUnitary H ≃*
+      ((H →L[ℂ] H) ≃⋆ₐ[ℂ] (H →L[ℂ] H)) :=
+  QuotientGroup.quotientKerEquivOfSurjective
+    _ conjStarAlgAut_surjective
+
+end OperatorAlgebra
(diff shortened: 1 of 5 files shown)


Physlib/QuantumMechanics/OperatorAlgebra/Automorphism.lean
```diff
@@ -0,0 +1,92 @@
...
+    [NormedAddCommGroup H]
+    [InnerProductSpace ℂ H]
+    [CompleteSpace H]
+
+/--
+Every ⋆-automorphism of `B(H)` is implemented by unitary conjugation.
+-/
+lemma conjStarAlgAut_surjective :
+    Function.Surjective
+      (Unitary.conjStarAlgAut ℂ (H →L[ℂ] H)) := by
```
jstoobysmith: Can we collapse to one line if possible. Likewise with e.g. the arguments to lemmas below.

Physlib/QuantumMechanics/OperatorAlgebra/Dynamics/Automorphism.lean
```diff
@@ -0,0 +1,92 @@
...
+  `A ↦ U A U⋆`.
+
+Two unitaries implement the same transformation exactly when they differ by a
+scalar phase. Consequently,
+  `Aut⋆(B(H)) ≅ U(H) / U(1) ≅ PU(H)`,
+the projective unitary group.
+
+For Hamiltonian dynamics, this projective ambiguity corresponds to
+the freedom to shift a Hamiltonian by a scalar multiple of the identity.
+-/
```
jstoobysmith, line 29: Is B(H) here `H →L[ℂ] H`, if so we should say this, and put an instance of an OperatorAlgebra on it.

jstoobysmith (commented):
Some small comments from me here.

jstoobysmith:
awaiting-author

Physlib/QuantumMechanics/OperatorAlgebra/Dynamics/Automorphism.lean
TomOleDiem: Done

Physlib/QuantumMechanics/OperatorAlgebra/Automorphism.lean
TomOleDiem: I collapsed several lines

TomOleDiem:
-awaiting-author

Physlib/QuantumMechanics/OperatorAlgebra/Basic.lean
```diff
@@ -90,6 +87,32 @@ abbrev Channel (A₁ A₂ : Type*) [OperatorAlgebra A₁] [OperatorAlgebra A₂]
...
+notation "B(" H ")" => H →L[ℂ] H
+
+variable {H : Type*} [NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
+
+/-- The physical operator algebra of bounded operators on a complex Hilbert space. -/
+abbrev HilbertSpaceOperatorAlgebra (H : Type*) [NormedAddCommGroup H] [InnerProductSpace ℂ H] :=
+  OperatorAlgebra B(H)
+
+noncomputable instance : PartialOrder B(H) := CStarAlgebra.spectralOrder _
+instance : StarOrderedRing B(H) := CStarAlgebra.spectralOrderedRing _
```
jstoobysmith: Are these instances already defined? If so maybe shouldn't define them again. Maybe this Hilbert space operator algebra deserves its own file. Maybe have already said this but we should make sure here not to overlap with things already defined in quantum mechanics!

jstoobysmith:
@gloges just checking if you have any thoughts here too?

Physlib/QuantumMechanics/OperatorAlgebra/Basic.lean
TomOleDiem: I redesigned some things and now defined HilbertSpace and B(H).
Im happy for opinions on how to do this best.
I agree that the folders HilbertSpaces and QuantumSystem should be integrated at some point but I dont see a clear path yet.
There are somewhat different philosophies at how general things are set up.

TomOleDiem:
-awaiting-author

Physlib/QuantumMechanics/OperatorAlgebra/HilbertSpace.lean
```diff
@@ -0,0 +1,45 @@
...
+C⋆-algebra of bounded operators on a complex Hilbert space. The C⋆-algebra, Loewner order, and
+ordered-star-ring instances for bounded operators are supplied by Mathlib.
+-/
+
+@[expose] public section
+
+namespace OperatorAlgebra
+
+/-- A complex Hilbert space: a complete complex inner-product space. -/
+class HilbertSpace (H : Type*) extends NormedAddCommGroup H, InnerProductSpace ℂ H, CompleteSpace H
```
jstoobysmith: This we should not be redefining, as it already exists in Mathlib.

jstoobysmith:
awaiting-author

Physlib/QuantumMechanics/OperatorAlgebra/Basic.lean
jstoobysmith: Im thinking e.g. about the commutatorss and the Lie things which you defined on a general operator algebra. They should carry over directly to the operator algebra of a Hilbert space, and be used throughout QuantumMechanics.

jstoobysmith: Maybe I'm viewing this wrong, but I see the OperatorAlgebra results as genearlizing some of the stuff in QuantumMechanics already, as well as adding new results around it (which is why this is very exciting).

Physlib/QuantumMechanics/OperatorAlgebra/HilbertSpace.lean
TomOleDiem: This is what I was able to find:
Mathlib has a `HilbertSpace` (`Mathlib.Analysis.InnerProductSpace.Defs`), but it is a
`@[variable_alias]` structure with no fields, used only as a name for the `variable?` command to
suggest, it carries no data connecting `NormedAddCommGroup`/`InnerProductSpace`/`CompleteSpace`,
so it cannot be used as a single hypothesis to derive them. This class is a genuine bundling of
those three (via `extends`) so that `[HilbertSpace H]` alone gives access to all of them, which is
what `B(H)` and the definitions below need.

Physlib/QuantumMechanics/OperatorAlgebra/Basic.lean
TomOleDiem: Yes I see this as generalizing and also the "correct" way to pursue.
I have the feeling Ill be replacing things rather than integrating them.

TomOleDiem:
-awaiting-author

Physlib/QuantumMechanics/OperatorAlgebra/HilbertSpace.lean
```diff
@@ -0,0 +1,45 @@
...
+
+/-- Mathlib's separate Hilbert-space assumptions provide Physlib's bundled physics interface. -/
+noncomputable instance (priority := low) instHilbertSpace (H : Type*) [NormedAddCommGroup H]
+    [InnerProductSpace ℂ H] [CompleteSpace H] : HilbertSpace H := {}
+
+/-- Bounded operators on a complex Hilbert space, written in the usual physics notation. -/
+notation "B(" H ")" => H →L[ℂ] H
+
+variable {H : Type*} [HilbertSpace H]
```
jstoobysmith: I don't get why here you just can't do 
```suggestion
variable {H : Type*}[NormedAddCommGroup H] [InnerProductSpace ℂ H] [CompleteSpace H]
```

My reluctance to defining `HilbertSpace` as a class is that it blocks you from using all of the lemmas in Mathlib defined around Hilbert space, and I think might lead to instance diamonds. 

I also don't see yet what is saved from this. 

Maybe we could remove this type class here, but keep in mind that it might be useful to define in the future?

jstoobysmith (approved):
This looks good to me, though will wait a bit to merge, until we get any other comments.

PR #1557, opened 2026-08-21
chore: remove 53 defeq set_option

zhikaip:
ran scripts/rm_set_option.py

Files processed:        204
Fully cleaned:          16
Partially cleaned:      13
Unchanged:              174
Lines removed:          53
Lines kept:             706


jstoobysmith (approved):
Approved - many thanks :). Maybe one day we could set up a linter which makes sure non of these are added needlessly

zhikaip:
I think it's moreso the case that mathlib upstream fixes are slowly coming through, but maybe at some point we do need to have a pass (or if AI gets good enough) to identify which ones require local fixes

PR #1565, opened 2026-08-22
feat(tensors): prove the Levi-Civita contraction identities

Robby955:
Proves the two remaining epsilon-epsilon identities for the Lorentzian Levi-Civita tensor,
`leviCivita_contract_self` and `leviCivita_contract_three`, left as stated goals in #1348. Both
statements are unchanged.

The route is through components. Both sides are compared under `(Tensor.basis _).repr` and the
contractions collapse to finite sums over `Fin 1 ⊕ Fin 3`, where the product of two symbols is a
`generalizedKroneckerDelta` and the contraction machinery from #1335 gives the 24 and the 6. The
sign comes out of the metric: when the symbol is nonzero the index map is injective, hence a
bijection on a four-element type, so the four diagonal metric factors multiply to the determinant,
which is minus one in this signature.

Two component lemmas that did not exist before carry the argument and may be useful beyond it:
`toDualMapAtIndex_basis_repr`, the basis representation of index lowering, and
`unitTensor_down_repr`, the basis representation of the unit tensor of color down. The rest of the
new material is rank-four bookkeeping for this file.

The proofs elaborate within the default heartbeat budget, so no `maxHeartbeats` option is needed.
The `nolint checkType` on `leviCivita_contract_self` stays: that linter re-typechecks the statement,
which this change does not alter, and its comment now names the current toolchain. The Tensors API
map entry for these identities is flipped to done with a real location.

One question on placement. The two component lemmas would sit naturally next to their definitions,
`toDualMapAtIndex_basis_repr` in Dual.lean in particular, rather than in this file. I kept
everything local so the diff stays reviewable, and can move them wherever you prefer, here or in a
follow-up.


jstoobysmith:
Will try and give a proper review tomorrow, but I think this could do with lemmas moved around bit. For example there are some IsReindexing lemmas here which should likely be in the corresponding file not here

Physlib/Relativity/Tensors/LeviCivita/Basic.lean
```diff
@@ -176,17 +178,470 @@ lemma leviCivita_antisymm_last : {ε4 | μ ν ρ σ = - (ε4 | μ ν σ ρ)}ᵀ
...
+  have h : (∑ v : Fin 4 → Fin 1 ⊕ Fin 3,
+      (((generalizedKroneckerDelta (fun i => finSumFinEquiv (v i))
+          (id : Fin 4 → Fin 4) : ℤ) : ℝ)
+        * ((generalizedKroneckerDelta (fun i => finSumFinEquiv (v i))
+            (id : Fin 4 → Fin 4) : ℤ) : ℝ))) = 24 := by
+    exact_mod_cast congrArg (fun z : ℤ => (z : ℝ)) hsum
+  rw [Finset.sum_congr rfl (fun v _ => epsEtaSummand_eq v), Finset.sum_neg_distrib, h]
+
+set_option backward.isDefEq.respectTransparency false in
+lemma unitTensor_down_repr {d : ℕ}
```
jstoobysmith: This should be moved to the unitTensor file

Physlib/Relativity/Tensors/LeviCivita/Basic.lean
```diff
@@ -176,17 +178,470 @@ lemma leviCivita_antisymm_last : {ε4 | μ ν ρ σ = - (ε4 | μ ν σ ρ)}ᵀ
...
+  simp only [Fintype.sum_prod_type]
+  rfl
+
+lemma sum_reindex_finSumFinEquiv {M : Type} [AddCommMonoid M] (F : (Fin 4 → Fin 4) → M) :
+    ∑ v : Fin 4 → Fin 1 ⊕ Fin 3, F (fun i => finSumFinEquiv (v i))
+      = ∑ g : Fin 4 → Fin 4, F g := by
+  exact Fintype.sum_equiv (Equiv.arrowCongr (Equiv.refl (Fin 4))
+    (finSumFinEquiv : (Fin 1 ⊕ Fin 3) ≃ Fin 4)) _ _ (fun _ => rfl)
+
+lemma prod_eta_diag_of_injective {v : Fin 4 → Fin 1 ⊕ Fin 3} (hv : Function.Injective v) :
```
jstoobysmith: Should be moved to the `minkowskiMatrix` file

Physlib/Relativity/Tensors/LeviCivita/Basic.lean
```diff
@@ -176,17 +178,470 @@ lemma leviCivita_antisymm_last : {ε4 | μ ν ρ σ = - (ε4 | μ ν σ ρ)}ᵀ
...
+
+lemma sum4_eq {M : Type} [AddCommMonoid M] (F : (Fin 4 → Fin 1 ⊕ Fin 3) → M) :
+    ∑ x : Fin 1 ⊕ Fin 3, ∑ x1 : Fin 1 ⊕ Fin 3, ∑ x2 : Fin 1 ⊕ Fin 3, ∑ x3 : Fin 1 ⊕ Fin 3,
+        F ![x, x1, x2, x3]
+      = ∑ v : Fin 4 → Fin 1 ⊕ Fin 3, F v := by
+  rw [← Equiv.sum_comp vec4Equiv F]
+  simp only [Fintype.sum_prod_type]
+  rfl
+
+lemma sum_reindex_finSumFinEquiv {M : Type} [AddCommMonoid M] (F : (Fin 4 → Fin 4) → M) :
```
jstoobysmith: This is not likely needed as a sperate lemma, similar with a lot of otherr small definitions here.

Physlib/Relativity/Tensors/LeviCivita/Basic.lean
```diff
@@ -176,17 +178,470 @@ lemma leviCivita_antisymm_last : {ε4 | μ ν ρ σ = - (ε4 | μ ν σ ρ)}ᵀ
...
+  · funext m
+    induction m using Fin.succAboveCases (i := j) with
+    | x => rw [Fin.insertNth_apply_same, Fin.insertNth_apply_same]
+    | p k =>
+      rw [Fin.insertNth_apply_succAbove, Fin.insertNth_apply_succAbove]
+      refine congrArg b ?_
+      rw [IsReindexing.inv_equiv_symm_eq, ← Fin.append_succAbove_const_eq_cycleIcc i,
+        Fin.append_right]
+
+lemma metricTensor_repr_apply {d : ℕ} (cc : Color)
```
jstoobysmith: MOve e.g. to the `metricTensor` file

Physlib/Relativity/Tensors/LeviCivita/Basic.lean
```diff
@@ -176,17 +178,470 @@ lemma leviCivita_antisymm_last : {ε4 | μ ν ρ σ = - (ε4 | μ ν σ ρ)}ᵀ
...
+      exact ComponentIdx.DropPairSection.ofFinEquiv_apply_snd _ _ _
+    | p k =>
+      rw [Fin.insertNth_apply_succAbove]
+      conv_lhs => rw [← Fin.succSuccAbove_castAdd_natAdd_apply_natAdd i j k]
+      simp only [Fin.cast_cast, Fin.cast_eq_self]
+      rw [ofFinEquiv_apply_succSuccAbove]
+      simp only [basisIdxCongr_eq_refl, Equiv.refl_apply]
+      exact congrArg b (IsReindexing.inv_id_eq _ _)
+
+lemma crossToSlot_basis_repr {d nA : ℕ} {c : Fin (nA+1) → Color} {cM : Fin 2 → Color}
```
jstoobysmith: Move to `crossToSlot` file.

jstoobysmith (commented):
Some more things to move. 

There are also lots of small definitions/lemmas that are: 
1. Either not needed, 
2. Can be reframed in terms of useful physics API.

jstoobysmith:
awaiting-author

jstoobysmith:
(May also help to split this PR up into smaller bits each touching a single file)

Robby955:
Thanks for the review. I have reorganized the proof around the owning APIs in `ab91a437`.

- The `crossToEnd` and `crossToSlot` component formulas now live in dedicated `RealTensor/Contraction` modules.
- Generic metric and unit component bridges now live in `MetricTensor.lean` and `UnitTensor.lean`; their Minkowski and Kronecker-delta specializations live in `RealTensor/Metrics/Basic.lean` and the new `RealTensor/Units/Basic.lean`.
- The diagonal metric product has been generalized to arbitrary spatial dimension and moved to `MinkowskiMatrix.lean`.
- The one-use sum-reindexing lemma was removed. The reusable free-last generalized Kronecker-delta contraction is now in `KroneckerDelta/Contraction.lean`.
- The two Lorentzian identities and their private component proof now live in `LeviCivita/Contractions.lean`; `LeviCivita/Basic.lean` is again limited to the definition, components, and antisymmetry.

The exact one-sum and Kronecker/Minkowski formulas are specific to the real Lorentz bases. Putting those statements directly in the generic tensor files would either create an import cycle or require basis/contraction compatibility not present in `TensorSpecies`, so the generic bridges and real specializations are separated.

The two headline theorem statements are unchanged. The targeted build, full `lake build Physlib`, changed-file style checks, API-map check, and declaration linters pass.

-awaiting-author

Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
```diff
@@ -113,3 +120,313 @@ lemma euclidLeviCivita_symbol_contract_two (r s t w : Fin 4) :
...
+          Fin 4 → Fin 1 ⊕ Fin 3) i)) (id : Fin 4 → Fin 4) : ℤ) : ℝ)
+        * (((Tensor.basis ![Color.up, Color.up, Color.up, Color.up]).repr ε4)
+            (![x,x1,x2,y1] : Fin 4 → Fin 1 ⊕ Fin 3)
+          * minkowskiMatrix x x * minkowskiMatrix x1 x1 * minkowskiMatrix x2 x2
+          * minkowskiMatrix y1 y1)
+      = -6 * (if y0 = y1 then 1 else 0) := by
+  rw [← sum_eps_three y0 y1]
+  refine Finset.sum_congr rfl fun x _ => Finset.sum_congr rfl fun x1 _ =>
+    Finset.sum_congr rfl fun x2 _ => ?_
+  rw [leviCivita_basis_repr_apply]
```
jstoobysmith: With all of these lemmas: I think the following needs to be done: 

- Rephrase them as API around objects that already exist. 
- Move them to the corresponding file 
- Or if they a small one-line proofs remove them entirely. 

The point is that they do not tell a good story, about what they are and why they are there. i understand they are needed for the calculation, but I think it is always possible to do such a reframing.

Maybe one idea is to reframe this whole file as a series of different forms of `ε4 | μ ν ρ σ ⊗ ε4 | τ(μ) τ(ν) τ(ρ) τ(τ) ` each getting simplier?

Physlib/Relativity/Tensors/RealTensor/Contraction/CrossToSlot.lean
jstoobysmith, line 1: Think this lemma should go next to CrossToSlot definition, or we move CrossToSlot here.

jstoobysmith:
awaiting-author

jstoobysmith:
The other alternative to these PRs is that we could add them to PhyslibAlpha, so we have them, and then could leave the tidying for the future

Robby955:
> The other alternative to these PRs is that we could add them to PhyslibAlpha, so we have them, and then could leave the tidying for the future

Thanks! I’m working through the new review comments now. I’ll first try to rework the proofs around the existing tensor APIs and make the intermediate results mathematically useful. No problem on PhyslibAlpha if you prefer as well.

Robby955:
I made another pass and pushed the revision at b6716831.
The Lorentzian proofs now follow a staged tensor API:
- crossToSlot_basis_repr_apply lives beside crossToSlot;
- full lowering of ε4 is exposed by leviCivita_lowered_basis_repr_apply;
- the tensor contractions reduce to basis-component sums;
- the existing generalized-Kronecker-delta contractions evaluate those sums.

The old private helper forest and separate real CrossToSlot module are gone, for a net reduction of 106 lines from the previous head. I also rebased onto current master, including #1567, and reran the builds and linters.
I kept this in Physlib for this pass because the remaining intermediate lemmas are reusable tensor APIs rather than one-off proof aliases. We can move either or both PRs to PhyslibAlpha if you still think that is the better home after this revision.
-awaiting-author

Physlib/Mathematics/KroneckerDelta/Contraction.lean
```diff
@@ -309,6 +309,25 @@ lemma sum_generalizedKroneckerDelta_mul_cons (σ τ : Fin 4) :
     sum_generalizedKroneckerDelta_cons σ τ 3]
   norm_num [Finset.prod_range_succ]
 
+/-- Symbol-level triple contraction with the free index in the last slot. -/
+lemma sum_generalizedKroneckerDelta_mul_snoc (σ τ : Fin 4) :
+    ∑ h : Fin 3 → Fin 4,
+        generalizedKroneckerDelta (Fin.snoc h σ) id *
```
jstoobysmith, line 315: Could be collapsed to one line. I think you should check for all of these throughout both your pull requests

Physlib/Relativity/Tensors/ComponentIdx/Contraction.lean
```diff
@@ -180,6 +180,15 @@ lemma ofFinEquiv_apply_snd {n : ℕ} {c : Fin (n + 1 + 1) → C}
     (ofFinEquiv (S := S) hij b x).1 j = x.2 := by
   simp [ofFinEquiv]
 
+/-- Away from the two restored positions, `ofFinEquiv` agrees with the original
+component index. -/
+lemma ofFinEquiv_apply_succSuccAbove {n : ℕ} {c : Fin (n + 1 + 1) → C}
```
jstoobysmith, line 185: This is a single rewriting of a lemma that already exists so I just wouldn't define it and use the definition wherever you use this lemma

Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
```diff
@@ -113,3 +140,222 @@ lemma euclidLeviCivita_symbol_contract_two (r s t w : Fin 4) :
...
+    _ = 6 * (if a = b then 1 else 0) := by
+      by_cases hab : a = b
+      · subst hab
+        simp [KroneckerDelta.eq_one_of_same]
+      · rw [if_neg hab,
+          KroneckerDelta.eq_zero_of_ne (fun h => hab (finSumFinEquiv.injective h))]
+        norm_num
+
+open ComponentIdx.DropPairSection in
+private lemma contractFour_route {cA cB : Fin 4 → Color}
```
jstoobysmith, line 219: Might be nice to put this as a nonprofit lemma near ofFinEquiv

Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
```diff
@@ -113,3 +140,222 @@ lemma euclidLeviCivita_symbol_contract_two (r s t w : Fin 4) :
...
+        ((ofFinEquiv h2
+          ((ofFinEquiv h1 (fun j => j.elim0) (x0, x0)).1) (x1, x1)).1) (x2, x2)).1)
+        (x3, x3)).1
+    (ComponentIdx.prod (S := realLorentzTensor 3) (c := cA) (c1 := cB)) v =
+      (![x0, x1, x2, x3], ![x0, x1, x2, x3]) := by
+  dsimp only
+  apply Prod.ext <;> funext m <;> fin_cases m <;> rfl
+
+open ComponentIdx.DropPairSection in
+private lemma contractThree_route {cA cB : Fin 4 → Color}
```
jstoobysmith, line 234: Put near ComponentIdx.prod

Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
```diff
@@ -113,3 +140,222 @@ lemma euclidLeviCivita_symbol_contract_two (r s t w : Fin 4) :
...
+      = ∑ v : Fin 4 → X, F v := by
+  rw [← Equiv.sum_comp vec4Equiv F]
+  simp only [Fintype.sum_prod_type]
+  rfl
+
+private lemma sum3_eq {M X : Type} [Fintype X] [AddCommMonoid M] (F : (Fin 3 → X) → M) :
+    ∑ x0 : X, ∑ x1 : X, ∑ x2 : X, F ![x0, x1, x2] = ∑ h, F h := by
+  rw [← Equiv.sum_comp vec3Equiv F]
+  simp only [Fintype.sum_prod_type]
+  rfl
```
jstoobysmith, line 270: I still really think these lemmas are not needed. Most of them are just three lines. Even just including them within the actual theorems of content, I think, will produce a cleaner file.

jstoobysmith:
I think after these changes, things are starting to look a lot better and we may be able to just go PhysLib. I still think there are a couple of lemmas that you can remove, especially the private ones, but I think we're almost there. Maybe just a couple more iterations with you, if that's okay.

Robby955:
Sounds good with me, I pushed the following:

- Collapsed the flagged Kronecker-delta statement.

- Removed the redundant ofFinEquiv wrapper and used the underlying section property directly.

- Added the general public ofFinEquiv_dropPair reconstruction lemma beside ofFinEquiv.

- Removed the six private routing and finite-sum helpers from LeviCivita/Contractions.lean. The fixed-rank details now live inside the two content proofs.

Robby955:
-awaiting-author

jstoobysmith (approved):
I've approved this. Many thanks for iterating. This now looks good. Will merge shortly.

PR #1567, opened 2026-08-23
fix(tensors): correct cyclic permutation elaboration

Robby955:
## Summary

Correct index-notation permutation elaboration for non-involutive reorderings. The previous implementation produced the inverse permutation for genuine cycles; swaps masked the issue because they are self-inverse.

## Changes

- Update getPermutation to map each target slot to its source position.
- Clarify the direction of the returned permutation in its documentation.
- Add rank-three regression checks for a non-involutive three-cycle in equality and addition expressions.

## Reviewer map

1. Review the getPermutation direction change.
2. Check the two three-cycle guard messages directly below the existing elaborator tests.

## Verification

- lake build Physlib.Relativity.Tensors.Elab
- lake exe lint_all
- ./scripts/lint-style.sh
- git diff --check upstream/master...HEAD

The build, import checks, sorry/pseudo checks, Lean declaration linters, Python style linter, and targeted whitespace check pass. lint_all also reports repository-wide pre-existing advisory style and transitive-import findings.

## Risk / Rollback

This changes elaboration only where an index reordering is not self-inverse. The full build passes with the corrected direction. Reverting this commit restores the previous behavior.

### DIFF
--- a/Physlib/Relativity/Tensors/Elab.lean
+++ b/Physlib/Relativity/Tensors/Elab.lean
@@ -267,19 +267,18 @@ def contrListAdjust (l : List (ℕ × ℕ)) : List (ℕ × ℕ) :=
 
 -/
 
-/-- Given two lists of indices, all of which are indent,
-  returns the `List (ℕ)` representing the how one list
-  permutes into the other. -/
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - many thanks for catching this.

Issue #883, opened 2026-01-05
API: Configuration space for pendulum


jstoobysmith:
## Key data structure 

The key data structure is the configuration space of the pendulum. There are multiple ways that it could be defined, e.g. by an x and y coordinates which are constrained, or by its angle. One should be chosen. 

- [x] is defined 

## Need

The pendulum is a simple example of a problem in classical mechanics, this API will show the more general power of APIs around classical mechanics.

## Requirements

- [x] The API shall contain the structure of a manifold on the configuration space. 
- [x] The API shall contain a map from the configuration space to `Space`, giving the position of the pendulum in real space. 
- [x] The API shall contain the definition of a trajectory based on the configuration space 
- [x] The API shall subsequently contain the definition of the lagrangian 

## Corresponding file system

https://github.com/HEPLean/PhysLean/tree/master/PhysLean/ClassicalMechanics/Pendulum

## Parent APIs

#863

aadarwal:
All five items of this issue are now merged into `./Physlib`:

- configuration space defined — #1561 (`SimplePendulum.ConfigurationSpace`, the circle)
- manifold structure on it — #1561
- map to `Space` (position in real space) — #1561, with the trajectory-level embedding in #1571
- trajectories on the configuration space — #1571 (`SimplePendulum.Trajectory`, `Trajectory.ofLift`)
- the Lagrangian — #1564 (`SimplePendulum.lagrangian`, with the equation of motion and the variational equivalence)

Beyond the checklist, the series also landed conservation and equilibria (#1569), the Hamiltonian formulation with the equivalence of all four formulations (#1570), the small-angle theory (#1572), and `Real.completeEllipticK` (#1573); the period formula (#1576) and existence/uniqueness (#1575) are open. I think this issue can be closed whenever you're happy — the remaining identification of the period formula with the nonlinear return time is a separate project I intend to continue.

jstoobysmith:
Closing as included in code as API-map, this should be followed instead of this PR. 
Also due to @aadarwal 's comments

PR #1573, opened 2026-08-24
feat(Mathematics): Real.completeEllipticK, Legendre's complete elliptic integral of the first kind

aadarwal:
Toward #883, but **standalone**: this PR is based on `master`, depends on nothing in the pendulum
stack, and nothing in the stack is needed to review it. It is pure mathematics written to Mathlib
standards — the file header says "This file may eventually be upstreamed to Mathlib", following
`Physlib/Mathematics/SpecialFunctions/PhysHermite.lean`.

**Why here.** Mathlib has the Weierstrass `℘` (`PeriodPair.weierstrassP`) but no Legendre-form
elliptic integrals. The exact period of a simple pendulum is `4√(ℓ/g)·K(sin²(θ₀/2))`, so Physlib
needs `K` before it can state that; `Physlib/Mathematics/` is where such prerequisites live.

**Convention.** The *parameter* convention `K(m) = ∫₀^{π/2} (1 − m sin²φ)^(−1/2) dφ`
(Abramowitz & Stegun 17.3.1), not the modulus convention `K(k)` of Landau & Lifshitz; `m = k²`, and
the module doc says so.

**Domain.** Everything is stated on `m < 1`, where the radicand is positive. The module doc is
explicit that `m ≥ 1` gives a junk value and says exactly which: at `m = 1` the integrand is
`1/cos φ`, not interval integrable, so `intervalIntegral.integral_undef` returns `0` (whereas
`K(1) = ∞`); for `m > 1` the radicand goes negative past `arcsin(1/√m)`, where
`Real.rpow_def_of_neg` contributes the factor `cos(−(1/2)·π) = 0`, leaving the finite integral over
`[0, arcsin(1/√m)]` — which is `K(1/m)/√m`, the real part of the continued Legendre integral
(DLMF §19.7(ii)). Every lemma about a general parameter therefore carries `m < 1` explicitly.

| Declaration | Statement |
|---|---|
| `Real.completeEllipticK` | `def : ℝ → ℝ` (`@[pp_nodot]`), `∫ φ in 0..π/2, (1 − m sin² φ)^(−1/2)` |
| `completeEllipticK_def` | the unfolding lemma (`rfl`) |
| `completeEllipticK_radicand_pos` | `m < 1 → ∀ φ, 0 < 1 − m sin² φ` |
| `continuous_completeEllipticK_integrand`, `intervalIntegrable_completeEllipticK_integrand` | `m < 1 →` the integrand is continuous / interval integrable on `[0, π/2]` |
| `completeEllipticK_zero` | `@[simp] K 0 = π/2` |
| `completeEllipticK_pos` | `m < 1 → 0 < K m` |
| `completeEllipticK_mono`, `completeEllipticK_strictMono` (both `@[gcongr]`) | `m₁ ≤ m₂ → m₂ < 1 → K m₁ ≤ K m₂`; strict for `m₁ < m₂` |
| `monotoneOn_completeEllipticK`, `strictMonoOn_completeEllipticK` | `MonotoneOn` / `StrictMonoOn K (Set.Iio 1)` |
| `pi_div_two_le_completeEllipticK`, `completeEllipticK_le` | `0 ≤ m < 1 → π/2 ≤ K m ≤ (π/2)(1 − m)^(−1/2)` |
| `continuousOn_completeEllipticK`, `continuousAt_completeEllipticK`, `_zero` | `ContinuousOn K (Set.Iio 1)`; `m < 1 → ContinuousAt K m`; at `0` |

**Reading order.** Module doc §i (the junk-value paragraph is where the design decision lives) →
§A definition and integrand → §B `K 0 = π/2`, positivity → §C monotonicity and bounds → §D
continuity. §D is the only non-obvious proof: the integrand is not jointly continuous in `(m, φ)`
on `ℝ × ℝ`, so the parameter is restricted to the subtype `Set.Iio 1`, where joint continuity is
global; `intervalIntegral.continuous_parametric_intervalIntegral_of_continuous'` then applies and
`continuousOn_iff_continuous_domRestrict` converts back.

**Verification.** Module builds zero-warning against Mathlib v4.33.0; no `sorry`, no `set_option`,
no linter exemptions; full `lake build` + Physlib linter battery pass; `lint-style` clean, TOC 1:1;
single import (`Mathlib.MeasureTheory.Integral.DominatedConvergence`) — shake-clean; `K` is *strictly*
increasing on `(-∞, 1)` and sandwiched there by `π/2 ≤ K m ≤ (π/2)(1 − m)^(−1/2)` for `0 ≤ m < 1`; no deprecated
names; `Set.Iio 1` is the maximal honest domain (`K` is genuinely discontinuous at `1`).

Developed with assistance from AI; all mathematics and proofs were reviewed and verified to compile.

### DIFF
--- a/Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
+++ b/Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
@@ -25,6 +25,14 @@ conventions are related by `m = k²`. Mathlib knows the Weierstrass elliptic fun
 complete integral of the first kind and develops its basic theory on the domain `m < 1`, where
 the radicand `1 - m sin² φ` is positive and the integrand continuous.
 
+The integral enters physics wherever a period or a potential is computed exactly rather than in a
+small-parameter expansion. Physlib's use of it so far is the simple pendulum: released from rest
+at amplitude `θ₀`, the pendulum has period `4 √(ℓ / g) K(sin² (θ₀ / 2))` (Landau & Lifshitz §11,
+Problem 1), the consumer being `Physlib.ClassicalMechanics.Pendulum.SimplePendulum.PeriodFormula`.
+More broadly, the complete integrals of the first and second kind give the magnetic field of a
+circular current loop and the potential of a uniformly charged ring, and the second kind, `E`,
+gives the arc length of an ellipse. This file defines `K` so that such results can be stated.
+
 For `m ≥ 1` the definition still elaborates, but its value is not Legendre's. At `m = 1` the
 integrand is `1 / cos φ`, which is not interval integrable on `[0, π/2]`, so the integral is `0`
 by `intervalIntegral.integral_undef`, whereas `K(1) = ∞`. For `m > 1` the radicand is negative on
--- a/Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
+++ b/Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
@@ -6,7 +6,6 @@ Authors: Aadarsh Agarwal
 module
 
 public import Mathlib.MeasureTheory.Integral.DominatedConvergence
-public import Mathlib.MeasureTheory.Integral.IntervalIntegral.Basic
 /-!
 
 # The complete elliptic integral of the first kind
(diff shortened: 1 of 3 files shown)


Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
```diff
@@ -0,0 +1,263 @@
...
+  rw [← sub_pos, ← cos_sq']
+  exact pow_pos (cos_pos_of_mem_Ioo ⟨by linarith, h₂⟩) 2
+
+/-- The half-angle form of `sin_sq_lt_one`: `sin (θ / 2) ^ 2 < 1` for `|θ| < π`. For the pendulum
+this says that the parameter `sin² (θ₀ / 2)` of the period formula lies in the domain of
+`completeEllipticK` for every libration amplitude `|θ₀| < π`. -/
+lemma sin_half_sq_lt_one {θ : ℝ} (h : |θ| < π) : sin (θ / 2) ^ 2 < 1 := by
+  refine sin_sq_lt_one ?_
+  rw [abs_div, abs_two]
+  linarith [abs_nonneg θ]
```
jstoobysmith: if we can get away with having these as separate lemmas, I think that would be good since they're not really related to the content of this file.

Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
```diff
@@ -0,0 +1,263 @@
...
+For `m ≥ 1` the definition still elaborates, but its value is not Legendre's. At `m = 1` the
+integrand is `1 / cos φ`, which is not interval integrable on `[0, π/2]`, so the integral is `0`
+by `intervalIntegral.integral_undef`, whereas `K(1) = ∞`. For `m > 1` the radicand is negative on
+`(arcsin (1/√m), π/2]`, where the real power at exponent `-(1/2)` of a negative base vanishes
+(`Real.rpow_def_of_neg` supplies the factor `cos (-(1 / 2) * π) = 0`), so the Lean value is the
+finite positive integral over `[0, arcsin (1/√m)]`; by the reciprocal-modulus transformation this
+is `K(1/m) / √m`, the real part of the complex Legendre integral (DLMF §19.7(ii)) — not proved
+here. Every lemma of this file about a general parameter therefore carries its domain hypothesis
+`m < 1` explicitly.
+
```
jstoobysmith, line 45: Despite being a math file, I think it is nice to have the physics here as well, basically saying, in a bit more detail, where these definitions from math are used in physics and why we're defining them.

Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
aadarwal: Done in `b302d55f` — moved to a new `Physlib/Mathematics/Trigonometry/SinSq.lean` (`Real.sin_sq_lt_one`, `Real.sin_half_sq_lt_one`). `EllipticIntegral.lean` no longer contains them; #1576 imports them from there.

Physlib/Mathematics/SpecialFunctions/EllipticIntegral.lean
aadarwal: Done in `b302d55f` — §i now says where `K` is used: the pendulum period `4√(ℓ/g) K(sin²(θ₀/2))` (L&L §11 Problem 1; the consumer is `SimplePendulum.periodFormula` in #1576), plus the other classical places `K` and `E` turn up — the field of a current loop, the potential of a charged ring, the arc length of an ellipse.

jstoobysmith (approved):
Approved this PR. Many thanks

PR #1578, opened 2026-08-24
QuantumInfo: StdBasis refactor

Timeroot:
This is a very large refactor, but it's also the kind of thing that made sense to do in one big sweep, and by nature it touches nearly everything in QuantumInfo.

The idea is to:
 * Change density matrix (and bra, and ket) to stated in terms of hermitian operators, instead of matrices. Similarly for CPTPMap and others. (CPTPOp in general, for working on operators; CPTPMap for when it's on matrices.)
 * Have a `StdBasis` instance that gives a "canonical" basis for much work, and a `.toMatrix` / `.ofMatrix` equivalence with the matrix form wherever that instance is available. This is heavily connected to `EuclideanSpace` as the space given on that basis then.
 * Do definitions without StdBasis where possible (a basis-free proof), but one can also do them with a basis and then show they are basis-invariant. Both can be natural at times.
 
 I think this will greatly increase the compatibility of the library with other efforts, down the line!


QuantumInfo/ForMathlib/StdBasis.lean
```diff
@@ -0,0 +1,588 @@
...
+  simp
+
+section Unitary
+
+omit [FiniteDimensional 𝕜 E]
+
+variable [StdBasis 𝕜 E ι]
+
+/-- The matrix of a unitary operator in the preferred basis, as a unitary matrix. -/
+noncomputable def toMatUnitary (U : unitary (E →L[𝕜] E)) : Matrix.unitaryGroup ι 𝕜 :=
```
goliath-klein, line 267: One could also use existing constructions, like so:

`noncomputable def toMatUnitary' (U : unitary (E →L[𝕜] E)) : Matrix.unitaryGroup ι 𝕜 :=
  ⟨(Unitary.linearIsometryEquiv U).toMatrix stdBasis.toBasis stdBasis.toBasis,
      LinearIsometryEquiv.toMatrix_mem_unitaryGroup ..⟩`

A related def from [QCLib](https://github.com/QuCompLean/QCLib/blob/dcf88ffc7a17e36419460952c66adeabb18b250e/QCLib/LinearAlgebra/Unitary.lean#L56), which gives a star algebra equivalence:

`noncomputable def unitaryGroupEquiv :
    unitaryGroup n 𝕜 ≃⋆* unitary ((EuclideanSpace 𝕜 n) →L[𝕜] (EuclideanSpace 𝕜 n)) :=
  Unitary.mapEquiv (StarMulEquiv.ofClass (Matrix.toEuclideanCLM (𝕜 := 𝕜)))`

goliath-klein, line 267: One could use the namespace `_root_.Subtype.toMatUnitary` to enable dot notation.

QuantumInfo/ForMathlib/HermitianOp.lean
```diff
@@ -0,0 +1,549 @@
...
+
+@[expose] public section
+
+open scoped ComplexOrder
+
+/-- The type of self-adjoint continuous linear operators on `E`, as a `Subtype`.
+
+This is the basis-free analogue of `HermitianMat`; see `HermitianOp.matEquiv` for the
+identification of the two given a preferred orthonormal basis. -/
+def HermitianOp (E : Type*) [NormedAddCommGroup E] [InnerProductSpace ℂ E] [CompleteSpace E] :=
```
goliath-klein, line 58: I know this came up on Zulip, and you were a bit exasperated recalling the story of this type synonym. So sorry for brining it up again...

...but why is this a definition (semi-reducible, even)? Could one not just introduce a custom notation for the somewhat lengthy `selfAdjoint (E →L[ℂ] E)`? This would have the advantage of making much of the boilerplate that follows redundant (instances, like `AddCommGroup`; simp lemmas like `op_zero`, `op_add`, ...). It would also remove one discrepancy between the Physlib and the Mathlib way of saying things.

You mentioned performance issues on Zulip. Is that the only reason? If so, I'd be happy to offer help looking into their origin. Maybe a few shortcut instances can address it?

QuantumInfo/ForMathlib/HermitianOp.lean
```diff
@@ -0,0 +1,549 @@
...
+/-- A self-adjoint operator is nonnegative exactly when it is a positive operator. -/
+theorem zero_le_iff {A : HermitianOp E} : 0 ≤ A ↔ A.op.IsPositive :=
+  ContinuousLinearMap.nonneg_iff_isPositive A.op
+
+/-- The trace of a self-adjoint operator. It is real because the operator is self-adjoint. -/
+noncomputable def trace [FiniteDimensional ℂ E] (A : HermitianOp E) : ℝ :=
+  RCLike.re (LinearMap.trace ℂ E (A.op : E →ₗ[ℂ] E))
+
+/-- The continuous functional calculus applied to a self-adjoint operator. -/
+noncomputable def cfc (A : HermitianOp E) (f : ℝ → ℝ) : HermitianOp E :=
```
goliath-klein, line 116: One could make this definition much more general. E.g. like this:

```
noncomputable def selfAdjoint.cfc {R : Type*} [TopologicalSpace R] [Ring R] [StarRing R]
  [Algebra ℂ R] [ContinuousFunctionalCalculus ℂ R IsStarNormal] 
  (A : selfAdjoint R) (f : ℝ → ℝ) : selfAdjoint R :=
  ⟨_root_.cfc f A.val, cfc_predicate f A.val⟩
```

This def would cover both your intended use case `selfAdjoint (E →L[ℂ] E)`, but also self-adjoint elements in a `C^*`-algebra.

QuantumInfo/ForMathlib/HermitianOp.lean
```diff
@@ -0,0 +1,549 @@
...
+  ⟨_root_.cfc f A.op, cfc_predicate _ _⟩
```
goliath-klein, line 117: `⟨_root_.cfc f A.op, cfc_predicate f A.op⟩`: more information at essentially the same length.

(Side remark: I'm amazed the elaborator can figure out the predicate (`IsSelfAdjoint`) in the first argument of the constructor. It took me some time to understand what the definition does. So it's maybe not the best place to hide information).

zblore:
Heads up #1655 merged yesterday and removed four files this refactor edits: Entropy/Axiomatized/Defs and Renyi, Regularized and ResourceTheory/ResourceTheory, and rewrote the imports in QuantumInfo.lean. They didn't build against the current pin. If you want them back they're one checkout away in the history.

PR #1579, opened 2026-08-25
feat(Units): reducible rational arithmetic via Exponent

RaunakChhatwal:
# Reducible rational arithmetic for dimension exponents

Ideally, the product of speed and time should type-check as length. This does not happen at present because the product of their dimensions is not definitionally equal to the length dimension:

```lean
example (v : WithDim (L𝓭 / T𝓭) ℝ) (t : WithDim T𝓭 ℝ) : WithDim L𝓭 ℝ :=
  v * t -- Type mismatch `v * t` has type `WithDim (L𝓭 / T𝓭 * T𝓭) ℝ` but is expected to have type `WithDim L𝓭 ℝ` (Lean 4)
```

The root cause is that dimension arithmetic uses rational arithmetic, whose operations are irreducible in Lean:

```lean
example : (2 : ℕ) + 2 = 4 := rfl -- succeeds
example : (2 : ℚ) + 2 = 4 := rfl -- fails
```

Locally unsealing rational arithmetic does not export reducibility to downstream modules, while globally changing the reducibility of imported declarations requires allowUnsafeReducibility. This PR instead introduces a wrapper around ℚ with reducible arithmetic, enabling:

```lean
example :
  let Length : Exponent × Exponent := (1, 0)
  let Time : Exponent × Exponent := (0, 1)
  let Speed := Length - Time
  Length = Time + Speed := rfl
```

This enables definitional equality for tuple- or structure-based dimensions, but unfortunately not for Physlib's current parametric dimensions. Lean's definitional equality is less effective at comparing unapplied functions than concrete data structures, while the parametric approach relies on function representation to maintain basis independence. I would therefore like to follow this PR with a simpler, non-parametric formalization of dimensions using Exponent.

### DIFF
--- a/Physlib.lean
+++ b/Physlib.lean
@@ -540,8 +540,8 @@ public import Physlib.Thermodynamics.Temperature.Basic
 public import Physlib.Thermodynamics.Temperature.TemperatureUnits
 public import Physlib.Units.Basic
 public import Physlib.Units.Dimension
-public import Physlib.Units.Exponent
 public import Physlib.Units.Examples
+public import Physlib.Units.Exponent
 public import Physlib.Units.FDeriv
 public import Physlib.Units.ISQBridge
 public import Physlib.Units.ISQDimensionBase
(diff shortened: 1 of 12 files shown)


jstoobysmith (approved):
Looks good to me. Approved.

PR #1581, opened 2026-08-26
docs(Relativity): correct dimension in Levi-Civita contraction module docstring

### LINKED ISSUE #1533: Documentation error in Levi-Civita tensor description
> **⚠️ Found by a local LLM, not a human.** This was flagged by an automated documentation-mistake finder (`qwen3:14b`) and passed 10/10 independent verification runs before filing, but it has **not** been reviewed by a person. Please check it's a genuine error before acting on it.

---

### Summary

The documentation in [`Physlib/Relativity/Tensors/LeviCivita/Contractions.lean`](https://github.com/leanprover-community/physlib/blob/master/Physlib/Relativity/Tensors/LeviCivita/Contractions.lean#L17-L18) (around line 17) is incorrect.

### Current text

```lean
the rank-four Levi-Civita tensor `leviCivita` (notation `ε4`) in `d = 3`
```

### Why this is wrong

The documentation claims that `ε4` represents a rank-four Levi-Civita tensor in 3-dimensional space. This is incorrect because the rank of a Levi-Civita tensor must match the dimension of the space it inhabits. A rank-four Levi-Civita tensor can only exist in 4-dimensional space, not 3.

The correction changes the description to refer to a rank-three Levi-Civita tensor in 4-dimensional space (`d = 4`), which aligns with the code's use of `Fin 4` indices and the mathematical properties being proved. The file's lemmas (e.g., `euclidLeviCivita_symbol_contract_zero`) operate on 4-index tensors, confirming the context is 4D space.

The error arose from a mismatch between the tensor's rank, the space dimension, and the notation used in the code, which consistently refers to 4-dimensional indices via `Fin 4`.

### Suggested correction

```lean
the rank-three Levi-Civita tensor `leviCivita` (notation `ε4`) in `d = 4`
```

oe-parks:
Corrects docstring for Levi-Civita contraction identities, which described `ε4` as living in `d = 3`.

The file’s results are in four dimensions. The full contraction is 24 = 4!, and the components are indexed by `Fin 4`.

Per @jstoobysmith’s comments on the issue, only the dimension was wrong.

Closes #1533

First contribution to Physlib so I figured I’d stick to docstrings. Happy to adjust anything.

### DIFF
--- a/Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
+++ b/Physlib/Relativity/Tensors/LeviCivita/Contractions.lean
@@ -15,7 +15,7 @@ public import Physlib.Meta.TODO.Basic
 ## i. Overview
 
 This file proves the "epsilon-epsilon" contraction identities for the rank-four Levi-Civita
-tensor `leviCivita` (notation `ε4`) in `d = 3`, stated in terms of the standard-basis
+tensor `leviCivita` (notation `ε4`) in `d = 4`, stated in terms of the standard-basis
 components of `ε4` itself (`realLorentzTensor.leviCivita_basis_repr_apply`).
 
 The underlying facts about the `generalizedKroneckerDelta` alone, with no


jstoobysmith (approved):
Looks great! And a nice place to start your Physlib journey :). 

I've approved - once the workflows have finished, I will merge!

PR #1582, opened 2026-08-27
feat: Lake cache workflow

Alex-Zughaid:


docs/cache-setup.md
```diff
@@ -0,0 +1,57 @@
...
+storage, with lots of people downloading it.
+
+## 1. Create the bucket
+
+Call it `physlib-cache`. Add its S3 API endpoint into `lake-cache.toml` as the write path.
+
+## 2. Allow anonymous reads — done
+
+Contributors fetch without credentials, so the bucket needs a Public Development URL.
+We can use the provided '.......r2.dev' domain for testing, but it is better to have a
```
jstoobysmith: Can we reframe this, and parts like this. I don’t think the framing in terms of “testing” is good

jstoobysmith (approved):
Approved - will merge now.

PR #1591, opened 2026-08-30
Speed up cache build workflow

Alex-Zughaid:
- Use the previous cache (if possible) when trying to build the next cache


jstoobysmith (approved):
Approved. Will merge shortly

PR #1604, opened 2026-09-01
refactor(ClassicalMechanics): switch to derivVec for simple-pendulum velocity

zhikaip:
`S.spaceTrajectory θ : Time → Space 2` is the bob's position, a curve of points, so
differentiating it with `Time.deriv` gave a velocity typed as a *point* of `Space 2` —
the docstring described it as "the vector tangent to the circle" while the type said
otherwise. This retypes it with `derivVec` (#1592), so the velocity is valued in
`EuclideanSpace ℝ (Fin 2)`.

Nothing changes numerically: both norms are the L² norm on coordinates.

AI disclosure: assisted in checking and PR summary, code is written by myself


morrison-daniel (approved):
Looks good!

PR #1605, opened 2026-09-01
EVJJI theorem twoState_entropy_eq

oe-parks:
Formalizes the informal lemma tagged `EVJJI`: a closed form for the
thermodynamic entropy of the two-state canonical ensemble.

Added, both in `Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean`:

- `twoState_entropy_eq` with `x = β T * (E₁ - E₀) / 2`, gives
  `S = kB * (log (2 * cosh x) - x * tanh x)`. Proved through
  `thermodynamicEntropy_eq_shannonEntropy`, which carries no side conditions,
  rather than through the Helmholtz free energy. That route needs
  `0 < T.val`, `IsFiniteMeasure`, and integrability.

- `twoState_entropy_eq_T_neq_zero` is the same statement carrying `T ≠ 0`, so
  downstream results on positive temperatures get a matching signature without
  having to know the unconditional version is safe at zero.

Removed the `Physlib.Meta.Informal.Basic` import, now unused in the file.

I left the primary result unconditional, matching the choice made for
`twoState_helmholtzFreeEnergy_eq` in #1474. Since `β 0 = 0`, the RHS is
`kB * log 2` at `T = 0`; the docstring notes this.

Notes to reviewer: look first at the statement, since that's the part that has to be right.
Then `hp0` and `hp1`, the two occupation probabilities in closed form. The rest
is algebra. The corollary and the import removal are a line each.

Second contribution to Physlib. Happy to adjust anything. I'm not particularly attached to the `T ≠ 0` restatement if you'd rather it weren't there.

### DIFF
--- a/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
+++ b/Physlib/StatisticalMechanics/CanonicalEnsemble/TwoState.lean
@@ -6,7 +6,6 @@ Authors: Matteo Cipollina, Joseph Tooby-Smith
 module
 
 public import Physlib.StatisticalMechanics.CanonicalEnsemble.Finite
-public import Physlib.Meta.Informal.Basic
 /-!
 
 # Two-state canonical ensemble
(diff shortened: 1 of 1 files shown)


jstoobysmith (approved):
Approved - looks good to me. Thanks for this PR

PR #1607, opened 2026-09-01
feat(OperatorAlgebra): unitary one-parameter group automorphisms and Hamiltonian dynamics

TomOleDiem:
Adds the automorphism-group action of a norm-continuous unitary one-parameter group on `B(H)` by conjugation, proves it satisfies (and is the unique solution of) the Heisenberg equation of motion, and specializes this to Hamiltonian dynamics `U_H(t) = exp(-itH/ℏ)`, including the classification of when two Hamiltonians generate the same flows.

Split out of #1590.

With the Discussion David Gross opened I expect some of these notions to be reworked in the future.
I wanted to get the code out there but can also put it into PhyslibAlpha if that is more appropriate.


jstoobysmith (approved):
Approved

PR #1608, opened 2026-09-02
feat(OperatorAlgebra): reorganize states, observables, and measurement

TomOleDiem:
Defined the first notions for Observable & State and how they together produce a probability measure.

- `Observables/Basic.lean`, `Positive.lean`, `Effects.lean`, `Jordan.lean`, `Lie.lean`, `Projection.lean`
- `States/Basic.lean`, `Convex.lean`, `Pure.lean`, `Statistics.lean`, `Uncertainty.lean`
- `Measurement/SpectralMeasure.lean`

With the Discussion David Gross opened I expect some of these notions to be reworked in the future.
I wanted to get the code out there but can also put it into PhyslibAlpha if that is more appropriate.


jstoobysmith (approved):
Approved

PR #1610, opened 2026-09-02
feat(Lorentz.Vector): mark Vector and CoVector implicit_reducible, add inner_eq_sum

zhikaip:
`Lorentz.Vector d` and `Lorentz.CoVector d` are defs for `Fin 1 ⊕ Fin d → ℝ` and are applied directly as functions throughout the library (v μ, x (Sum.inl 0), …). Lean 4.33 enables `backward.isDefEq.respectTransparency.types` by default, under which such applications are not type-correct at implicit transparency (the note "function expected v μ" in error messages). Once a goal contains one under a binder, `rw` and `simp` can fail to match patterns that are visibly present. The release notes recommend marking the definitions involved `@[implicit_reducible]`; this PR does that for both types.

`Lorentz.Vector.inner_eq_sum` is added to simplify proofs previously going through equivEuclid, WithLp and PiLp.

as a result, all `set_option backward.isDefEq.respectTransparency false` and occurences of `erw` in these files can be removed 

PR drafted by claude and finalised by myself

### DIFF
--- a/Physlib/Relativity/Tensors/RealTensor/Vector/Basic.lean
+++ b/Physlib/Relativity/Tensors/RealTensor/Vector/Basic.lean
@@ -125,7 +125,7 @@ instance innerProductSpace (d : ℕ) : InnerProductSpace ℝ (Vector d) where
 /-- The inner product on `Vector d` as a sum over components. -/
 lemma inner_eq_sum {d : ℕ} (v w : Vector d) : ⟪v, w⟫_ℝ = ∑ μ, v μ * w μ := by
   rw [inner_eq_equivEuclid, PiLp.inner_apply]
-  simp [-Fintype.sum_sum_type, mul_comm]
+  simp [mul_comm]
 
 /-- The instance of a `ChartedSpace` on `Vector d`. -/
 instance : ChartedSpace (Vector d) (Vector d) := chartedSpaceSelf (Vector d)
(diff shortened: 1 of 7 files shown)


Physlib/Relativity/Tensors/RealTensor/Vector/Basic.lean
```diff
@@ -31,6 +31,12 @@ namespace Lorentz
 /-- Real contravariant Lorentz vector. -/
 def Vector (d : ℕ := 3) := Fin 1 ⊕ Fin d → ℝ
 
+/- `Vector d` is applied directly as a function throughout the library. Marking it
+  implicit-reducible lets such applications typecheck at implicit transparency, so that
+  `rw` and `simp` can match patterns containing them (see the Lean 4.33 release notes on
+  `backward.isDefEq.respectTransparency.types`). -/
+attribute [implicit_reducible] Vector
```
jstoobysmith: Would it not be better to do 
``` Lean4
@[implicit_reducible] 
def Vector .... 
```

zhikaip: fixed, wondering if it's a good idea to keep the explanation in or remove them

jstoobysmith: Think it would be nice to keep it somewhere

jstoobysmith (approved):
Approved

PR #1611, opened 2026-09-02
fix(Tensors): rank the tensor action above the Tensorial action

zhikaip:
## Changes

Tensors/Basic.lean: `instSMul`, `actionT` and the `DistribMulAction` on `S.Tensor c` get priority := high + 1.

`SMulCommClass k G (S.Tensor c)` and `SMulCommClass G k (S.Tensor c)` added (the second via `SMulCommClass.symm`, as in Mathlib); the G k order also added for Tensorial types.

as a result, multiple `set_option backward.isDefEq.respectTransparency false` removed and a `maxHeartbeats 600000`.

## Why: an instance diamond on `g • t`

Two `SMul G (S.Tensor c)` instances exist:

1. `Tensor.instSMul` in `Tensors/Basic.lean`, the direct action. All `*_equivariant` lemmas are
   stated with it.
2. `Tensorial.smulAction` in `Tensorial.lean`, defined for any `[Tensorial S c M]` as
   `g • m = toTensor.symm (g • toTensor m)`. Since `S.Tensor c` is `Tensorial` via `self`
   (`toTensor = LinearEquiv.refl`), it also applies to bare tensors, and with `priority := high`
   it won.

So in a goal, `g • t` carried instance 2; the lemma's left-hand side carries instance 1. They are
defeq (unfold `refl`) but not syntactically equal, hence `rw`/`simp` failed while `exact` and
the `respectTransparency false` option (unfolding fallback, paid for in heartbeats) worked.

Fix: rank instance 1 at `high + 1`. Bare tensors now use the direct action. Instance 2 keeps
`high` and still governs `Tensorial` types and `M ⊗ M₂`, where it must beat Mathlib's left-only
`TensorProduct` action (`Tensorial.smul_prod`); lowering it instead breaks that lemma.
The `SMulCommClass` instances are stated for instance 1 so `smul_comm` fires on it.

fix found by claude and confirmed/edited by myself

### DIFF
--- a/Physlib/Electromagnetism/Dynamics/IsExtrema.lean
+++ b/Physlib/Electromagnetism/Dynamics/IsExtrema.lean
@@ -171,7 +171,6 @@ the speed with which an electromagnetic wave propagates is invariant under Loren
 
 -/
 
-set_option maxHeartbeats 600000 in
 lemma isExtrema_lorentzGroup_apply_iff {𝓕 : FreeSpace}
     (A : ElectromagneticPotential d)
     (hA : ContDiff ℝ ∞ A) (J : LorentzCurrentDensity d) (hJ : ContDiff ℝ ∞ J)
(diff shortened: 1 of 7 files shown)


jstoobysmith (approved):
Looks good to me. Approved.

PR #1612, opened 2026-09-03
feat(ClassicalMechanics): Newtonian point-particle systems

RaunakChhatwal:
## Summary

This PR introduces basic definitions for forces and Newtonian point-particle systems in classical mechanics.

### Forces

Adds `Force`, representing a time-dependent force acting on an object, together with:

* `InternalForce`, which additionally records the source of a force
* `InternalForce.reverse`, representing the equal-and-opposite force with source and target exchanged
* `netForce`, which sums the internal and external forces acting on a given object

### Point particles

Adds `Particle`, consisting of:

* positive mass
* a trajectory relative to a reference frame
* twice-differentiability of the trajectory in inertial frames

and defines the associated:

* velocity
* acceleration
* position in affine absolute space

Particles and forces are indexed by the reference frame in which they are described. This makes the frame dependence explicit in the types and prevents accidentally combining positions, velocities, accelerations, or forces expressed in different frames.

### Point-particle systems

Adds `PointParticle.System`, representing a finite collection of particles in an inertial reference frame, together with internal and external forces.

Newton's laws are encoded as properties of a system:

* **Newton's second law:** the net force on each particle equals its mass times its acceleration
* **Newton's third law:** internal forces come in equal-and-opposite pairs

The PR also introduces convenient system-level definitions for:

* particle mass, position, velocity, acceleration, momentum, and kinetic energy
* internal/external system forces
* reverse and central internal forces
* total mass
* total momentum
* net external force
* total kinetic energy

Finally, the new `Force` and `PointParticle.Defs` modules are exported from `Physlib.lean`.

### DIFF
--- a/Physlib/ClassicalMechanics/Force.lean
+++ b/Physlib/ClassicalMechanics/Force.lean
@@ -1,0 +1,68 @@
+/-
+Copyright (c) 2026 Raunak Chhatwal. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Raunak Chhatwal
+-/
+module
+
+public import Mathlib.Data.Multiset.Fintype
+public import Physlib.SpaceAndTime.ReferenceFrame
+/-!
+# Forces
+
+This module defines time-dependent forces in a reference frame.
+-/
+
+@[expose] public noncomputable section
+
+open scoped BigOperators Classical
+
+namespace ClassicalMechanics.ReferenceFrame
+
+variable {d : ℕ} {frame : ReferenceFrame d} {Object : Type}
+
+/-!
+## A. Individual forces
+-/
+
+/-- A time-dependent force acting on an object. -/
+structure Force (frame : ReferenceFrame d) (Object : Type) where
+  /-- The force vector. -/
+  value : Time → frame.Vector
+  /-- The target object. -/
+  target : Object
+
+instance : CoeFun (frame.Force Object) (fun _ => Time → frame.Vector) where
+  coe := Force.value
+
+/-- A force between two objects. -/
+structure InternalForce (frame : ReferenceFrame d) (Object : Type) extends frame.Force Object where
+  /-- The source object. -/
+  source : Object
+  source_ne_target : source ≠ target
+
+instance : CoeFun (frame.InternalForce Object) (fun _ => Time → frame.Vector) where
+  coe force := force.value
+
+instance : Coe (frame.InternalForce Object) (frame.Force Object) where
+  coe := InternalForce.toForce
+
+/-- The equal-and-opposite force with source and target exchanged. -/
+def InternalForce.reverse (force : frame.InternalForce Object) : frame.InternalForce Object where
+  value := -force.value
+  target := force.source
+  source := force.target
+  source_ne_target := force.source_ne_target.symm
+
+/-!
+## B. Resultant force
+-/
+
+/-- The net force on `object`. -/
+def netForce
+    (object : Object)
+    (internalForces : Multiset (frame.InternalForce Object))
+    (externalForces : Multiset (frame.Force Object))
+    (t : Time) : frame.Vector :=
+  let forces := internalForces.map InternalForce.toForce + externalForces
+  ∑ force : forces with force.1.target = object, force.1 t
--- a/Physlib/ClassicalMechanics/Force.lean
+++ b/Physlib/ClassicalMechanics/Force.lean
@@ -10,7 +10,27 @@ public import Physlib.SpaceAndTime.ReferenceFrame
 /-!
 # Forces
 
-This module defines time-dependent forces in a reference frame.
+This module defines forces acting on objects, expressed in a reference frame.
+A `Force` records its target and its vector value over time. An `InternalForce`
+additionally records a distinct source object.
+
+The time-dependent vector describes the force acting on the target, without
+prescribing how that force is determined. For example, a gravitational force can
+be specified directly from the target's mass, while a constraint force can be
+specified by its relation to the motion of the objects. Representing both by the
+same type allows Newton's laws to be stated independently of the particular
+interactions present.
+
+The reference frame appears in the type of a force's vector values, keeping their
+coordinate dependence explicit. The definition lives in the
+`ClassicalMechanics.ReferenceFrame` namespace to support dot notation such as
+`frame.Force Object`, but also occupies a general name that future non-particle
+Newtonian formalizations may need.
+
+The target type `Object` is therefore deliberately left arbitrary. Newtonian mechanics is
+not limited to point particles, so this force representation is kept independent
+of any particular model of matter, with the intention it may get generalized for
+rigid body mechanics, continuum mechanics, or other Newtonian models in the future.
 -/
 
 @[expose] public noncomputable section
@@ -20,10 +40,6 @@ open scoped BigOperators Classical
 namespace ClassicalMechanics.ReferenceFrame
 
 variable {d : ℕ} {frame : ReferenceFrame d} {Object : Type}
-
-/-!
-## A. Individual forces
--/
 
 /-- A time-dependent force acting on an object. -/
 structure Force (frame : ReferenceFrame d) (Object : Type) where
@@ -53,10 +69,6 @@ def InternalForce.reverse (force : frame.InternalForce Object) : frame.InternalF
   target := force.source
   source := force.target
   source_ne_target := force.source_ne_target.symm
-
-/-!
-## B. Resultant force
--/
 
 /-- The net force on `object`. -/
 def netForce
--- a/Physlib/ClassicalMechanics/Force.lean
+++ b/Physlib/ClassicalMechanics/Force.lean
@@ -71,10 +71,7 @@ def InternalForce.reverse (force : frame.InternalForce Object) : frame.InternalF
   source_ne_target := force.source_ne_target.symm
 
 /-- The net force on `object`. -/
-def netForce
-    (object : Object)
-    (internalForces : Multiset (frame.InternalForce Object))
-    (externalForces : Multiset (frame.Force Object))
-    (t : Time) : frame.Vector :=
+def netForce (object : Object) (internalForces : Multiset (frame.InternalForce Object))
+    (externalForces : Multiset (frame.Force Object)) (t : Time) : frame.Vector :=
   let forces := internalForces.map InternalForce.toForce + externalForces
   ∑ force : forces with force.1.target = object, force.1 t
--- a/Physlib/ClassicalMechanics/PointParticle/Basic.lean
+++ b/Physlib/ClassicalMechanics/PointParticle/Basic.lean
@@ -45,16 +45,6 @@ variable {d : ℕ} {frame : ReferenceFrame d}
 /-- Positive real numbers. -/
 notation "ℝ+" => {x : ℝ // 0 < x}
 
-/-- Scalar multiplication by a positive real. -/
-instance : SMul ℝ+ frame.Vector where
-  smul c x := c.val • x
-
-/-!
-## A. Point particles
-
-A point particle has a positive mass and a trajectory relative to a reference frame.
--/
-
 /-- A point particle in `frame`. -/
 structure Particle (frame : ReferenceFrame d) where
   /-- The particle's mass. -/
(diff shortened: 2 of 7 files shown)


Physlib/ClassicalMechanics/PointParticle/Basic.lean
```diff
@@ -0,0 +1,241 @@
...
+@[expose] public noncomputable section
+
+open scoped BigOperators Classical
+
+namespace ClassicalMechanics.ReferenceFrame
+
+variable {d : ℕ} {frame : ReferenceFrame d}
+
+/-- Positive real numbers. -/
+notation "ℝ+" => {x : ℝ // 0 < x}
```
jstoobysmith, line 46: We should check, but I think we might already have PosReals defined somewhere

Physlib/ClassicalMechanics/PointParticle/Defs.lean
```diff
@@ -0,0 +1,241 @@
...
+namespace Particle
+
+variable (particle : frame.Particle)
+
+/-- Position is differentiable in an inertial frame. -/
+instance [h : Fact frame.IsInertial] : Fact (Differentiable ℝ particle.pos) :=
+  ⟨particle.pos_twice_differentiable h.out |>.left⟩
+
+/-- The particle's velocity. -/
+def vel [_h : Fact (Differentiable ℝ particle.pos)] : Time → frame.Vector :=
```
jstoobysmith: Would rename to `velocity`

Physlib/ClassicalMechanics/PointParticle/Basic.lean
```diff
@@ -0,0 +1,241 @@
...
+  smul c x := c.val • x
+
+/-!
+## A. Point particles
+
+A point particle has a positive mass and a trajectory relative to a reference frame.
+-/
+
+/-- A point particle in `frame`. -/
+structure Particle (frame : ReferenceFrame d) where
```
jstoobysmith, line 49: Would copy the structure of QuantumHarmonicOscillator here. Let `Particle` contain only the mass (definning the system). Then define `ConfigurationSpace` for the particles position. Then trajectories are maps from `Time` to `ConfigurationSpace`.

Physlib/ClassicalMechanics/PointParticle/Defs.lean
```diff
@@ -0,0 +1,241 @@
...
+open ReferenceFrame
+
+variable {d : ℕ}
+
+/-!
+## B. Systems
+-/
+
+/-- A finite system of point particles satisfying Newton's laws. -/
+structure System (d : ℕ) where
```
jstoobysmith: Would rename this `ManyPointParticles` or similar, and give it its own directory.

jstoobysmith (commented):
Gave some high level comments to start. I think you should look at the other classical mechanics files to see how they have been done.

Physlib/ClassicalMechanics/PointParticle/Basic.lean
RaunakChhatwal: I don't think there is one. For instance, [Positive/Ring.lean](https://github.com/leanprover-community/mathlib4/blob/30a58f795ae8c95b8299fe3c9ded1e6498a9e9fe/Mathlib/Algebra/Order/Positive/Ring.lean) uses `{x : R // 0 < x}` directly. I did narrow the SMul instance for R+ to frame.Vector to minimize redundancy.

Physlib/ClassicalMechanics/PointParticle/Defs.lean
RaunakChhatwal: Done, also renamed acc. Though I would like to keep `pos` abbreviated

Physlib/ClassicalMechanics/PointParticle/Basic.lean
RaunakChhatwal: `Particle` is supposed to represent the properties (i.e. mass and position over time) of an actual particle. It is not a dynamical system by itself, and especially not the unconstrained one-particle lagrangian system. It's meant to constitute Newtonian particle systems in [NewtonianSystem/Defs.lean](https://github.com/leanprover-community/physlib/pull/1612/changes#diff-dfda5b11c6385c42d98e2b0f1d9de0b7c8766fb1976730681850a3b0eb223874R37), and I don't think configuration spaces should be involved in a foundation for newtonian mechanics

Physlib/ClassicalMechanics/PointParticle/Defs.lean
RaunakChhatwal: I would not include "PointParticle" in the name, since that is already the namespace. `PointParticle.System` should therefore be read as "point-particle system". I moved it to PointParticle/NewtonianSystem/Defs.lean, which is a subfolder of PointParticle because this formalization only applies to point particles and not rigid bodies, continuums, etc.

Physlib/ClassicalMechanics/PointParticle/NewtonianSystem/Basic.lean
jstoobysmith, line 1: Would rename to NewtonianSystem/Basic.lean

Physlib/ClassicalMechanics/PointParticle/NewtonianSystem/Defs.lean
```diff
@@ -0,0 +1,207 @@
...
+open ReferenceFrame
+
+variable {d : ℕ}
+
+/-!
+## A. Systems
+-/
+
+/-- A finite system of point particles satisfying Newton's laws. -/
+structure System (d : ℕ) where
```
jstoobysmith: `NewtonianSystem` to match directory name?

Physlib/ClassicalMechanics/PointParticle/NewtonianSystem/Defs.lean
```diff
@@ -0,0 +1,207 @@
...
+
+/-- Vectors in the system's frame. -/
+abbrev Vector := system.frame.Vector
+
+/-- A particle in `system`. -/
+abbrev Particle : Type := system.particles
+
+namespace Particle
+
+variable {system : System d} (particle : system.Particle) (t : Time)
```
jstoobysmith: I would not define `particle : system.Particle` or any of the results which follow this. Instead you have all of these defined for a general `frame.Particle`, and they should all follow from that. Similar with force below.

Physlib/ClassicalMechanics/PointParticle/Basic.lean
jstoobysmith: But you have included the trajectory in it, which is a specification of the kinematics of a particle. Would maybe rename this to `ParticleKinematics` in this case.

Physlib/ClassicalMechanics/Force.lean
```diff
@@ -0,0 +1,80 @@
...
+open scoped BigOperators Classical
+
+namespace ClassicalMechanics.ReferenceFrame
+
+variable {d : ℕ} {frame : ReferenceFrame d} {Object : Type}
+
+/-- A time-dependent force acting on an object. -/
+structure Force (frame : ReferenceFrame d) (Object : Type) where
+  /-- The force vector. -/
+  value : Time → frame.Vector
```
jstoobysmith, line 47: Does `ReferenceFrame` contain a notion of time dimension (a basis on `Time`)? If not, I think it needs to if you want to define a force as an element of `frame.Vector`. More generally it would be nice to get all of this to work with the dimensions and units in PHyslib.

jstoobysmith (commented):
Some more comments

Physlib/ClassicalMechanics/PointParticle/NewtonianSystem/Defs.lean
RaunakChhatwal: The system level data types are an alias for the system's multisets, not the top level data types. The top level particles and forces don't identify corresponding elements of multisets, because different multiset elements may have the same underlying value. My intent is for downstream to work mostly with the multiset elements directly through the convenient aliases. Also, elements of multisets don't get the methods of the underlying type, they're accessible only through an inelegant `.1` (like [here](https://github.com/leanprover-community/physlib/pull/1612/changes#diff-7120c667c22de8ab6731c3f85133952360a3ca62e80f764bfd9b8a4a8dd3e37fR67)).

Physlib/ClassicalMechanics/PointParticle/Basic.lean
RaunakChhatwal: A particle's motion is just one property, there's also mass (and maybe charge in the future). Also, lean in general allows one to hypothetically consider different values of a variable. Like different hypothetical trajectories for the same particle variable

Physlib/ClassicalMechanics/Force.lean
RaunakChhatwal: [Time/Basic.lean](https://github.com/leanprover-community/physlib/blob/67d742760382efce9419feefa348902aab797a07/Physlib/SpaceAndTime/Time/Basic.lean#L16)
> In this module we define the type `Time`, corresponding to time in a given (but arbitrary) set of units, with a given (but arbitrary) choice of origin (time zero)

`ReferenceFrame` contains neither origin nor basis in time because time coordinates are relative to that global arbitrary unit and origin. Even if we removed the global unit and origin, reference frames should only contain time origin because galilean frame transformations do not change the basis in time. Time coordinates then would need to be dimensionful.

Regarding the adoption of dimensions and units, my first step would be to refactor space as an arbitrary coordinate-free affine space over the displacement vector quantity and time as an arbitrary affine space over the time scalar quantity (used as 1d vector space).

An example from a prototype:
```
lemma affine_space_exists (V : Type) [AddGroup V] :
    ∃ P : Type, Nonempty (AffineSpace V P) :=
  ⟨V, ⟨inferInstance⟩⟩

def Space (dim : ℕ) : Type :=
  Classical.choose (affine_space_exists (Displacement dim))

instance {dim : ℕ} : AffineSpace (Displacement dim) (Space dim) :=
  Classical.choice (affine_space_exists (Displacement dim)).choose_spec

def Time : Type :=
  Classical.choose (affine_space_exists Duration)

instance : AffineSpace Duration Time :=
  Classical.choice (affine_space_exists Duration).choose_spec
```

Regardless, I think fundamental issues with the present design of dimensions/units should be discussed on zulip before undertaking adoption.

jstoobysmith: Ok, but then I think we should make it clear that here we are using that global origin and dimension on `Time`, meaning that `value t` is the value of the force in these units.

Physlib/ClassicalMechanics/PointParticle/Defs.lean
```diff
@@ -0,0 +1,92 @@
...
+
+namespace ClassicalMechanics.ReferenceFrame
+
+variable {d : ℕ} {frame : ReferenceFrame d}
+
+/-- Positive real numbers. -/
+notation "ℝ+" => {x : ℝ // 0 < x}
+
+/-- Scalar multiplication by a positive real. -/
+instance : SMul ℝ+ frame.Vector where
```
jstoobysmith: Move to near definition of frame.Vector

Physlib/ClassicalMechanics/PointParticle/Basic.lean
jstoobysmith, line 1: would rename to Basic.lean

Physlib/ClassicalMechanics/PointParticle/NewtonianSystem/Basic.lean
```diff
@@ -0,0 +1,207 @@
...
+/-!
+## A. Systems
+-/
+
+/-- A finite system of point particles satisfying Newton's laws. -/
+structure NewtonianSystem (d : ℕ) where
+  /-- The system's reference frame. -/
+  frame : ReferenceFrame d
+  [isInertial : Fact frame.IsInertial]
+
```
jstoobysmith: Delete new lines in definitions

Physlib/ClassicalMechanics/Force.lean
```diff
@@ -0,0 +1,80 @@
...
+  target := force.source
+  source := force.target
+  source_ne_target := force.source_ne_target.symm
+
+/-- The net force on `object`. -/
+def netForce
+    (object : Object)
+    (internalForces : Multiset (frame.InternalForce Object))
+    (externalForces : Multiset (frame.Force Object))
+    (t : Time) : frame.Vector :=
```
jstoobysmith: Would collapse the lines here, so that the arguments appear on one line

Physlib/ClassicalMechanics/Force.lean
RaunakChhatwal: Force is just one vector function of time, among many others (like pos, velocity, momentum, etc). I clarified in ReferenceFrame.lean

jstoobysmith (approved):
Approved. Many thanks for the iterations. Will merge shortly.

PR #1614, opened 2026-09-03
refactor(Electromagnetism): replace `fieldStrengthMatrix` with index-evaluation components

zhikaip:
First of two PRs for the TODOs in `Electromagnetism/Kinematics/FieldStrength.lean`; the
follow-up `fieldstrength_refactor` removes the basis-representation lemmas.

## Changes

- Components of the field strength are accessed as `toField {A.toFieldStrength x | [μ] [ν]}ᵀ`
  everywhere. `fieldStrengthMatrix` and its API are removed, and every lemma stated with it is
  restated in eval form, with `fieldStrengthMatrix` ↦ `toFieldStrength_eval` in the name.
- Two new lemmas, `toFieldStrength_eval_eq_tensor_basis_repr` and
  `toFieldStrength_tensor_basis_repr_eq_eval`, bridge eval components and the coefficients in
  `Tensor.basis`, the tensor library's canonical basis. They replace the role of the
  `fieldStrengthMatrix` bridge lemmas: `KineticTerm` expands `F_{μν}F^{μν}` as a sum over
  `Tensor.basis` coefficients and needs to turn each coefficient into a component, and
  antisymmetry, additivity and homogeneity of `toFieldStrength` are derived from their
  component versions through the same bridge.
- `Tensors/Elab.lean`: the bracket index syntax accepts a `term` instead of an `ident`, so
  concrete components such as `[Sum.inl 0] [Sum.inr i]` can be written. Existing uses are
  unaffected.
- Four files gain `attribute [-simp] Fin.succAbove_zero` (MagneticField, Boosts, KineticTerm,
  Hamiltonian); each fails without it.
- `canonicalMomentum_eq` is reproved via `Lorentz.Vector.inner_eq_sum`, and the two options
  in `Vector/MinkowskiProduct.lean` are dropped. With #1610 and #1611 no
  `respectTransparency` option remains in the electromagnetism files.

## Why

- Index evaluation is the tensor library's own notation for components; `fieldStrengthMatrix`
  was a second representation of the same thing with a duplicate API.
- `Fin.succAbove_zero`: plain `simp` rewrites `Fin.succAbove 0` to `Fin.succ` inside the
  implicit colour index of `evalT`, after which no eval lemma matches. The tensor library
  disables it file-locally for the same reason; the attribute does not propagate.

replacements are mostly mechanical and AI seem to do a good job. I checked it to the best of my ability

### DIFF
--- a/Physlib/ClassicalFieldTheory/GaugeTheory/API-map.yaml
+++ b/Physlib/ClassicalFieldTheory/GaugeTheory/API-map.yaml
@@ -48,11 +48,11 @@ Requirements:
 
   - description: "Gauge invariance of the field strength under abelian U(1) gauge transformations"
     done: true
-    location: Physlib/Electromagnetism/Kinematics/GaugeTransformation.lean (toFieldStrength_gaugeTransform, fieldStrengthMatrix_gaugeTransform)
+    location: Physlib/Electromagnetism/Kinematics/GaugeTransformation.lean (toFieldStrength_gaugeTransform, toFieldStrength_eval_gaugeTransform)
 
   - description: "Pure-gauge (flat) configurations have vanishing curvature, and the bare gradient does not (necessity of the metric contraction)"
     done: true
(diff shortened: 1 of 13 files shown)


jstoobysmith (approved):
This all looks good to me - have approved

PR #1616, opened 2026-09-06
feat(EMPotential): differentiability of derivatives of the electromagnetic potential

zhikaip:
Resolves the `EMPotential.lean` TODO on differentiability of the derivative of the potential.

The guiding rule is that a lemma is stated only where `fun_prop` cannot get there on its own. Three such gaps exist: `fun_prop` does not lower `ContDiff ℝ ∞` to a finite order (the `_of_smooth` lemmas), it cannot unify a symbolic `n + 2` with `?m + 1` (the `C^{n+2}` second-derivative lemmas, in both the `_component` and the `_apply` shape), and it will not `unfold deriv` (the tensor lemmas `differentiable_deriv`, `contDiff_deriv`). Everything a bare `fun_prop` already proves, including three pre-existing component lemmas, is removed under the same rule.

Naming follows the object: plain names for the tensor `A.deriv`, `_component` for "take the component, then differentiate", `_apply` for "differentiate, then take the component".

Riding along: two `FieldStrength.lean` proofs now use `fun_prop` instead of re-deriving component differentiability, and the stale `respectTransparency` options in `EMPotential.lean` are dropped.

prepared by claude, sanity checked and suggested cleanup by myself

### DIFF
--- a/Physlib/Electromagnetism/Kinematics/EMPotential.lean
+++ b/Physlib/Electromagnetism/Kinematics/EMPotential.lean
@@ -286,17 +286,12 @@ noncomputable instance {d} :
 
 ### A.4. Differentiability
 
-We show that the components of the potential, and of its derivatives, are differentiable
-(or smooth) if the potential is.
+We show that the potential and its derivatives are differentiable (or smooth) if the potential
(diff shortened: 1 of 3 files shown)


jstoobysmith (approved):
Approved - looks good to me.

PR #1618, opened 2026-09-07
feat(Time): finite-order regularity of the time derivative and a coordinate chain rule

zhikaip:
adding some API in Time.Derivatives

- The regularity lemmas for ∂ₜ in Time.Derivatives are stated for smooth curves only. This adds the C^n versions so that C² dynamics can use them.
- Adds a chain rule for a real function of one coordinate of a curve, the time-derivative counterpart of gradient_comp_coord.
- `deriv_deriv_comp_neg` now asks for C² instead of smooth, with its one caller in SimplePendulum adjusted.

Prepared by cluade, trimmed down and tidied up by myself


jstoobysmith (approved):
Looks good to me. Approved.

PR #1619, opened 2026-09-08
feat(PhyslibAlpha): the Norton dome

zhikaip:
- Norton's dome is a much-debated example in the philosophy of physics: a mass at rest on its apex may stay there or slide off at any instant, in obedience to Newton's laws, because the force is continuous but not Lipschitz at the apex. This formalises the dome and the non-uniqueness of its motions.
- It illustrates the arguments in the debate by making determinism and the candidate regularity conditions predicates on a minimal Newtonian system, so that what each condition buys is a theorem; which condition to adopt is left open, since that choice is not something a formalisation can settle.
- Local existence rests on a sorried statement of Peano's theorem copied from Mathlib PR #42148 under its own names, so the file should be deleted when that lands.

^written by Claude

this is a side project that I've been directing Claude to do, this probably doesn't belong in Physlib both in terms of quality and content, but I think this is a good example of what formalisation enables us to clarify, and a good reminder of the pitfalls of trying to come up with a universal framework that serves all functions i.e. sometimes there may not be one single good choice and it's our role to map out the boundary of theories instead.


jstoobysmith (approved):
Approved.

PR #1620, opened 2026-09-09
refactor(Time): make Time an affine space, add time origin to reference frames

RaunakChhatwal:
## Summary

Follow-up to the time-origin discussion in #1612. There, `ReferenceFrame` carried no time data and its docstring declared `Time` equivalent to time coordinates in every frame. This PR removes that convention: `Time` becomes an affine space over `ℝ`, each `ReferenceFrame` carries its own `timeOrigin`, and frame-relative quantities become functions of the frame's real-valued time coordinate.

## Motivation

* **Time is affine, like space.** `ReferenceFrame.lean` already argues that a point of `Space d` is not a coordinate: two points determine a displacement, but the zero point belongs to the frame. The same holds for instants of time. Two instants determine a real elapsed duration, but no instant is canonically zero. `AddTorsor ℝ Time` encodes this, and `Time.val` is demoted to an implementation coordinate.
* **The time origin belongs to the frame; the time basis does not.** A Galilean transformation may translate time but never rescales or reverses it, so a frame records `timeOrigin : Time` and nothing else about time. Unit and orientation stay global, exactly as spatial units are shared across frames.
* **Coordinates are real numbers.** Positions, velocities, and forces are already frame coordinate tuples (`frame.Vector`), not points of `Space d`. Their time argument should likewise be the frame coordinate `ℝ`, not the absolute `Time`. `ReferenceFrame.timeEquiv` converts back when an absolute instant is needed.
* **Existing uses of `Time` are preserved.** The full scalar structure moves to `Time.Scalar`, with imports updated where needed. `Time.Basic` contains the affine structure, so `ReferenceFrame` only depends on it.

### DIFF
--- a/Physlib/SpaceAndTime/Time/Basic.lean
+++ b/Physlib/SpaceAndTime/Time/Basic.lean
@@ -16,7 +16,7 @@ such a duration to an instant produces another instant. There is no distinguishe
 
 The field `Time.val` is an implementation coordinate, not a frame-relative time
 coordinate. A reference frame chooses its own time origin. Import
-`Physlib.SpaceAndTime.Time.Scalar` to use the additional scalar structure with the
+`Physlib.SpaceAndTime.Time.InnerProductSpace` to use the inner product space structure with the
 implicit origin `Time.mk 0`, including addition of instants, norms, and derivatives.
 -/
 
--- a/Physlib/SpaceAndTime/Time/Basic.lean
+++ b/Physlib/SpaceAndTime/Time/Basic.lean
@@ -5,457 +5,59 @@ Authors: Joseph Tooby-Smith
 -/
 module
 
-public import Mathlib.Analysis.Calculus.FDeriv.Linear
-public import Mathlib.MeasureTheory.Measure.Haar.InnerProductSpace
+public import Mathlib.Data.Real.Basic
+public import Mathlib.LinearAlgebra.AffineSpace.Defs
 /-!
-# Time
+# Affine time
 
-## i. Overview
+`Time` represents instants with a fixed but arbitrary unit and orientation. It is an
+affine space over `ℝ`: two instants determine a real-valued elapsed time, and adding
+such a duration to an instant produces another instant. There is no distinguished zero.
 
-In this module we define the type `Time`, corresponding to time in a given
-(but arbitrary) set of units, with a given (but arbitrary) choice of origin (time zero),
-and a choice of orientation (i.e. a positive time direction).
-
-We note that this is the version of time most often used in undergraduate and
-non-mathematical physics.
-
-The choice of units or origin can be made on a case-by-case basis, as
-long as they are done consistently. However, since the choice of units and origin is
-left implicit, Lean will not catch inconsistencies in the choice of units or origin when
-working with `Time`.
-
-For example, for the classical mechanics system corresponding to the harmonic oscillator,
-one can take the origin of time to be the time at which the initial conditions are specified,
-and the units can be taken as anything as long as the units chosen for time `t` and
-the angular frequency `ω` are consistent.
-
-With this notion of `Time`, becomes a 1d vector space over `ℝ` with an inner product.
-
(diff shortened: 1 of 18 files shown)


Physlib/SpaceAndTime/Time/Affine.lean
```diff
@@ -0,0 +1,65 @@
...
+
+The field `Time.val` is an implementation coordinate, not a frame-relative time
+coordinate. A reference frame chooses its own time origin. Import
+`Physlib.SpaceAndTime.Time.Basic` to use the additional scalar structure with the
+implicit origin `Time.mk 0`, including addition of instants, norms, and derivatives.
+-/
+
+@[expose] public section
+
+open scoped Affine
```
zhikaip: have you checked the implementation of `Space`? i.e. maybe the definitions here should belong in `Time.Basic`, and the relevant instances currently in `Time.Basic` should be moved into something like `Time.Module`.

Physlib/SpaceAndTime/Time/Affine.lean
```diff
@@ -0,0 +1,65 @@
...
+@[simp]
+lemma vadd_val (dt : ℝ) (t : Time) : (dt +ᵥ t).val = dt + t.val := rfl
+
+instance : VSub ℝ Time where
+  vsub t₁ t₂ := t₁.val - t₂.val
+
+@[simp]
+lemma vsub_eq_val (t₁ t₂ : Time) : t₁ -ᵥ t₂ = t₁.val - t₂.val := rfl
+
+instance : AffineSpace ℝ Time where
```
zhikaip: `AffineSpace` is just notation for `AddTorsor`. Again to draw parallel with `Space` (and to avoid `open scoped Affine` in every relevant file), I suggest we use `AddTorsor` here so the language for `Time` and `Space` is uniform.

zhikaip (commented):
Unless there's something I haven't thought of, I'm fine with changing this to match the `Space` structure, but this is a foundational change that we want to do properly. if you don't mind I'd suggest you open a PR specifically on adding the `AddTorsor` instance to `Time`, and restructure the material accordingly there. there are also subtleties that I'd prefer to have an extra eye on, e.g. how this interacts with `TimeMan` and `TimeTransMan`, maybe @jstoobysmith and others have more to add

Physlib/SpaceAndTime/Time/InnerProductSpace.lean
jstoobysmith, line 1: I would maybe rename this to `Module` so it matches `Space.

Physlib/SpaceAndTime/Time/TimeTransMan.lean
jstoobysmith, line 1: I think this change superseeds `TimeTransMan` which we should at least add a TODO item about removing.

Physlib/SpaceAndTime/Time/Basic.lean
```diff
@@ -5,457 +5,59 @@ Authors: Joseph Tooby-Smith
 -/
 module
 
-public import Mathlib.Analysis.Calculus.FDeriv.Linear
-public import Mathlib.MeasureTheory.Measure.Haar.InnerProductSpace
+public import Mathlib.Data.Real.Basic
+public import Mathlib.LinearAlgebra.AffineSpace.Defs
 /-!
-# Time
+# Affine time
```
jstoobysmith: Would still call this `Time`.

Physlib/SpaceAndTime/Time/Scalar.lean
RaunakChhatwal: I think Scalar.lean better captures the intent of that file, namely treating time as a scalar quantity after choosing an implicit origin. It also defines coercions and equivalences involving the naturals and reals

RaunakChhatwal (commented):
-awaiting-author

RaunakChhatwal (commented):
Thanks for the review, I simplified the `Time` header and added removal TODO

jstoobysmith (approved):
Approved. Will merge shortly

PR #1622, opened 2026-09-09
feat(PhyslibAlpha): Replace OperatorAlgebra with AlgebraicFramework

TomOleDiem:
Adds `PhyslibAlpha/AlgebraicFramework/`: an order-unit-space-first treatment
of states, effects, channels, and measurements, plus C*-algebra observables
built on top, and the first slice of the unbounded-operator spectral
theorem (Cayley transform, spectral integrals, both directions of Stone's
theorem).

Removes `QuantumMechanics/OperatorAlgebra`, which this supersedes.


jstoobysmith (approved):
Ok, approved. Will look into the linter

PR #1629, opened 2026-09-10
feat(QuantumMechanics/Blackbody): Added Canonical Definition Planck's law

Sammybro11:
## Description of Changes

New Directory at:
`Physlib/QuantumMechanics/Blackbody` to store further blackbody radiation related definitions and proofs.

New Files at:
`Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean` Added definitions and positive and zero values lemmas for Spectral Radiance due to Planck's Law

Modified File at:
`Physlib.lean` added import for Planck's Law

## Reviewer Guide

I have added frequency with Real number type right now but plan to make changes to that later when frequency is implemented. Also the conversion between frequency and wavelength can be added after that.

### DIFF
--- a/Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
+++ b/Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
@@ -79,3 +79,9 @@ lemma spectralRadiance_pos (c : SpeedOfLight) (ν : ℝ) (T : Temperature)
         div_pos (mul_pos h_pos ν_pos) (mul_pos kB_pos (by exact_mod_cast T_pos))
       exact mul_pos (pow_pos c.val_pos 2)
        (sub_pos.mpr (by simpa using Real.exp_strictMono expo_term))
+
+/-- For a given ν frequency, the spectral radiance of a blackbody is 0 at
+    absolute zero temperature. -/
+lemma spectralRadiance_absZero (c : SpeedOfLight) (ν : ℝ) :
+    spectralRadiance c ν ⟨0⟩ = 0 := by
+    simp [spectralRadiance]
--- a/Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
+++ b/Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
@@ -36,6 +36,7 @@ where `h` is Planck's constant, `c` the speed of light, and `k_B` the Boltzmann 
 - `spectralRadiance` : The spectral radiance per unit frequency of blackbody radiation.
 - `spectralRadiance_pos` : The spectral radiance is positive for positive frequency
   and temperature.
+- `spectralRadiance_absZero` : The spectral radiance is 0 at absolute zero.
 
 ## iii. Table of contents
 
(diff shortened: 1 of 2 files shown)


Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
```diff
@@ -0,0 +1,87 @@
...
+## iv. References
+
+* https://en.wikipedia.org/wiki/Planck%27s_law
+
+-/
+
+@[expose] public section
+
+
+namespace Constants
```
nateabr: `spectralRadiance` is not itself a physical constant, so putting it in Constants makes the API misleading. Could this be moved into a namespace such as `Blackbody` or another appropriate radiation namespace?

Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
```diff
@@ -0,0 +1,87 @@
...
+    refine div_pos ?numerator ?denominator
+    · exact mul_pos (mul_pos (by norm_num) h_pos) (pow_pos ν_pos 3)
+    · have expo_term : 0 < h * ν / (kB * (T : ℝ)) :=
+        div_pos (mul_pos h_pos ν_pos) (mul_pos kB_pos (by exact_mod_cast T_pos))
+      exact mul_pos (pow_pos c.val_pos 2)
+       (sub_pos.mpr (by simpa using Real.exp_strictMono expo_term))
+
+/-- For a given ν frequency, the spectral radiance of a blackbody is 0 at
+    absolute zero temperature. -/
+lemma spectralRadiance_absZero (c : SpeedOfLight) (ν : ℝ) :
```
nateabr, line 92: If we look at the definiton of `spectralRadiance` at T = 0, kB * (T : ℝ) = 0, therefore the computation below is 0/0, which in lean, by convention, is 0. Ofcourse mathematically, this proof is not valid even though the code has compiled. 

We could define the physical formula only on positive frequency and temperature, and extend it by zero otherwise:


``noncomputable def spectralRadiance (c : SpeedOfLight) (ν : ℝ) (T : Temperature) : ℝ :=
  if 0 < ν ∧ 0 < (T : ℝ) then
    2 * h * ν ^ 3 / ((c : ℝ) ^ 2 * (Real.exp (h * ν / (kB * (T : ℝ))) - 1))
  else
    0``


Then the absolute-zero lemma reflects an explicit choice in the API, rather than Lean’s division convention. A later PR could state and prove that this extension agrees with the limit as $T \to 0^+$.

nateabr (commented):
Looks good! Some comments to fix

Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
Sammybro11: That makes alot more sense, I was wondering why I was getting no errors. Thank you

Physlib/QuantumMechanics/Blackbody/PlancksLaw.lean
Sammybro11: Decided to name it Blackbody itself since the folder is also blackbody, other name possibly could have been EMR but that would intersect with EM based modules, so I just went with blackbody

jstoobysmith (approved):

PR #1632, opened 2026-09-10
feat(Cosmology): de Sitter solution of the Friedmann equations

KevorkianPhilippe:
## AI disclosure

**AI disclosure.** This PR was generated with Claude Fable 5.1 (Claude Code) under my
supervision, following `AI-POLICY.md` and `AGENTS.md`. I have read every definition and lemma
statement and vouch that they state what the docstrings say. The statements were fixed before
the proofs were written and checked symbolically (sympy) against the Friedmann equations of
`Physlib.Cosmology.FLRW.Basic`; no textbook is cited, the TODO items of the file are the
specification.

Resolves the first TODO of `Physlib/Cosmology/FLRW/Solutions.lean` (de Sitter).

Declarations added (namespace `Cosmology.FLRW.FriedmannEquation`):
- `Time.deriv_comp_val` (in `SpaceAndTime/Time/Derivatives.lean`): the time derivative of `t ↦ γ t.val` is the Mathlib derivative of `γ`
  (a thin wrapper of `Time.deriv_comp_toRealCLE_of_hasDerivAt`; could move to
  `Physlib.SpaceAndTime.Time.Derivatives` if preferred).
- `deSitterScaleFactor`: `a(t) = a₀ exp(σ √(Λ/3) c t)`, `σ = ±1`.
- `deriv_deSitterScaleFactor`, `deriv_deriv_deSitterScaleFactor`:
  its first and second time derivatives.
- `sq_deSitterRate`: `(σ √(Λ/3) c)² = Λ c² / 3`.
- `deSitterScaleFactor_firstOrderFriedmann`, `deSitterScaleFactor_secondOrderFriedmann`: it
  solves both Friedmann equations with `ρ = 0`, `p = 0`, `k = 0`, `Λ > 0`.
- `hubbleConstant_deSitterScaleFactor`: `H = σ √(Λ/3) c`.
- `decelerationParameter_deSitterScaleFactor`: `q = -1`.

Reviewer map: module docstring, then `deSitterScaleFactor` and the two Friedmann lemmas (B.2),
then the derivative lemmas (B.1) and B.3.

Checks: `lake build Physlib`, `lake exe runPhyslibLinters`, `./scripts/lint-style.sh`, codespell;
`#print axioms` standard on every declaration.

[...]

PR #1633, opened 2026-09-10
feat(Cosmology): flat power-law solutions of the Friedmann equations (radiation, Einstein-de Sitter)

KevorkianPhilippe:
## AI disclosure

specification.

**Stacked on #1632** (`flrw-solutions-1`): the diff includes the earlier commit(s); only the last commit is new here.

Resolves the radiation-dominated and Einstein-de Sitter TODOs.

Declarations added:
- `Time.deriv_eq_of_hasDerivAt` (in `SpaceAndTime/Time/Derivatives.lean`): the time derivative of any `f : Time → ℝ` from a `HasDerivAt` of
  `τ ↦ f ⟨τ⟩`.
- `powerLawScaleFactor`: `a(t) = (t / t₀) ^ n` (real power); 
  `deriv_powerLawScaleFactor` (for `t ≠ 0`), `deriv_deriv_powerLawScaleFactor` (for `t > 0`,
  through `HasDerivAt.congr_of_eventuallyEq` since `∂ₜ a` is only known away from `t = 0`).
- `hubbleConstant_powerLawScaleFactor`: `H = n / t`; `decelerationParameter_powerLawScaleFactor`:
  `q = (1 - n) / n`.
- `powerLawDensity`: `ρ = 3 n² / (8 π G t²)`, the density imposed by the first-order equation;
  `powerLawScaleFactor_firstOrderFriedmann` (`k = 0`, `Λ = 0`);
  `powerLawScaleFactor_secondOrderFriedmann` with `p = w ρ c²`, `1 + 3 w = 2 (1 - n) / n`.
- `radiationScaleFactor` (`n = 1/2`), `radiationDensity` (`3 / (32 π G t²)`),
  `radiationPressure` (`ρ c² / 3`), `radiationDensity_eq`,
  `radiationScaleFactor_firstOrderFriedmann`, `radiationScaleFactor_secondOrderFriedmann`,
  `decelerationParameter_radiationScaleFactor` (`q = 1`),
  `hubbleConstant_radiationScaleFactor_t₀` (`H(t₀) = 1 / (2 t₀)`).
- `einsteinDeSitterScaleFactor` (`n = 2/3`), `einsteinDeSitterDensity` (`1 / (6 π G t²)`),
  `einsteinDeSitterDensity_eq`, `einsteinDeSitterScaleFactor_firstOrderFriedmann`,
  `einsteinDeSitterScaleFactor_secondOrderFriedmann` (`p = 0`),
  `decelerationParameter_einsteinDeSitterScaleFactor` (`q = 1 / 2`),
  `hubbleConstant_einsteinDeSitterScaleFactor_t₀` (`H(t₀) = 2 / (3 t₀)`).

Note: the TODOs do not give the densities; they are imposed by the first-order equation with
`k = 0`, `Λ = 0` and stated explicitly.

Reviewer map: C.1 (power law and derivatives), C.2 (`H`, `q`, the two general Friedmann lemmas),
then C.3 and C.4 (instances).

**Update (Sept 11, after the merge of #1632).** `master` was merged into this branch (no force-push). Since #1632 was merged before its review fixes were pushed, this PR now also carries those fixes for the de Sitter part of `Solutions.lean` (the bridge `deriv_comp_val` moved to `Time/Derivatives.lean`, the one-off exponential helper inlined, sections renumbered).

### DIFF
--- a/Physlib/Cosmology/FLRW/Solutions.lean
+++ b/Physlib/Cosmology/FLRW/Solutions.lean
@@ -1,27 +1,186 @@
 /-
 Copyright (c) 2026 Jinzheng Li. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Jinzheng Li
+Authors: Jinzheng Li, Philippe Kevorkian
 -/
 module
 
 public import Physlib.Meta.TODO.Basic
+public import Physlib.Cosmology.FLRW.Basic
 /-!
 
 # Exact solutions of the Friedmann equations
 
-Placeholder file collecting TODO items for the standard closed-form solutions of
-the Friedmann equations: the de Sitter, radiation-dominated, Einstein-de Sitter
-(dust) and Milne models, together with the Einstein static universe. This file
-contains only TODO items; no definitions or lemmas are formalized yet.
+## i. Overview
+
+This file collects the standard closed-form solutions of the Friedmann equations
+(`FirstOrderFriedmann` and `SecondOrderFriedmann` of `Physlib.Cosmology.FLRW.Basic`) and
+proves that they solve them: the de Sitter solution here, the radiation-dominated,
+Einstein-de Sitter and Milne models and the Einstein static universe being still TODO items.
+
+Each solution is a scale factor `a : Time → ℝ` given by an explicit function of the time
+coordinate `t.val`. Its time derivative `∂ₜ a` is computed through the bridge
+`deriv_comp_toRealCLE_of_hasDerivAt` from Mathlib's `HasDerivAt` on `ℝ`.
+
+## ii. Key results
+
+- `deSitterScaleFactor`: `a(t) = a₀ exp(σ √(Λ/3) c t)` with `σ = ±1`.
+- `deSitterScaleFactor_firstOrderFriedmann`, `deSitterScaleFactor_secondOrderFriedmann`:
+  it solves both Friedmann equations with `ρ = 0`, `p = 0`, `k = 0` and `Λ > 0`.
+- `hubbleConstant_deSitterScaleFactor`: its Hubble parameter is the constant `σ √(Λ/3) c`
+  (for any `σ`, `Λ`, `c`).
+- `decelerationParameter_deSitterScaleFactor`: its deceleration parameter is `q = -1`.
+
+## iii. Table of contents
+
+- A. Time derivatives of curves given by a function of the coordinate
+- B. The de Sitter solution
+  - B.1. The scale factor and its derivatives
+  - B.2. The Friedmann equations
+  - B.3. The Hubble and deceleration parameters
+- C. Remaining TODO items
 
 -/
 
 @[expose] public section
 
-TODO "Prove that the de Sitter solution `a(t) = a₀ exp(±√(Λ/3) c t)` (with `ρ = 0`,
-  `K = 0`, `Λ > 0`) solves the Friedmann equations, and that its deceleration
-  parameter is `q = −1`."
+namespace Cosmology.FLRW.FriedmannEquation
+
+open Real Time
+
+/-!
+
+## A. Time derivatives of curves given by a function of the coordinate
+
+A curve `t ↦ γ t.val` on `Time` is the pull-back through `toRealCLE` of the curve `γ` on `ℝ`;
+its time derivative is the Mathlib derivative of `γ`.
+
+-/
+
+/-- The time derivative of `t ↦ γ t.val` at `t` is the derivative of `γ` at `t.val`. -/
+lemma deriv_comp_val {γ : ℝ → ℝ} {t : Time} {v : ℝ} (h : HasDerivAt γ v t.val) :
+    ∂ₜ (fun s : Time => γ s.val) t = v :=
+  deriv_comp_toRealCLE_of_hasDerivAt γ t v h
+
+/-!
+
+## B. The de Sitter solution
+
+-/
+
+/-!
+
+### B.1. The scale factor and its derivatives
+
+-/
+
+/-- The de Sitter scale factor `a(t) = a₀ exp(σ √(Λ/3) c t)`, for `σ = ±1`
+  (the expanding branch is `σ = 1`). -/
+noncomputable def deSitterScaleFactor (a₀ σ Λ c : ℝ) : Time → ℝ :=
+  fun t => a₀ * Real.exp (σ * √(Λ / 3) * c * t.val)
+
+/-- Mathlib derivative of `y ↦ a₀ exp (K y)`. -/
+lemma hasDerivAt_mul_exp_mul (a₀ K x : ℝ) :
+    HasDerivAt (fun y : ℝ => a₀ * Real.exp (K * y)) (a₀ * K * Real.exp (K * x)) x := by
+  have h := (((hasDerivAt_id x).const_mul K).exp).const_mul a₀
+  refine h.congr_deriv ?_
+  simp only [id_eq]
+  ring
+
+lemma deriv_deSitterScaleFactor (a₀ σ Λ c : ℝ) :
+    ∂ₜ (deSitterScaleFactor a₀ σ Λ c) =
+      fun t => a₀ * (σ * √(Λ / 3) * c) * Real.exp (σ * √(Λ / 3) * c * t.val) := by
+  funext t
+  exact deriv_comp_val (hasDerivAt_mul_exp_mul a₀ (σ * √(Λ / 3) * c) t.val)
+
+lemma deriv_deriv_deSitterScaleFactor (a₀ σ Λ c : ℝ) :
+    ∂ₜ (∂ₜ (deSitterScaleFactor a₀ σ Λ c)) =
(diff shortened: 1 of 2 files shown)


Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -79,28 +82,33 @@ lemma deriv_comp_val {γ : ℝ → ℝ} {t : Time} {v : ℝ} (h : HasDerivAt γ
 noncomputable def deSitterScaleFactor (a₀ σ Λ c : ℝ) : Time → ℝ :=
   fun t => a₀ * Real.exp (σ * √(Λ / 3) * c * t.val)
 
-/-- Mathlib derivative of `y ↦ a₀ exp (K y)`. -/
-lemma hasDerivAt_mul_exp_mul (a₀ K x : ℝ) :
```
doxtor6, line 83: not sure why you delete this lemma. Does it make the code cleaner?

doxtor6 (commented):
Moving here the comments I left on #1634 that actually concern this PR's diff (#1634 is stacked on this one, so its diff showed these defs).

General principle: `def`s are the load-bearing part of the library — downstream files refer to them by name — so we only add them for objects that deserve a name, and those need to be high quality. This PR adds 7 `def`s to the file; I think 2 should stay (the two named scale factors) and the rest should be inlined into the theorem statements. The two `_eq` glue lemmas (`radiationDensity_eq`, `einsteinDeSitterDensity_eq`) are the symptom: they exist only to undo defs.

For the defs that stay, two points on quality: `t.val = 0` hard-codes the singularity at the origin of the chart, whereas `Time` is an `AddTorsor ℝ Time` with no distinguished origin; and `t₀` is a time, so it should be `t₀ : Time` rather than `ℝ` (the `⟨t₀⟩` in `hubbleConstant_radiationScaleFactor_t₀` is the tell).

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+-/
+
+/-!
+
+### B.1. The power-law scale factor and its derivatives
+
+-/
+
+/-- The power-law scale factor `a(t) = (t / t₀) ^ n` (real power). -/
+noncomputable def powerLawScaleFactor (t₀ n : ℝ) : Time → ℝ :=
```
doxtor6: Is this def necessary? It is proof-sharing scaffolding rather than a named solution. The general-`n` lemmas below work just as well quantified over `n` with `fun t => (t.val / t₀) ^ n` inline in the statement, and then `radiationScaleFactor` / `einsteinDeSitterScaleFactor` can be defined directly instead of as aliases that have to be unfolded.

If a power-law def survives anyway: `t.val` puts the Big Bang at the origin of the chart (`Time` is an `AddTorsor ℝ Time`, no distinguished origin), `t₀` should be a `Time`, and the docstring should say that `t₀` is the normalisation epoch (`a(t₀) = 1`) and that values for `t ≤ 0` are junk (`Real.rpow` on a negative base).

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  rw [deriv_deriv_powerLawScaleFactor ht₀.ne' n ht, deriv_powerLawScaleFactor ht₀.ne' n ht.ne',
+    powerLawScaleFactor, Real.rpow_sub_one (div_pos ht ht₀).ne',
+    Real.rpow_sub_one (div_pos ht ht₀).ne']
+  have hx : (t.val / t₀) ^ n ≠ 0 := (Real.rpow_pos_of_pos (div_pos ht ht₀) n).ne'
+  field_simp
+  ring
+
+/-- The density imposed on the flat power-law solution by the first-order Friedmann equation,
+  `ρ = 3 n² / (8 π G t²)`. -/
+noncomputable def powerLawDensity (G n : ℝ) : Time → ℝ :=
```
doxtor6: Same question. This only ever appears as an argument to its own lemmas. The density is forced by the scale factor, so the reader is better served by seeing `fun t => 3 * n ^ 2 / (8 * π * G * t.val ^ 2)` in the theorem statement than by chasing a name.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  ring
+
+/-!
+
+### B.3. The radiation-dominated solution
+
+-/
+
+/-- The radiation-dominated scale factor `a(t) = (t / t₀) ^ (1/2)`. -/
+noncomputable def radiationScaleFactor (t₀ : ℝ) : Time → ℝ :=
```
doxtor6: Keep this one (it is the named solution the TODO asks for), but define it directly as `fun t => (t.val / t₀) ^ (1 / 2 : ℝ)` rather than as an alias of `powerLawScaleFactor`. Same remarks on `t₀ : Time` and the origin of `t.val`.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  powerLawScaleFactor t₀ (1 / 2)
+
+/-- The density of the flat radiation-dominated solution, `ρ = 3 / (32 π G t²)`, imposed by
+  the first-order Friedmann equation. -/
+noncomputable def radiationDensity (G : ℝ) : Time → ℝ :=
```
doxtor6: Not an independent object: the density is forced by the scale factor. Please put the formula in the statement of `radiationScaleFactor_firstOrderFriedmann` and drop this def; `radiationDensity_eq` then goes with it.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  fun t => 3 / (32 * π * G * t.val ^ 2)
+
+/-- The radiation pressure `p = ρ c² / 3`. -/
+noncomputable def radiationPressure (G c : ℝ) : Time → ℝ :=
```
doxtor6: Same: inline `fun t => ρ t * c ^ 2 / 3` in the second-order statement.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  ring
+
+/-!
+
+### B.4. The Einstein-de Sitter solution
+
+-/
+
+/-- The Einstein-de Sitter (flat, dust) scale factor `a(t) = (t / t₀) ^ (2/3)`. -/
+noncomputable def einsteinDeSitterScaleFactor (t₀ : ℝ) : Time → ℝ :=
```
doxtor6: As for `radiationScaleFactor`: keep, but define directly as `fun t => (t.val / t₀) ^ (2 / 3 : ℝ)`, with `t₀ : Time`.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,16 +186,238 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  powerLawScaleFactor t₀ (2 / 3)
+
+/-- The dust density of the Einstein-de Sitter solution, `ρ = 1 / (6 π G t²)`, imposed by the
+  first-order Friedmann equation. -/
+noncomputable def einsteinDeSitterDensity (G : ℝ) : Time → ℝ :=
```
doxtor6: As for `radiationDensity`: inline in the theorem statement and drop, together with `einsteinDeSitterDensity_eq`.

Physlib/Cosmology/FLRW/Solutions.lean
KevorkianPhilippe: The helper `deriv_comp_val` was not deleted but moved to `Physlib/SpaceAndTime/Time/Derivatives.lean` (next to the definition of `∂ₜ`, where it is reusable; it is still used here), and the one-off lemma `hasDerivAt_mul_exp_mul` was inlined into the only two proofs that used it, following jstoobysmith's remark on #1642 that isolated single-use helpers should not be added to the library. If you would rather keep it as a named lemma I can restore it.

doxtor6: I see, keep it like this.

doxtor6 (approved):
This PR is good to merge now.

PR #1634, opened 2026-09-10
feat(Cosmology): Milne solution and Einstein static universe relations

KevorkianPhilippe:
## AI disclosure

specification.

**Stacked on #1633** (`flrw-solutions-2`): the diff includes the earlier commit(s); only the last commit is new here.

Partially resolves the last two TODOs; what remains open is said in the narrowed TODOs.

Declarations added:
- `milneScaleFactor`: `a(t) = c t`; `deriv_milneScaleFactor`, `deriv_deriv_milneScaleFactor`;
  `milneScaleFactor_firstOrderFriedmann` (`ρ = 0`, `k = -1`, `Λ = 0`, `t > 0`),
  `milneScaleFactor_secondOrderFriedmann`, `decelerationParameter_milneScaleFactor` (`q = 0`).
  NOT included: the vanishing of the curvature (Minkowski in expanding coordinates): the FLRW
  metric is not an object of Physlib yet. The TODO is kept, narrowed.
- `cosmologicalConstantDensity`: `ρ_Λ = Λ c² / (8 π G)`.
- `einsteinStatic_density`: if `∂ₜ ∂ₜ a = 0` and the second-order equation holds with `p = 0`,
  then `ρ = 2 ρ_Λ`; `einsteinStatic_curvature`: with also `∂ₜ a = 0` and the first-order
  equation, `k c² / a² = 4 π G ρ`; `einsteinStatic_curvature_pos`: `k > 0` when `ρ > 0`.
  NOT included: the instability of the equilibrium (no perturbation theory of the Friedmann
  equations yet). The TODO is kept, narrowed.

Reviewer map: D (Milne, short), then E (`cosmologicalConstantDensity` and the three lemmas).

### DIFF
--- a/Physlib/Cosmology/FLRW/Solutions.lean
+++ b/Physlib/Cosmology/FLRW/Solutions.lean
@@ -1,27 +1,186 @@
 /-
 Copyright (c) 2026 Jinzheng Li. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Jinzheng Li
+Authors: Jinzheng Li, Philippe Kevorkian
 -/
 module
 
 public import Physlib.Meta.TODO.Basic
+public import Physlib.Cosmology.FLRW.Basic
 /-!
 
 # Exact solutions of the Friedmann equations
 
-Placeholder file collecting TODO items for the standard closed-form solutions of
-the Friedmann equations: the de Sitter, radiation-dominated, Einstein-de Sitter
-(dust) and Milne models, together with the Einstein static universe. This file
-contains only TODO items; no definitions or lemmas are formalized yet.
+## i. Overview
+
+This file collects the standard closed-form solutions of the Friedmann equations
+(`FirstOrderFriedmann` and `SecondOrderFriedmann` of `Physlib.Cosmology.FLRW.Basic`) and
+proves that they solve them: the de Sitter solution here, the radiation-dominated,
+Einstein-de Sitter and Milne models and the Einstein static universe being still TODO items.
+
+Each solution is a scale factor `a : Time → ℝ` given by an explicit function of the time
+coordinate `t.val`. Its time derivative `∂ₜ a` is computed through the bridge
+`deriv_comp_toRealCLE_of_hasDerivAt` from Mathlib's `HasDerivAt` on `ℝ`.
+
+## ii. Key results
+
+- `deSitterScaleFactor`: `a(t) = a₀ exp(σ √(Λ/3) c t)` with `σ = ±1`.
+- `deSitterScaleFactor_firstOrderFriedmann`, `deSitterScaleFactor_secondOrderFriedmann`:
+  it solves both Friedmann equations with `ρ = 0`, `p = 0`, `k = 0` and `Λ > 0`.
+- `hubbleConstant_deSitterScaleFactor`: its Hubble parameter is the constant `σ √(Λ/3) c`
+  (for any `σ`, `Λ`, `c`).
+- `decelerationParameter_deSitterScaleFactor`: its deceleration parameter is `q = -1`.
+
+## iii. Table of contents
+
+- A. Time derivatives of curves given by a function of the coordinate
+- B. The de Sitter solution
+  - B.1. The scale factor and its derivatives
+  - B.2. The Friedmann equations
+  - B.3. The Hubble and deceleration parameters
+- C. Remaining TODO items
 
 -/
 
 @[expose] public section
 
-TODO "Prove that the de Sitter solution `a(t) = a₀ exp(±√(Λ/3) c t)` (with `ρ = 0`,
-  `K = 0`, `Λ > 0`) solves the Friedmann equations, and that its deceleration
-  parameter is `q = −1`."
+namespace Cosmology.FLRW.FriedmannEquation
+
+open Real Time
+
+/-!
+
+## A. Time derivatives of curves given by a function of the coordinate
+
+A curve `t ↦ γ t.val` on `Time` is the pull-back through `toRealCLE` of the curve `γ` on `ℝ`;
+its time derivative is the Mathlib derivative of `γ`.
+
+-/
+
+/-- The time derivative of `t ↦ γ t.val` at `t` is the derivative of `γ` at `t.val`. -/
+lemma deriv_comp_val {γ : ℝ → ℝ} {t : Time} {v : ℝ} (h : HasDerivAt γ v t.val) :
+    ∂ₜ (fun s : Time => γ s.val) t = v :=
+  deriv_comp_toRealCLE_of_hasDerivAt γ t v h
+
+/-!
+
+## B. The de Sitter solution
+
+-/
+
+/-!
+
+### B.1. The scale factor and its derivatives
+
+-/
+
+/-- The de Sitter scale factor `a(t) = a₀ exp(σ √(Λ/3) c t)`, for `σ = ±1`
+  (the expanding branch is `σ = 1`). -/
+noncomputable def deSitterScaleFactor (a₀ σ Λ c : ℝ) : Time → ℝ :=
+  fun t => a₀ * Real.exp (σ * √(Λ / 3) * c * t.val)
+
+/-- Mathlib derivative of `y ↦ a₀ exp (K y)`. -/
+lemma hasDerivAt_mul_exp_mul (a₀ K x : ℝ) :
+    HasDerivAt (fun y : ℝ => a₀ * Real.exp (K * y)) (a₀ * K * Real.exp (K * x)) x := by
+  have h := (((hasDerivAt_id x).const_mul K).exp).const_mul a₀
+  refine h.congr_deriv ?_
+  simp only [id_eq]
+  ring
+
+lemma deriv_deSitterScaleFactor (a₀ σ Λ c : ℝ) :
+    ∂ₜ (deSitterScaleFactor a₀ σ Λ c) =
+      fun t => a₀ * (σ * √(Λ / 3) * c) * Real.exp (σ * √(Λ / 3) * c * t.val) := by
+  funext t
+  exact deriv_comp_val (hasDerivAt_mul_exp_mul a₀ (σ * √(Λ / 3) * c) t.val)
+
+lemma deriv_deriv_deSitterScaleFactor (a₀ σ Λ c : ℝ) :
+    ∂ₜ (∂ₜ (deSitterScaleFactor a₀ σ Λ c)) =
+      fun t => a₀ * (σ * √(Λ / 3) * c) * (σ * √(Λ / 3) * c) *
+        Real.exp (σ * √(Λ / 3) * c * t.val) := by
+  rw [deriv_deSitterScaleFactor]
+  funext t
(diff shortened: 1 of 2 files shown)


Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -79,28 +92,33 @@ lemma deriv_comp_val {γ : ℝ → ℝ} {t : Time} {v : ℝ} (h : HasDerivAt γ
 noncomputable def deSitterScaleFactor (a₀ σ Λ c : ℝ) : Time → ℝ :=
   fun t => a₀ * Real.exp (σ * √(Λ / 3) * c * t.val)
 
-/-- Mathlib derivative of `y ↦ a₀ exp (K y)`. -/
-lemma hasDerivAt_mul_exp_mul (a₀ K x : ℝ) :
```
doxtor6, line 83: same here, see my other comment in another PR

doxtor6 (commented):
General remark, then per-`def` comments inline.

`def`s are the load-bearing part of the library: they are what downstream files refer to by name, so we only add them for objects that deserve a name, and those need to be high quality. This series adds 9 `def`s to this file; I think 4 should stay (the four named scale factors) and the rest should be inlined into the theorem statements. The two `_eq` glue lemmas (`radiationDensity_eq`, `einsteinDeSitterDensity_eq`) are the symptom: they exist only to undo defs.

For the defs that stay, two points on quality: `t.val = 0` hard-codes the singularity at the origin of the chart, whereas `Time` is an `AddTorsor ℝ Time` with no distinguished origin; and `t₀` is a time, so it should be `t₀ : Time` rather than `ℝ` (the `⟨t₀⟩` in `hubbleConstant_radiationScaleFactor_t₀` is the tell).

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+-/
+
+/-!
+
+### B.1. The power-law scale factor and its derivatives
+
+-/
+
+/-- The power-law scale factor `a(t) = (t / t₀) ^ n` (real power). -/
+noncomputable def powerLawScaleFactor (t₀ n : ℝ) : Time → ℝ :=
```
doxtor6: Is this def necessary? It is proof-sharing scaffolding rather than a named solution. The general-`n` lemmas below work just as well quantified over `n` with `fun t => (t.val / t₀) ^ n` inline in the statement, and then `radiationScaleFactor` / `einsteinDeSitterScaleFactor` can be defined directly instead of as aliases that have to be unfolded.

If a power-law def survives anyway: `t.val` puts the Big Bang at the origin of the chart (`Time` is an `AddTorsor ℝ Time`, no distinguished origin), `t₀` should be a `Time`, and the docstring should say that `t₀` is the normalisation epoch (`a(t₀) = 1`) and that values for `t ≤ 0` are junk (`Real.rpow` on a negative base).

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  rw [deriv_deriv_powerLawScaleFactor ht₀.ne' n ht, deriv_powerLawScaleFactor ht₀.ne' n ht.ne',
+    powerLawScaleFactor, Real.rpow_sub_one (div_pos ht ht₀).ne',
+    Real.rpow_sub_one (div_pos ht ht₀).ne']
+  have hx : (t.val / t₀) ^ n ≠ 0 := (Real.rpow_pos_of_pos (div_pos ht ht₀) n).ne'
+  field_simp
+  ring
+
+/-- The density imposed on the flat power-law solution by the first-order Friedmann equation,
+  `ρ = 3 n² / (8 π G t²)`. -/
+noncomputable def powerLawDensity (G n : ℝ) : Time → ℝ :=
```
doxtor6: Same question. This only ever appears as an argument to its own lemmas. The density is forced by the scale factor, so the reader is better served by seeing `fun t => 3 * n ^ 2 / (8 * π * G * t.val ^ 2)` in the theorem statement than by chasing a name.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  ring
+
+/-!
+
+### B.3. The radiation-dominated solution
+
+-/
+
+/-- The radiation-dominated scale factor `a(t) = (t / t₀) ^ (1/2)`. -/
+noncomputable def radiationScaleFactor (t₀ : ℝ) : Time → ℝ :=
```
doxtor6: Keep this one (it is the named solution the TODO asks for), but define it directly as `fun t => (t.val / t₀) ^ (1 / 2 : ℝ)` rather than as an alias of `powerLawScaleFactor`. Same remarks on `t₀ : Time` and the origin of `t.val`.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  powerLawScaleFactor t₀ (1 / 2)
+
+/-- The density of the flat radiation-dominated solution, `ρ = 3 / (32 π G t²)`, imposed by
+  the first-order Friedmann equation. -/
+noncomputable def radiationDensity (G : ℝ) : Time → ℝ :=
```
doxtor6: Not an independent object: the density is forced by the scale factor. Please put the formula in the statement of `radiationScaleFactor_firstOrderFriedmann` and drop this def; `radiationDensity_eq` then goes with it.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  fun t => 3 / (32 * π * G * t.val ^ 2)
+
+/-- The radiation pressure `p = ρ c² / 3`. -/
+noncomputable def radiationPressure (G c : ℝ) : Time → ℝ :=
```
doxtor6: Same: inline `fun t => ρ t * c ^ 2 / 3` in the second-order statement.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  ring
+
+/-!
+
+### B.4. The Einstein-de Sitter solution
+
+-/
+
+/-- The Einstein-de Sitter (flat, dust) scale factor `a(t) = (t / t₀) ^ (2/3)`. -/
+noncomputable def einsteinDeSitterScaleFactor (t₀ : ℝ) : Time → ℝ :=
```
doxtor6: As for `radiationScaleFactor`: keep, but define directly as `fun t => (t.val / t₀) ^ (2 / 3 : ℝ)`, with `t₀ : Time`.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+  powerLawScaleFactor t₀ (2 / 3)
+
+/-- The dust density of the Einstein-de Sitter solution, `ρ = 1 / (6 π G t²)`, imposed by the
+  first-order Friedmann equation. -/
+noncomputable def einsteinDeSitterDensity (G : ℝ) : Time → ℝ :=
```
doxtor6: As for `radiationDensity`: inline in the theorem statement and drop, together with `einsteinDeSitterDensity_eq`.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+## C. The Milne solution
+
+The Milne universe is the empty (`ρ = 0`, `p = 0`, `Λ = 0`) solution with `k = -1` and
+`a(t) = c t` for `t > 0`. That it is Minkowski space in expanding coordinates (vanishing
+curvature) is not stated here: the FLRW metric is not yet an object of Physlib.
+
+-/
+
+/-- The Milne scale factor `a(t) = c t`. -/
+noncomputable def milneScaleFactor (c : ℝ) : Time → ℝ :=
```
doxtor6, line 439: Fine to keep as a named model (the curvature TODO still points at it). Same origin remark: `c * t.val` puts the Bang at `t.val = 0` of the chart.

Physlib/Cosmology/FLRW/Solutions.lean
```diff
@@ -178,21 +196,346 @@ lemma decelerationParameter_deSitterScaleFactor {a₀ σ Λ c : ℝ} (hΛ : 0 <
...
+force the density `ρ = Λ c² / (4 π G)`, twice the density `ρ_Λ = Λ c² / (8 π G)` associated
+with the cosmological constant, and `k c² / a² = 4 π G ρ`, hence a positive curvature
+parameter when `ρ > 0`. That this equilibrium is unstable is not stated here.
 
 -/
 
-TODO "Prove that the spatially flat radiation-dominated solution `a = (t/t₀)^(1/2)`
-  solves the Friedmann equations, with `q₀ = 1` and `t₀ = 1 / (2 H₀)`."
+/-- The density `ρ_Λ = Λ c² / (8 π G)` associated with the cosmological constant. -/
+noncomputable def cosmologicalConstantDensity (Λ G c : ℝ) : ℝ :=
```
doxtor6: This is a real object but the wrong home and the wrong scope: `Physlib/Cosmology/FLRW/MatterContent.lean` already has a TODO for exactly `ρ_Λ = Λ c² / (8 π G)` together with its `w = −1` content. Adding it here just to say `ρ = 2 ρ_Λ` means it will end up defined twice. Please state `ρ t = Λ * c ^ 2 / (4 * π * G)` directly and leave `ρ_Λ` to the PR that does that TODO.

doxtor6 (approved):
This PR is good and ready to merge

jstoobysmith:
Merge conflict on this PR now. @KevorkianPhilippe could you sort this?

KevorkianPhilippe:
Sorted: master merged into this PR and into the stacked #1635 to #1641. After the squash of #1633, `Solutions.lean` and `Time/Derivatives.lean` on master are identical to the reviewed `flrw-solutions-2`, so the conflict was resolved by keeping the stacked version of `Solutions.lean`; the FLRW modules build locally on the tip of the chain.

PR #1635, opened 2026-09-10
feat(Cosmology): continuity equation of the cosmic fluid from the Friedmann equations

KevorkianPhilippe:
## AI disclosure

specification.

#1634 is merged: this PR is now the first of the series and its diff is its own.

Resolves the continuity-equation and non-independence TODOs of
`Physlib/Cosmology/FLRW/MatterContent.lean`.

Declarations added (namespace `Cosmology.FLRW.FriedmannEquation`):
- `Time.hasDerivAt_mk_of_differentiableAt` (in `SpaceAndTime/Time/Derivatives.lean`): for `f : Time → ℝ` differentiable at `t`, the curve
  `σ ↦ f ⟨σ⟩` has Mathlib derivative `∂ₜ f t` at `t.val` (wrapper of
  `Time.hasDerivAt_comp_toRealCLE_symm`).
- `ContinuityEquation`: `∂ₜ ρ + 3 H (ρ + p / c²) = 0` at an instant.
- `deriv_firstOrderFriedmann`: differentiating the first-order Friedmann equation (assumed at
  all times): `2 H (a''/a - H²) = (8 π G / 3) ρ' + 2 k c² H / a²`.
- `continuityEquation_of_friedmann`: continuity from the first-order equation (at all times)
  and the second-order equation at `t`, for `a` twice differentiable and `ρ` differentiable
  at `t`, `a t ≠ 0`, `G ≠ 0`.
- `secondOrderFriedmann_of_continuityEquation`: the converse, second-order equation from the
  first-order equation and continuity, when `∂ₜ a t ≠ 0`. Together: the three equations are
  not independent (the Bianchi identity itself is not stated, there is no metric in Physlib).

Reviewer map: B.1 (definition), B.2 (the derivative lemma, the technical part), B.3 and B.4
(two `linear_combination`s).

**Update (Sept 17).** Review by doxtor6: `Time.hasDerivAt_mk_of_differentiableAt` and `deriv_firstOrderFriedmann` are stated at `t : Time`; `Time.deriv_comp_val` is dropped (special case of `Time.deriv_eq_of_hasDerivAt`, which replaces it in `Solutions.lean`); `continuityEquation_of_friedmann` only asks for `G ≠ 0`; the narrowed Bianchi TODO, the fluid-energy-equation TODO, the design note on the future `FLRW` structure and the `API-map.yaml` entry are added. #1634 is merged, so this PR is now the first of the series.

### DIFF
--- a/Physlib/Cosmology/FLRW/MatterContent.lean
+++ b/Physlib/Cosmology/FLRW/MatterContent.lean
@@ -1,34 +1,177 @@
 /-
 Copyright (c) 2026 Jinzheng Li. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Jinzheng Li
+Authors: Jinzheng Li, Philippe Kevorkian
 -/
 module
 
 public import Physlib.Meta.TODO.Basic
+public import Physlib.Cosmology.FLRW.Basic
 /-!
 
 # Matter content of FLRW cosmology
 
-Placeholder file collecting TODO items for the matter content of a
-Friedmann-Lemaître-Robertson-Walker universe: the perfect-fluid stress-energy
-tensor, the equation of state, the continuity equation, and the resulting
-density scaling for the standard components (dust, radiation, vacuum energy).
-This file contains only TODO items; no definitions or lemmas are formalized yet.
+## i. Overview
+
+This file describes the matter content of a Friedmann-Lemaître-Robertson-Walker universe
+through the continuity equation `∂ₜ ρ + 3 H (ρ + p / c²) = 0` of the cosmic fluid and its
+relation to the two Friedmann equations of `Physlib.Cosmology.FLRW.Basic`: the continuity
+equation follows from the first-order equation (holding at all times) and the second-order
+equation, and conversely the second-order equation follows from the first-order equation and
+the continuity equation whenever `∂ₜ a ≠ 0`. The three equations are therefore not
+independent. The equation of state, the density scaling laws and the perfect-fluid
+stress-energy tensor are still TODO items.
+
+Time derivatives of curves `Time → ℝ` are computed through the bridge
+`Time.hasDerivAt_comp_toRealCLE_symm` to Mathlib's `HasDerivAt` on `ℝ`.
+
+## ii. Key results
+
+- `ContinuityEquation`: the continuity equation at an instant `t`.
+- `deriv_firstOrderFriedmann`: the time derivative of the first-order Friedmann equation,
+  `2 H (a''/a - H²) = (8 π G / 3) ρ' + 2 k c² H / a²`.
+- `continuityEquation_of_friedmann`: the continuity equation follows from the two Friedmann
+  equations.
+- `secondOrderFriedmann_of_continuityEquation`: the second-order Friedmann equation follows
+  from the first-order equation and the continuity equation when `∂ₜ a ≠ 0`.
+
+## iii. Table of contents
+
+- A. Time derivatives through the coordinate
+- B. The continuity equation
+  - B.1. The definition
+  - B.2. The derivative of the first-order Friedmann equation
+  - B.3. Continuity from the Friedmann equations
+  - B.4. The second-order Friedmann equation from continuity
+- C. Remaining TODO items
 
 -/
 
 @[expose] public section
 
+namespace Cosmology.FLRW.FriedmannEquation
+
+open Real Time
+
+/-!
+
+## A. Time derivatives through the coordinate
+
+-/
+
+/-- A curve `f : Time → ℝ` differentiable at `⟨τ⟩` gives the curve `σ ↦ f ⟨σ⟩` on `ℝ`,
+  whose Mathlib derivative at `τ` is the time derivative `∂ₜ f ⟨τ⟩`. -/
+lemma hasDerivAt_mk_of_differentiableAt {f : Time → ℝ} {τ : ℝ}
+    (hf : DifferentiableAt ℝ f ⟨τ⟩) :
+    HasDerivAt (fun σ : ℝ => f ⟨σ⟩) (∂ₜ f ⟨τ⟩) τ :=
+  hasDerivAt_comp_toRealCLE_symm f τ hf
+
+/-!
+
+## B. The continuity equation
+
+-/
+
+/-!
+
+### B.1. The definition
+
+-/
+
+/-- The continuity equation of the cosmic fluid at the instant `t`:
+  `∂ₜ ρ + 3 H (ρ + p / c²) = 0`, with `H = hubbleConstant a`. -/
+def ContinuityEquation (a ρ p : Time → ℝ) (c : ℝ) (t : Time) : Prop :=
+  ∂ₜ ρ t + 3 * hubbleConstant a t * (ρ t + p t / c ^ 2) = 0
+
+/-!
+
+### B.2. The derivative of the first-order Friedmann equation
+
+-/
+
+/-- Differentiating the first-order Friedmann equation, assumed at all times, at `⟨τ⟩`:
+  `2 H (a''/a - H²) = (8 π G / 3) ρ' + 2 k c² H / a²`. -/
+lemma deriv_firstOrderFriedmann {a ρ : Time → ℝ} {k Λ G c τ : ℝ} (ha : a ⟨τ⟩ ≠ 0)
+    (hd1 : DifferentiableAt ℝ a ⟨τ⟩) (hd2 : DifferentiableAt ℝ (∂ₜ a) ⟨τ⟩)
+    (hdρ : DifferentiableAt ℝ ρ ⟨τ⟩) (hF1 : ∀ s, FirstOrderFriedmann a ρ k Λ G c s) :
+    2 * (∂ₜ a ⟨τ⟩ / a ⟨τ⟩) * (∂ₜ (∂ₜ a) ⟨τ⟩ / a ⟨τ⟩ - (∂ₜ a ⟨τ⟩ / a ⟨τ⟩) ^ 2)
+      = 8 * π * G / 3 * ∂ₜ ρ ⟨τ⟩
+        + 2 * k * c ^ 2 * (∂ₜ a ⟨τ⟩ / a ⟨τ⟩) / (a ⟨τ⟩) ^ 2 := by
+  have hA := hasDerivAt_mk_of_differentiableAt hd1
+  have hA' := hasDerivAt_mk_of_differentiableAt hd2
+  have hR := hasDerivAt_mk_of_differentiableAt hdρ
+  have hL := (hA'.div hA ha).pow 2
+  have hRd := ((hR.const_mul (8 * π * G / 3)).sub
+    ((hasDerivAt_const τ (k * c ^ 2)).div (hA.pow 2) (pow_ne_zero 2 ha))).add_const
+    (Λ * c ^ 2 / 3)
+  have hu := hL.unique
+    (hRd.congr_of_eventuallyEq (Filter.Eventually.of_forall fun σ => hF1 ⟨σ⟩))
+  simp only [Pi.pow_apply, Pi.div_apply, Nat.cast_ofNat, Nat.add_one_sub_one, pow_one, zero_mul,
+    zero_sub] at hu
+  field_simp at hu ⊢
+  linear_combination hu
+
+/-!
+
+### B.3. Continuity from the Friedmann equations
+
+-/
+
+/-- The continuity equation follows from the first-order Friedmann equation, assumed at all
+  times, and the second-order Friedmann equation at `t`; `a` must be twice differentiable and
+  `ρ` differentiable at `t`. -/
+lemma continuityEquation_of_friedmann {a ρ p : Time → ℝ} {k Λ G c : ℝ} {t : Time}
+    (hG : 0 < G) (ha : a t ≠ 0) (hd1 : DifferentiableAt ℝ a t)
+    (hd2 : DifferentiableAt ℝ (∂ₜ a) t)
+    (hdρ : DifferentiableAt ℝ ρ t) (hF1 : ∀ s, FirstOrderFriedmann a ρ k Λ G c s)
+    (hF2 : SecondOrderFriedmann a ρ p Λ G c t) :
+    ContinuityEquation a ρ p c t := by
+  obtain ⟨τ⟩ := t
+  have hd := deriv_firstOrderFriedmann ha hd1 hd2 hdρ hF1
+  have h1 := hF1 ⟨τ⟩
+  unfold FirstOrderFriedmann at h1
+  unfold SecondOrderFriedmann at hF2
+  unfold ContinuityEquation hubbleConstant
+  have hπ := Real.pi_pos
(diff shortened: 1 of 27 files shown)


doxtor6:
claim

doxtor6 (commented):
Thanks — the mathematics is right (I rederived both directions by hand: `2 H H' = (8πG/3) ρ' + 2 k c² H / a²` with `H' = a''/a − H²`, then both Friedmann equations give exactly `ρ' + 3H(ρ + p/c²) = 0`), the `∂ₜ a ≠ 0` hypothesis on the converse is the right one, and the `def` is justified: the fluid sector (#1636) consumes `ContinuityEquation` as a hypothesis in eight statements, and it takes no `G`, `Λ`, `k`, which is the correct physics (matter conservation is more primitive than the Friedmann equations, it is only *derived* from them via Einstein + Bianchi). Builds clean here, CI green.

One design point worth recording, because the whole series is building the API that the future `FLRW` structure (`Basic.lean:104` TODO, untouched by the stack) will have to bundle. Any two of {F1, F2, continuity} imply the third, but not symmetrically: F1 + continuity leave `p` completely unconstrained on any interval where `∂ₜ a = 0` (continuity collapses to `∂ₜ ρ = 0`, F1 to a `p`-free relation), while F2 gives `ρ + 3p/c² = Λc²/(4πG)` there — that is exactly the Einstein static universe, and `einsteinStatic_density` in `Solutions.lean` uses F2 at such an instant. So when `FLRW` becomes a structure it should carry F1 and F2 as fields and obtain continuity as a lemma via `continuityEquation_of_friedmann`, never the other way round. Please add that as a sentence to the `FLRW` TODO in `Basic.lean` (or the overview here).

Not in the diff, so here rather than inline: `Physlib/Cosmology/FLRW/API-map.yaml` still lists the continuity equation inside a `done: false, location: N/A` requirement (bundled with the equation of state and scaling laws). Please split that entry so the continuity-equation half reads `done: true` with `location: "Physlib/Cosmology/FLRW/MatterContent.lean (ContinuityEquation, deriv_firstOrderFriedmann, continuityEquation_of_friedmann, secondOrderFriedmann_of_continuityEquation)"`, leaving the rest `done: false` for #1636.

Inline comments follow; the one that matters for the rest of the library is the `⟨τ⟩` statement shape in `Time/Derivatives.lean`.

Physlib/SpaceAndTime/Time/Derivatives.lean
```diff
@@ -126,6 +126,13 @@ lemma deriv_eq_of_hasDerivAt {f : Time → ℝ} {t : Time} {v : ℝ}
     (h : HasDerivAt (fun τ : ℝ => f ⟨τ⟩) v t.val) : ∂ₜ f t = v :=
   deriv_comp_toRealCLE_of_hasDerivAt (fun τ : ℝ => f ⟨τ⟩) t v h
 
+/-- A curve `f : Time → ℝ` differentiable at `⟨τ⟩` gives the curve `σ ↦ f ⟨σ⟩` on `ℝ`,
+  whose Mathlib derivative at `τ` is the time derivative `∂ₜ f ⟨τ⟩`. -/
+lemma hasDerivAt_mk_of_differentiableAt {f : Time → ℝ} {τ : ℝ}
```
doxtor6: Please state this at a `t : Time` rather than at `⟨τ⟩` with `τ : ℝ`:

```lean
lemma hasDerivAt_mk_of_differentiableAt {f : Time → ℝ} {t : Time}
    (hf : DifferentiableAt ℝ f t) :
    HasDerivAt (fun σ : ℝ => f ⟨σ⟩) (∂ₜ f t) t.val :=
  hasDerivAt_comp_toRealCLE_symm f t.val hf
```

(verified: same proof term.) The `⟨τ⟩` form leaks the proof technique into the shared `Time` API and forces every caller to `obtain ⟨τ⟩ := t` first, which is what both theorems in `MatterContent.lean` end up doing.

While you are in this hunk: `deriv_comp_val` (line 119, from #1633) is an exact special case of `deriv_eq_of_hasDerivAt` (line 125) — for `f := fun s => γ s.val` the curve `fun τ => f ⟨τ⟩` is definitionally `γ`, so `deriv_eq_of_hasDerivAt h` closes every goal `deriv_comp_val h` does. Two names for one fact in the `Time` API is one too many; please drop `deriv_comp_val`. I checked that in `Solutions.lean` this is

```
sed -i 's/deriv_comp_val (γ := fun _ => c)/deriv_eq_of_hasDerivAt (f := fun _ => c)/; s/deriv_comp_val/deriv_eq_of_hasDerivAt/g' Physlib/Cosmology/FLRW/Solutions.lean
```

and the file builds; nothing else in the library uses it. Also please list whichever bridges survive in the `## ii. Key results` block of this file (the new ones are not mentioned there).

Physlib/Cosmology/FLRW/MatterContent.lean
```diff
@@ -1,34 +1,163 @@
...
+
+/-!
+
+### A.2. The derivative of the first-order Friedmann equation
+
+-/
+
+/-- Differentiating the first-order Friedmann equation, assumed at all times, at `⟨τ⟩`:
+  `2 H (a''/a - H²) = (8 π G / 3) ρ' + 2 k c² H / a²`. -/
+lemma deriv_firstOrderFriedmann {a ρ : Time → ℝ} {k Λ G c τ : ℝ} (ha : a ⟨τ⟩ ≠ 0)
```
doxtor6: Same point as in `Derivatives.lean`: please state this at `{t : Time}` with `a t`, `∂ₜ a t`, … rather than at `⟨τ⟩`, and do the `obtain ⟨τ⟩ := t` inside the proof. Verified that the current proof goes through unchanged after the destructuring, and the two callers below then no longer need to destructure `t` themselves.

Physlib/Cosmology/FLRW/MatterContent.lean
```diff
@@ -1,34 +1,163 @@
...
+
+### A.3. Continuity from the Friedmann equations
+
+-/
+
+/-- The continuity equation follows from the first-order Friedmann equation, assumed at all
+  times, and the second-order Friedmann equation at `t`; `a` must be twice differentiable and
+  `ρ` differentiable at `t`. -/
+lemma continuityEquation_of_friedmann {a ρ p : Time → ℝ} {k Λ G c : ℝ} {t : Time}
+    (hG : 0 < G) (ha : a t ≠ 0) (hd1 : DifferentiableAt ℝ a t)
```
doxtor6: Nit: only `G ≠ 0` is used here (to cancel `8 π G / 3`); `0 < G` is fine if you prefer file-wide consistency with the other lemmas, just noting it is stronger than needed.

Physlib/Cosmology/FLRW/MatterContent.lean
```diff
@@ -1,34 +1,163 @@
...
+
+### A.4. The second-order Friedmann equation from continuity
+
+-/
+
+/-- The second-order Friedmann equation at `t` follows from the first-order Friedmann
+  equation, assumed at all times, and the continuity equation at `t`, provided `∂ₜ a t ≠ 0`.
+  Together with `continuityEquation_of_friedmann`, the three equations are not independent. -/
+lemma secondOrderFriedmann_of_continuityEquation {a ρ p : Time → ℝ} {k Λ G c : ℝ}
+    {t : Time} (ha : a t ≠ 0) (hd1' : ∂ₜ a t ≠ 0) (hd1 : DifferentiableAt ℝ a t)
```
doxtor6: Nit: `hd1'` names a non-vanishing hypothesis, not a differentiability one — `ha'` (next to `ha : a t ≠ 0`) reads better.

Physlib/Cosmology/FLRW/MatterContent.lean
```diff
@@ -1,34 +1,163 @@
...
+through the continuity equation `∂ₜ ρ + 3 H (ρ + p / c²) = 0` of the cosmic fluid and its
+relation to the two Friedmann equations of `Physlib.Cosmology.FLRW.Basic`: the continuity
+equation follows from the first-order equation (holding at all times) and the second-order
+equation, and conversely the second-order equation follows from the first-order equation and
+the continuity equation whenever `∂ₜ a ≠ 0`. The three equations are therefore not
+independent. The equation of state, the density scaling laws and the perfect-fluid
+stress-energy tensor are still TODO items.
+
+Time derivatives of curves `Time → ℝ` are computed through the bridge
+`Time.hasDerivAt_comp_toRealCLE_symm` to Mathlib's `HasDerivAt` on `ℝ`.
```
doxtor6: Nit: the proofs go through the wrapper `hasDerivAt_mk_of_differentiableAt`, not the generic bridge; name the one the reader will actually find in the proofs.

Physlib/Cosmology/FLRW/MatterContent.lean
```diff
@@ -1,34 +1,163 @@
...
+  apply mul_left_cancel₀ hH
+  linear_combination hd + 8 * π * G / 3 * hC + 2 * (∂ₜ a ⟨τ⟩ / a ⟨τ⟩) * h1
+
+/-!
+
+## B. Remaining TODO items
+
+-/
+
+TODO "Define the perfect-fluid stress-energy tensor
```
doxtor6, line 159: Two TODOs to keep in this section, in the spirit of the narrowed TODOs of `Solutions.lean`:

1. The old "not independent (a consequence of the Bianchi identity `∇_ν Gᵘᵛ = 0`)" item was deleted rather than narrowed. The non-independence is now proved, but the Bianchi-identity derivation is not (no metric yet, as the PR body says). Please keep a narrowed TODO: derive the non-independence from `∇_ν Gᵘᵛ = 0` once the FLRW metric exists.

2. A bridge to fluid dynamics, so it is not forgotten: `ContinuityEquation a ρ p c` is the fluid *energy* equation `∂ₜ ρ + ∇ ⬝ (ρ u) + (p / c²) ∇ ⬝ u = 0` for the Hubble flow `u = H x` on `Space 3` with homogeneous `ρ`, `p` (Newtonian cosmology with pressure, McCrea–Milne; `∇ ⬝ (ρ u) = 3Hρ`, `∇ ⬝ u = 3H`). `Physlib.FluidDynamics` currently has only the mass equation `FluidFlow.ClassicalContinuityEquation` (no pressure work term), so this needs an energy equation there first; a TODO here recording the statement is enough for now.

Physlib/SpaceAndTime/Time/Derivatives.lean
KevorkianPhilippe: Done in e29b83e8: stated at `t : Time` with the proof term you gave, `deriv_comp_val` dropped (your `sed` on `Solutions.lean`; nothing else used it), and both real-valued bridges are now in the key results block.

Physlib/Cosmology/FLRW/MatterContent.lean
KevorkianPhilippe: Done in e29b83e8; the proof goes through unchanged after `obtain ⟨τ⟩ := t`, and the two callers no longer destructure `t`.

Physlib/Cosmology/FLRW/MatterContent.lean
KevorkianPhilippe: Took `G ≠ 0` (e29b83e8); the file already states the `_iff_lambdaFluid` lemmas of #1636 with `G ≠ 0`, so the weakest hypothesis is the consistent choice here.

Physlib/Cosmology/FLRW/MatterContent.lean
KevorkianPhilippe: Renamed to `ha'` (e29b83e8).

Physlib/Cosmology/FLRW/MatterContent.lean
KevorkianPhilippe: The overview now names `Time.hasDerivAt_mk_of_differentiableAt` (e29b83e8).

Physlib/Cosmology/FLRW/MatterContent.lean
KevorkianPhilippe: Both TODOs added in 0235ed41, the Bianchi one in narrowed form (derive the non-independence from `∇_ν Gᵘᵛ = 0` once the FLRW metric is defined), the fluid one with the statement you gave.

KevorkianPhilippe:
Thanks for the careful review, and for rederiving both directions. Everything is addressed in two commits on top of a merge of master (#1634 is in):

- e29b83e8 (code): `Time.hasDerivAt_mk_of_differentiableAt` and `deriv_firstOrderFriedmann` are stated at `t : Time`; the destructuring now happens once, inside the proof of `deriv_firstOrderFriedmann`, and the two theorems below it no longer destructure `t`. `Time.deriv_comp_val` is dropped and `Solutions.lean` uses `deriv_eq_of_hasDerivAt` (your `sed`, verbatim); the two surviving real-valued bridges are listed in the key results of `Time/Derivatives.lean`. Nits: `continuityEquation_of_friedmann` takes `G ≠ 0` (the file already uses the weakest hypothesis elsewhere, e.g. the `_iff_lambdaFluid` lemmas of #1636), `hd1'` is now `ha'`, and the overview names `Time.hasDerivAt_mk_of_differentiableAt`.
- 0235ed41 (docs): the Bianchi TODO is back in narrowed form, the fluid-energy-equation TODO is added with the statement you gave (`FluidFlow.ClassicalContinuityEquation` has no pressure work term, so it needs an energy equation in `Physlib.FluidDynamics` first), the design point is recorded both in the `FLRW` TODO of `Basic.lean` and in the overview of this file (F1 and F2 as fields, continuity as a lemma through `continuityEquation_of_friedmann`, never the other way round, with the Einstein static universe as the reason), and `API-map.yaml` has the continuity-equation requirement split off and marked done with the location you wrote. While there I fixed the overview of the map, which still said that `MatterContent.lean` and `Solutions.lean` contain only TODO items.

I left the McCrea-Milne attribution out of the TODO string because I have not gone back to the papers myself; glad to add a reference if you have the one you had in mind.

The rest of the stack (#1636 to #1641) is merged and rebuilt on top of this. Two things there follow from your remarks: in #1636 the one-off helper `hasDerivAt_mul_rpow_of_continuityEquation`, which was stated on the real curve `σ ↦ ρ ⟨σ⟩ * a ⟨σ⟩ ^ (3 * (1 + w))`, is inlined into `density_scaling`, so no statement of the series is posed at `⟨τ⟩` any more; what remains of the constructor form in statements is the integrand of the distance and conformal-time integrals (#1639, #1641), where the integration variable is a real number, and the chart origin `⟨0⟩` in the two Einstein-de Sitter closed forms (`particleHorizon_einsteinDeSitter`, `conformalTime_einsteinDeSitter`). And each PR now updates `API-map.yaml` for what it completes.

jstoobysmith:
@doxtor6 if you think you will have chance to look at this PR again, would you mind `claim`ing it

doxtor6:
> @doxtor6 if you think you will have chance to look at this PR again, would you mind `claim`ing it

Almost forget about this one, I will finish the review soon.

KevorkianPhilippe:
#1634 is merged, so this PR no longer depends on another one.

Drafted with Claude, reviewed and validated by me.

-blocked-by-PR

PR #1642, opened 2026-09-11
feat(Relativity): reverse Cauchy-Schwarz and triangle inequalities, twin paradox age gap nonnegative

KevorkianPhilippe:
## AI disclosure

**AI disclosure.** This PR was generated with Claude Fable 5.1 (Claude Code) under my
supervision, following `AI-POLICY.md` and `AGENTS.md`. I have read every definition and lemma
statement and vouch that they state what the docstrings say. The statements were fixed before
the proofs were written; the elementary inequality at the core was checked symbolically.

Turns the informal lemma `ageGap_nonneg` (tag 7ROVE) of
`Physlib/Relativity/Special/TwinParadox/Basic.lean` into a proved lemma, through the reverse
triangle inequality of Minkowski space, which was not in Physlib.

New file `Physlib/Relativity/Tensors/RealTensor/Vector/Causality/CausallyFollows.lean`
(namespace `Lorentz.Vector`), about vectors in the causal future of the origin, `causallyFollows 0 u`:
- `causallyFollows_zero_iff`: `causallyFollows 0 u ↔ 0 ≤ ⟪u, u⟫ₘ ∧ 0 ≤ u.timeComponent`;
  `causallyFollows_zero_sub`: `causallyFollows p q → causallyFollows 0 (q - p)`.
- `norm_spatialPart_le_timeComponent_of_causallyFollows`: `‖u_spatial‖ ≤ u⁰`.
- `sqrt_mul_sqrt_le_minkowskiProduct_of_causallyFollows`: the reverse Cauchy-Schwarz inequality
  `√⟪u, u⟫ₘ √⟪v, v⟫ₘ ≤ ⟪u, v⟫ₘ` for `u`, `v` in the causal future of the origin (Euclidean
  Cauchy-Schwarz on the spatial parts plus the elementary inequality
  `√(a² - c²) √(b² - d²) ≤ a b - c d` for `0 ≤ c ≤ a`, `0 ≤ d ≤ b`, inlined in the proof).
- `sqrt_add_sqrt_le_sqrt_add_of_causallyFollows`: the reverse triangle inequality
  `√⟪u, u⟫ₘ + √⟪v, v⟫ₘ ≤ √⟪u + v, u + v⟫ₘ`; `causallyFollows_zero_add_of_causallyFollows`.

Added to `MinkowskiProduct.lean`, next to `minkowskiProduct_self_eq_timeComponent_spatialPart`:
[...]

PR #1643, opened 2026-09-11
feat(Relativity): equality case of the reverse triangle inequality, twin paradox age gap zero

KevorkianPhilippe:
## AI disclosure

the proofs were written; the elementary inequality at the core was checked symbolically.

**Stacked on #1642** (`twin-paradox-11`): the diff includes its commit; only the last commit is new here.

Resolves the TODO "conditions for which the age gap is zero".

Added to `CausallyFollows.lean` (section D, the equality cases):
- `minkowskiProduct_eq_iff_of_causallyFollows`: for `u`, `v` in the causal future of the origin, equality
  `⟪u, v⟫ₘ = √⟪u, u⟫ₘ √⟪v, v⟫ₘ` holds if and only if `∃ μ ≥ 0, v = μ • u ∨ u = μ • v`
  (light-like cases included).
- `sqrt_add_eq_iff_of_causallyFollows`: equality in the reverse triangle inequality if and only if equality in the
  reverse Cauchy-Schwarz inequality.

Added next to the definitions they are about: `minkowskiProduct_smul_self`,
`sqrt_minkowskiProduct_smul_self` in `MinkowskiProduct.lean`; `eq_zero_of_timeComponent_of_spatialPart`,
`apply_inr_eq_zero_of_norm_spatialPart` in `Vector/Basic.lean`.

Added to `TwinParadox/Basic.lean`:
- `ageGap_eq_zero_iff`: `T.ageGap = 0 ↔ ∃ μ ∈ [0, 1], T.twinBMid = T.startPoint +
  μ • (T.endPoint - T.startPoint)`, i.e. the age gap vanishes exactly when Twin B does not turn.
- `example1_ageGap_ne_zero`.

Reviewer map: `minkowskiProduct_eq_iff_of_causallyFollows` (the case analysis, the longest proof), then
`sqrt_add_eq_iff_of_causallyFollows`, then `ageGap_eq_zero_iff`.

**Update (Sept 11).** Review remarks of #1642 applied (third commit): `causallyFollows 0 u` instead of a new predicate, lemmas placed next to the definitions of their objects, file renamed `CausallyFollows.lean`.

**Update (Sept 11, later).** Second review remark applied: the lemmas whose hypotheses are `causallyFollows 0 u` now end with `_of_causallyFollows`.

**Update (Sept 11, evening).** Review remark applied: the expression of the age gap in terms of the two legs of Twin B is now its own lemma, `ageGap_eq`, used by `ageGap_eq_zero_iff` and by the merged `ageGap_nonneg`.


Physlib/Relativity/Special/TwinParadox/Basic.lean
```diff
@@ -66,7 +66,54 @@ def properTimeTwinB : ℝ := SpaceTime.properTime T.startPoint T.twinBMid +
...
+  equality case of the reverse triangle inequality (`sqrt_add_eq_iff_of_causallyFollows`,
+  `minkowskiProduct_eq_iff_of_causallyFollows`).
+  -/
+lemma ageGap_eq_zero_iff : T.ageGap = 0 ↔
+    ∃ μ ∈ Set.Icc (0 : ℝ) 1, T.twinBMid = T.startPoint + μ • (T.endPoint - T.startPoint) := by
+  have hu := causallyFollows_zero_sub T.twinBMid_causallyFollows_startPoint
+  have hv := causallyFollows_zero_sub T.endPoint_causallyFollows_twinBMid
+  have hsum : T.endPoint - T.startPoint
+      = (T.twinBMid - T.startPoint) + (T.endPoint - T.twinBMid) := by abel
+  have hgap : T.ageGap = √⟪(T.twinBMid - T.startPoint) + (T.endPoint - T.twinBMid),
```
jstoobysmith: I think it might be useful to extract this as its own separate lemma, as it might be useful for other circumstances as well.

jstoobysmith (commented):
One more comment from me here.

jstoobysmith (approved):
I've approved this. This looks good, and many thanks for splitting up the two different changes into different commits and linking to them here. I wonder if we should actually add something like that to our agents.md because it makes the review process easier, especially if one can click through the commits to see

PR #1648, opened 2026-09-11
feat(QuantumMechanics): Rydberg formula and spectral series of the hydrogen atom

KevorkianPhilippe:
Resolves the two TODOs of `QuantumMechanics/Hydrogen/Basic.lean` on the Rydberg formula and the Lyman, Balmer and Paschen series, in a new module `QuantumMechanics/Hydrogen/Rydberg.lean` (registered in `Physlib.lean`).

**What is defined.** The bound-state energies expected for the `d`-dimensional atom, as announced in the neighbouring TODO, `E_n = -Ry / (n + (d - 1) / 2) ^ 2` with the Rydberg energy `Ry = m k ^ 2 / (2 ℏ ^ 2)` (`rydbergEnergy`, `levelIndex`, `energyLevel`); the energy, frequency and wavelength of the photon emitted in a transition (`transitionEnergy`, `transitionFrequency`, `wavelength`, the speed of light being a parameter since physlib has no such constant in quantum mechanics); the Rydberg frequency `R_ν = m k ^ 2 / (4 π ℏ ^ 3)`; and the three series `lymanFrequency`, `balmerFrequency`, `paschenFrequency` (transitions down to the levels of index 0, 1, 2, i.e. `n = 1, 2, 3` in the usual numbering for `d = 3`).

**What is proved** (all by algebra, nothing depends on the Hamiltonian):
- the levels are negative and strictly increasing when `2 ≤ d` and `k ≠ 0` (`energyLevel_neg`, `energyLevel_lt_energyLevel`);
- the Rydberg formula in energy, frequency and wavenumber form: `transitionEnergy_eq`, `transitionFrequency_eq` (`ν = R_ν (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`), `wavelength_inv`, with `rydbergFrequency_eq : R_ν = Ry / h`;
- positivity and monotonicity of the transition frequencies (`transitionFrequency_pos`, `transitionFrequency_lt_transitionFrequency`);
- the series limit `tendsto_transitionFrequency : ν → R_ν / L n₁ ^ 2` as the upper level goes to infinity;
- in dimension 3, every Balmer line lies below every Lyman line (`balmerFrequency_lt_lymanFrequency`: the Balmer limit is `R_ν / 4`, the first Lyman line `3 R_ν / 4`). This is specific to low dimension: the same statement fails for `6 ≤ d`, which is why the lemma assumes `d = 3`.

**What is not proved.** That these levels are the point spectrum of the hydrogen Hamiltonian; the docstrings say so and the corresponding TODO is left in place.

Checks: `lake build`, `lint-style.sh`, `runPhyslibLinters`, `check_file_imports`, codespell; standard axioms only.

**AI disclosure.** The statements were frozen first (with `sorry`) and their proofs were produced by Kimi K2.7 Code (Moonshot AI) running in the DeepSeek Harness against physlib, then rewritten and shortened by Claude Fable 5.1 (Claude Code) for the physlib style; a type-identity check between the frozen statements and the final lemmas was compiled. I reviewed the statements, the docstrings and the proofs.

### DIFF
--- a/Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
+++ b/Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
@@ -1,0 +1,245 @@
+/-
+Copyright (c) 2026 Philippe Kevorkian. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Philippe Kevorkian
+-/
+module
+
+public import Physlib.QuantumMechanics.Hydrogen.Basic
+/-!
+
+# The Rydberg formula and the spectral series of the hydrogen atom
+
+## i. Overview
+
+The bound-state energies expected for the `d`-dimensional hydrogen atom with potential `-k / r`
+are `E_n = -Ry / (n + (d - 1) / 2) ^ 2` for `n : ℕ`, with the Rydberg energy
+`Ry = m k ^ 2 / (2 ℏ ^ 2)`. This module defines these Bohr levels and derives from them, by
+algebra alone, the Rydberg formula for the energy, the frequency and the wavelength of the
+photon emitted in a transition between two levels, together with the Lyman, Balmer and Paschen
+series and their series limits.
+
+The statement that these values are the point spectrum of the hydrogen Hamiltonian is not
+proved here; it is a separate TODO of the `Basic` module. Nothing in this module depends on
+the Hamiltonian.
+
+## ii. Key results
+
+- `energyLevel` defines the Bohr levels `E_n = -Ry / (n + (d - 1) / 2) ^ 2`; they are
+  negative and increasing in `n` when `2 ≤ d` and `k ≠ 0`.
+- `transitionFrequency_eq` is the Rydberg formula
+  `ν = R_ν (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)` with `R_ν = m k ^ 2 / (4 π ℏ ^ 3) = Ry / h`, and
+  `wavelength_inv` is its wavenumber form `1 / λ = (R_ν / c) (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`.
+- `tendsto_transitionFrequency` gives the series limit `ν → R_ν / L n₁ ^ 2`.
+- `lymanFrequency`, `balmerFrequency` and `paschenFrequency` define the three classical series;
+  `balmerFrequency_lt_lymanFrequency` shows that in dimension `3` every Balmer line lies below
+  every Lyman line.
+
+## iii. Table of contents
+
+- A. The Bohr levels
+- B. Transitions and the Rydberg formula
+- C. The spectral series
+
+## iv. References
+
+* None.
+-/
+
+@[expose] public section
+
+namespace QuantumMechanics
+namespace HydrogenAtom
+noncomputable section
+open Constants Real Filter Topology
+
+variable (H : HydrogenAtom)
+
+/-!
+
+## A. The Bohr levels
+
+The Rydberg energy `Ry = m k ^ 2 / (2 ℏ ^ 2)` sets the scale of the spectrum, and the level
+index `L n = n + (d - 1) / 2` is the effective principal quantum number in dimension `d`; for
+`d = 3` it is `n + 1`, the usual principal quantum number, with `n = 0` the ground state.
+
+-/
+
+/-- The Rydberg energy of the atom, `Ry = m k ^ 2 / (2 ℏ ^ 2)`. -/
+def rydbergEnergy : ℝ := H.m * H.k ^ 2 / (2 * (ℏ : ℝ) ^ 2)
+
+/-- The level index `n + (d - 1) / 2` of the `n`-th Bohr level; for `d = 3` it is the principal
+quantum number `n + 1`. -/
+def levelIndex (n : ℕ) : ℝ := n + ((H.d : ℝ) - 1) / 2
+
+/-- The `n`-th Bohr level `E_n = -Ry / (n + (d - 1) / 2) ^ 2`.
+
+These are the bound-state energies expected for the hydrogen Hamiltonian; that they form its
+point spectrum is not proved here. For `d = 1` and `n = 0` the level index vanishes and the value
+is `0` by the convention `x / 0 = 0`. -/
+def energyLevel (n : ℕ) : ℝ := -H.rydbergEnergy / H.levelIndex n ^ 2
+
+/-- The Rydberg energy is positive when `k ≠ 0`. -/
+lemma rydbergEnergy_pos (hk : H.k ≠ 0) : 0 < H.rydbergEnergy :=
+  div_pos (mul_pos H.m_pos (pow_two_pos_of_ne_zero hk)) (mul_pos two_pos (pow_pos ℏ_pos 2))
+
+/-- The level index is positive when `2 ≤ d`. -/
+lemma levelIndex_pos (hd : 2 ≤ H.d) (n : ℕ) : 0 < H.levelIndex n := by
+  have hd' : (2 : ℝ) ≤ H.d := by exact_mod_cast hd
+  have hn : (0 : ℝ) ≤ n := n.cast_nonneg
+  unfold levelIndex
+  linarith
+
+/-- The level index is strictly increasing. -/
+lemma levelIndex_lt_levelIndex {n₁ n₂ : ℕ} (h : n₁ < n₂) : H.levelIndex n₁ < H.levelIndex n₂ := by
+  have h' : (n₁ : ℝ) < n₂ := by exact_mod_cast h
+  unfold levelIndex
+  linarith
+
+/-- The Bohr levels are negative when `2 ≤ d` and `k ≠ 0`. -/
+lemma energyLevel_neg (hd : 2 ≤ H.d) (hk : H.k ≠ 0) (n : ℕ) : H.energyLevel n < 0 := by
+  rw [energyLevel, neg_div, neg_lt_zero]
+  exact div_pos (H.rydbergEnergy_pos hk) (pow_pos (H.levelIndex_pos hd n) 2)
+
+/-- The Bohr levels are strictly increasing when `2 ≤ d` and `k ≠ 0`. -/
+lemma energyLevel_lt_energyLevel (hd : 2 ≤ H.d) (hk : H.k ≠ 0) {n₁ n₂ : ℕ} (h : n₁ < n₂) :
+    H.energyLevel n₁ < H.energyLevel n₂ := by
+  unfold energyLevel
+  rw [neg_div, neg_div, neg_lt_neg_iff]
+  exact div_lt_div_of_pos_left (H.rydbergEnergy_pos hk) (pow_pos (H.levelIndex_pos hd n₁) 2)
+    (pow_lt_pow_left₀ (H.levelIndex_lt_levelIndex h) (H.levelIndex_pos hd n₁).le two_ne_zero)
+
+/-!
+
+## B. Transitions and the Rydberg formula
+
+A transition from the level `n₂` down to the level `n₁` emits a photon of energy
+`E_{n₂} - E_{n₁}`, frequency `ν = (E_{n₂} - E_{n₁}) / h` and wavelength `λ = c / ν`, where the
+speed of light `c` is taken as a parameter. The Rydberg formula expresses these through the
+Rydberg frequency `R_ν = m k ^ 2 / (4 π ℏ ^ 3) = Ry / h`.
+
+-/
+
+/-- The energy `E_{n₂} - E_{n₁}` of the transition from the level `n₂` to the level `n₁`. -/
+def transitionEnergy (n₁ n₂ : ℕ) : ℝ := H.energyLevel n₂ - H.energyLevel n₁
+
+/-- The frequency `(E_{n₂} - E_{n₁}) / h` of the photon emitted in the transition from `n₂`
+to `n₁`. -/
+def transitionFrequency (n₁ n₂ : ℕ) : ℝ := H.transitionEnergy n₁ n₂ / (h : ℝ)
+
+/-- The Rydberg frequency `R_ν = m k ^ 2 / (4 π ℏ ^ 3)`, equal to `Ry / h`. -/
+def rydbergFrequency : ℝ := H.m * H.k ^ 2 / (4 * π * (ℏ : ℝ) ^ 3)
+
+/-- The wavelength `c / ν` of the photon emitted in the transition from `n₂` to `n₁`, for a
+speed of light `c`. -/
+def wavelength (c : ℝ) (n₁ n₂ : ℕ) : ℝ := c / H.transitionFrequency n₁ n₂
+
+/-- The Rydberg formula for the transition energy,
+`E_{n₂} - E_{n₁} = Ry (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`; an identity valid for every `d`. -/
+lemma transitionEnergy_eq (n₁ n₂ : ℕ) :
+    H.transitionEnergy n₁ n₂ =
+      H.rydbergEnergy * (1 / H.levelIndex n₁ ^ 2 - 1 / H.levelIndex n₂ ^ 2) := by
+  unfold transitionEnergy energyLevel
+  ring
+
+/-- The Rydberg frequency equals `Ry / h`, with `h = 2 π ℏ`. -/
+lemma rydbergFrequency_eq : H.rydbergFrequency = H.rydbergEnergy / (h : ℝ) := by
+  rw [rydbergFrequency, rydbergEnergy, show (h : ℝ) = 2 * π * (ℏ : ℝ) from rfl]
+  field_simp
+  ring
+
+/-- The Rydberg frequency is positive when `k ≠ 0`. -/
+lemma rydbergFrequency_pos (hk : H.k ≠ 0) : 0 < H.rydbergFrequency := by
+  rw [H.rydbergFrequency_eq]
+  exact div_pos (H.rydbergEnergy_pos hk) h_pos
+
+/-- The Rydberg formula for the frequency, `ν = R_ν (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`. -/
+lemma transitionFrequency_eq (n₁ n₂ : ℕ) :
+    H.transitionFrequency n₁ n₂ =
+      H.rydbergFrequency * (1 / H.levelIndex n₁ ^ 2 - 1 / H.levelIndex n₂ ^ 2) := by
+  rw [transitionFrequency, H.transitionEnergy_eq, H.rydbergFrequency_eq]
+  ring
+
+/-- The frequency of a transition from `n₂` down to `n₁ < n₂` is positive when `2 ≤ d` and
+`k ≠ 0`. -/
+lemma transitionFrequency_pos (hd : 2 ≤ H.d) (hk : H.k ≠ 0) {n₁ n₂ : ℕ} (h : n₁ < n₂) :
+    0 < H.transitionFrequency n₁ n₂ :=
+  div_pos (sub_pos.mpr (H.energyLevel_lt_energyLevel hd hk h)) h_pos
+
+/-- The Rydberg formula for the wavenumber,
+`1 / λ = (R_ν / c) (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`; for `c = 0` both sides are `0`. -/
+lemma wavelength_inv (c : ℝ) (n₁ n₂ : ℕ) :
+    (H.wavelength c n₁ n₂)⁻¹ =
+      H.rydbergFrequency / c * (1 / H.levelIndex n₁ ^ 2 - 1 / H.levelIndex n₂ ^ 2) := by
+  rw [wavelength, inv_div, H.transitionFrequency_eq]
+  ring
+
+/-- For a fixed lower level, the transition frequency increases with the upper level when
+`2 ≤ d` and `k ≠ 0`. -/
+lemma transitionFrequency_lt_transitionFrequency (hd : 2 ≤ H.d) (hk : H.k ≠ 0) (n₁ : ℕ)
+    {n₂ n₃ : ℕ} (h : n₂ < n₃) : H.transitionFrequency n₁ n₂ < H.transitionFrequency n₁ n₃ := by
+  rw [H.transitionFrequency_eq, H.transitionFrequency_eq]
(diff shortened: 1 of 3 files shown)


Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+quantum number `n + 1`. -/
+def levelIndex (n : ℕ) : ℝ := n + ((H.d : ℝ) - 1) / 2
+
+/-- The `n`-th Bohr level `E_n = -Ry / (n + (d - 1) / 2) ^ 2`.
+
+These are the bound-state energies expected for the hydrogen Hamiltonian; that they form its
+point spectrum is not proved here. For `d = 1` and `n = 0` the level index vanishes and the value
+is `0` by the convention `x / 0 = 0`. -/
+def energyLevel (n : ℕ) : ℝ := -H.rydbergEnergy / H.levelIndex n ^ 2
+
```
gloges, line 81: For these definitions let's add lemmas like
```lean
lemma levelIndex_eq (n : ℕ) : H.levelIndex n = n + ((H.d : ℝ) - 1) / 2 := rfl
```
to the API so that we can avoid having to use `unfold` below.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+/-- The level index is positive when `2 ≤ d`. -/
+lemma levelIndex_pos (hd : 2 ≤ H.d) (n : ℕ) : 0 < H.levelIndex n := by
+  have hd' : (2 : ℝ) ≤ H.d := by exact_mod_cast hd
+  have hn : (0 : ℝ) ≤ n := n.cast_nonneg
+  unfold levelIndex
+  linarith
+
+/-- The level index is strictly increasing. -/
+lemma levelIndex_lt_levelIndex {n₁ n₂ : ℕ} (h : n₁ < n₂) : H.levelIndex n₁ < H.levelIndex n₂ := by
```
gloges, line 94:
```suggestion
lemma levelIndex_strictMono : StrictMono H.levelIndex := by
```

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/-- The `n`-th Bohr level `E_n = -Ry / (n + (d - 1) / 2) ^ 2`.
+
+These are the bound-state energies expected for the hydrogen Hamiltonian; that they form its
+point spectrum is not proved here. For `d = 1` and `n = 0` the level index vanishes and the value
+is `0` by the convention `x / 0 = 0`. -/
+def energyLevel (n : ℕ) : ℝ := -H.rydbergEnergy / H.levelIndex n ^ 2
+
+/-- The Rydberg energy is positive when `k ≠ 0`. -/
+lemma rydbergEnergy_pos (hk : H.k ≠ 0) : 0 < H.rydbergEnergy :=
```
gloges, line 83: I think it would be good to also have a slightly weaker version `rydbergEnergy_nonneg` which doesn't require the hypothesis `hk`.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+  div_pos (mul_pos H.m_pos (pow_two_pos_of_ne_zero hk)) (mul_pos two_pos (pow_pos ℏ_pos 2))
+
+/-- The level index is positive when `2 ≤ d`. -/
+lemma levelIndex_pos (hd : 2 ≤ H.d) (n : ℕ) : 0 < H.levelIndex n := by
```
gloges, line 87: Similarly here: also `levelIndex_nonneg` with the weaker hypothesis `hd : NeZero H.d`.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+  linarith
+
+/-- The level index is strictly increasing. -/
+lemma levelIndex_lt_levelIndex {n₁ n₂ : ℕ} (h : n₁ < n₂) : H.levelIndex n₁ < H.levelIndex n₂ := by
+  have h' : (n₁ : ℝ) < n₂ := by exact_mod_cast h
+  unfold levelIndex
+  linarith
+
+/-- The Bohr levels are negative when `2 ≤ d` and `k ≠ 0`. -/
+lemma energyLevel_neg (hd : 2 ≤ H.d) (hk : H.k ≠ 0) (n : ℕ) : H.energyLevel n < 0 := by
```
gloges, line 100: Similarly here.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+  rw [energyLevel, neg_div, neg_lt_zero]
+  exact div_pos (H.rydbergEnergy_pos hk) (pow_pos (H.levelIndex_pos hd n) 2)
+
+/-- The Bohr levels are strictly increasing when `2 ≤ d` and `k ≠ 0`. -/
+lemma energyLevel_lt_energyLevel (hd : 2 ≤ H.d) (hk : H.k ≠ 0) {n₁ n₂ : ℕ} (h : n₁ < n₂) :
+    H.energyLevel n₁ < H.energyLevel n₂ := by
```
gloges, line 106:
```suggestion
/-- The Bohr levels are strictly increasing when `2 ≤ d` and `k ≠ 0`. -/
lemma energyLevel_lt_energyLevel (hd : 2 ≤ H.d) (hk : H.k ≠ 0) : StrictMono H.energyLevel := by
```
and, as above, a variant proving `Monotone H.energyLevel` without the hypotheses on `H.d` and `H.k`.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/-- The energy `E_{n₂} - E_{n₁}` of the transition from the level `n₂` to the level `n₁`. -/
+def transitionEnergy (n₁ n₂ : ℕ) : ℝ := H.energyLevel n₂ - H.energyLevel n₁
+
+/-- The frequency `(E_{n₂} - E_{n₁}) / h` of the photon emitted in the transition from `n₂`
+to `n₁`. -/
+def transitionFrequency (n₁ n₂ : ℕ) : ℝ := H.transitionEnergy n₁ n₂ / (h : ℝ)
+
+/-- The Rydberg frequency `R_ν = m k ^ 2 / (4 π ℏ ^ 3)`, equal to `Ry / h`. -/
+def rydbergFrequency : ℝ := H.m * H.k ^ 2 / (4 * π * (ℏ : ℝ) ^ 3)
```
gloges, line 131:
```suggestion
/-- The Rydberg frequency `R_ν = m k ^ 2 / (4 π ℏ ^ 3)`, equal to `Ry / h`. -/
abbrev rydbergFrequency : ℝ := H.rydbergEnergy / h
```
I think it is cleaner to have this as an abbreviation which matches the definition `Ry / h`.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/-- The wavelength `c / ν` of the photon emitted in the transition from `n₂` to `n₁`, for a
+speed of light `c`. -/
+def wavelength (c : ℝ) (n₁ n₂ : ℕ) : ℝ := c / H.transitionFrequency n₁ n₂
```
gloges, line 135: Same comment as above - let's add "definitional lemmas" to avoid using `unfold` below.

gloges, line 135:
```suggestion
def wavelength (c : SpeedOfLight) (n₁ n₂ : ℕ) : ℝ := c / H.transitionFrequency n₁ n₂
```
Let's incorporate the speed of light available in `Relativity/SpeedOfLight.lean` to tie in with more of the library.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/-- The frequency of a transition from `n₂` down to `n₁ < n₂` is positive when `2 ≤ d` and
+`k ≠ 0`. -/
+lemma transitionFrequency_pos (hd : 2 ≤ H.d) (hk : H.k ≠ 0) {n₁ n₂ : ℕ} (h : n₁ < n₂) :
+    0 < H.transitionFrequency n₁ n₂ :=
+  div_pos (sub_pos.mpr (H.energyLevel_lt_energyLevel hd hk h)) h_pos
+
+/-- The Rydberg formula for the wavenumber,
+`1 / λ = (R_ν / c) (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`; for `c = 0` both sides are `0`. -/
+lemma wavelength_inv (c : ℝ) (n₁ n₂ : ℕ) :
```
gloges, line 171:
```suggestion
/-- The Rydberg formula for the wavenumber, `1 / λ = (R_ν / c) (1 / L n₁ ^ 2 - 1 / L n₂ ^ 2)`. -/
lemma wavelength_inv (c : SpeedOfLight) (n₁ n₂ : ℕ) :
```
`c : SpeedOfLight` does not require the caveat since it is always positive!

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+    (H.wavelength c n₁ n₂)⁻¹ =
+      H.rydbergFrequency / c * (1 / H.levelIndex n₁ ^ 2 - 1 / H.levelIndex n₂ ^ 2) := by
+  rw [wavelength, inv_div, H.transitionFrequency_eq]
+  ring
+
+/-- For a fixed lower level, the transition frequency increases with the upper level when
+`2 ≤ d` and `k ≠ 0`. -/
+lemma transitionFrequency_lt_transitionFrequency (hd : 2 ≤ H.d) (hk : H.k ≠ 0) (n₁ : ℕ)
+    {n₂ n₃ : ℕ} (h : n₂ < n₃) : H.transitionFrequency n₁ n₂ < H.transitionFrequency n₁ n₃ := by
```
gloges, line 180:
```suggestion
lemma transitionFrequency_strictMono (hd : 2 ≤ H.d) (hk : H.k ≠ 0) (n₁ : ℕ) :
    StrictMono (H.transitionFrequency n₁) := by
```
plus again a `Monotone` variant with weaker hypotheses.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+-/
+
+/-- The Lyman series: the transitions down to the level of index `0`. -/
+def lymanFrequency (n : ℕ) : ℝ := H.transitionFrequency 0 n
+
+/-- The Balmer series: the transitions down to the level of index `1`. -/
+def balmerFrequency (n : ℕ) : ℝ := H.transitionFrequency 1 n
+
+/-- The Paschen series: the transitions down to the level of index `2`. -/
+def paschenFrequency (n : ℕ) : ℝ := H.transitionFrequency 2 n
```
gloges, line 217: I think these should be abbreviations.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/-- In dimension `3`, every Balmer line has a lower frequency than every Lyman line: the Balmer
+series limit is `R_ν / 4` while the first Lyman line is at `3 R_ν / 4`. This fails for `6 ≤ d`. -/
+lemma balmerFrequency_lt_lymanFrequency (hd : H.d = 3) (hk : H.k ≠ 0) {n₂ n₃ : ℕ} (h₂ : 1 < n₂)
+    (h₃ : 0 < n₃) : H.balmerFrequency n₂ < H.lymanFrequency n₃ := by
```
gloges, line 222: As you say, this fails for `6 ≤ d` but it is true for `d = 2, 3, 4, 5` with essentially the same proof. Can this be adapted to account for these cases as well?

gloges (commented):
Thanks for making progress on this! Two general comments:
1. Best to avoid using `unfold` and rather have lemmas which express the definition and can be used with a rewrite.
2. The "positive" and "nonnegative" lemmas should probably all be marked with `@[simp]` (e.g. see the Planck constant file). This makes `simp` much more powerful when doing calculation-type proofs since these facts are then just automatically filled in when using something like `Real.sq_sqrt`.

KevorkianPhilippe:
Thanks for the careful review. Everything is applied in `9ca0eb64`; two places where I did not
follow the suggestion literally are flagged below (the `Monotone` variants and one now-unused
hypothesis).

**1. No more `unfold`.** Each definition now has its defining lemma and every proof goes through
a rewrite: `rydbergEnergy_eq`, `levelIndex_eq`, `energyLevel_eq`, `transitionEnergy_eq_sub`,
`transitionFrequency_eq_div`, `wavelength_eq_div`. The names `transitionEnergy_eq` and
`transitionFrequency_eq` were already taken by the Rydberg formulas, so for those two (and for
`wavelength`) the definitional lemma is named after the shape of its right-hand side.

**2. `@[simp]` on the sign lemmas**, following `PlanckConstant.lean`: `rydbergEnergy_nonneg`,
`rydbergEnergy_pos`, `levelIndex_nonneg`, `levelIndex_pos`, `energyLevel_nonpos`,
`energyLevel_neg`, `rydbergFrequency_nonneg`, `rydbergFrequency_pos`, `transitionFrequency_pos`.

**3. Weaker variants.** Added `rydbergEnergy_nonneg` (no hypothesis), `levelIndex_nonneg`,
`energyLevel_nonpos` (no hypothesis), `rydbergFrequency_nonneg`, `levelIndex_monotone`,
`energyLevel_monotone`, `transitionFrequency_monotone`.

For `levelIndex_nonneg` I took `NeZero H.d` as an instance argument rather than a named
hypothesis, so that the `@[simp]` form can discharge it by instance synthesis:

```lean
@[simp]
lemma levelIndex_nonneg [NeZero H.d] (n : ℕ) : 0 ≤ H.levelIndex n
```

**4. `StrictMono`.** `levelIndex_strictMono`, `energyLevel_strictMono` and
`transitionFrequency_strictMono` replace the three `_lt_` lemmas. I renamed
`energyLevel_lt_energyLevel` to `energyLevel_strictMono` (your suggestion kept the old name, but
the statement is now `StrictMono` and the neighbouring lemma is `levelIndex_strictMono`).

**5. The `Monotone` variants keep `2 ≤ H.d`.** The hypothesis `H.k ≠ 0` does drop (`k = 0` gives
`Ry = 0`, so every level is `0`), but `2 ≤ H.d` cannot: in dimension `1` the level index of the
ground state vanishes, `L 0 = 0 + (1 - 1) / 2 = 0`, so `E_0 = -Ry / 0 = 0` by the `x / 0 = 0`
convention while `E_1 = -Ry < 0` as soon as `k ≠ 0`. So `Monotone H.energyLevel` is false for
`d = 1`, and likewise `Monotone (H.transitionFrequency n₁)` is false there (`1 / L 0 ^ 2 = 0`
sits below `1 / L 1 ^ 2`). Dimension `0` happens to be harmless, but `2 ≤ d` is the clean
statement and it is the hypothesis the rest of the file uses. The docstrings say this.

**6. `rydbergFrequency` is now an `abbrev` for `Ry / h`**, as suggested. The closed form has
become a lemma, keeping the name `rydbergFrequency_eq`:

```lean
abbrev rydbergFrequency : ℝ := H.rydbergEnergy / h

lemma rydbergFrequency_eq : H.rydbergFrequency = H.m * H.k ^ 2 / (4 * π * (ℏ : ℝ) ^ 3)
```

**7. `SpeedOfLight`.** `wavelength (c : SpeedOfLight)` and `wavelength_inv (c : SpeedOfLight)`,
with `public import Physlib.Relativity.SpeedOfLight`; the `c = 0` caveat is gone from the
docstring, as you note it cannot happen.

**8. The three series are `abbrev`s.**

**9. `balmerFrequency_lt_lymanFrequency` for `2 ≤ d ≤ 5`.** Generalised, with the same proof
skeleton. The sharp condition is visible in the two extremes the statement compares: the Balmer
frequencies increase to the series limit `R_ν / L 1 ^ 2`, and the smallest Lyman line is
`R_ν (1 / L 0 ^ 2 - 1 / L 1 ^ 2)` at `n₃ = 1`, so the comparison holds exactly when

```
2 / L 1 ^ 2 ≤ 1 / L 0 ^ 2   i.e.   2 L 0 ^ 2 ≤ L 1 ^ 2   i.e.   2 (d - 1) ^ 2 ≤ (d + 1) ^ 2
                                                          i.e.   d ^ 2 - 6 d + 1 ≤ 0
```

that is `3 - 2 √2 ≤ d ≤ 3 + 2 √2`, hence `d ∈ {2, 3, 4, 5}` among the dimensions `2 ≤ d`
(`3 + 2 √2 ≈ 5.83`). Dimension `6` is the first failure, and it fails by an exact tie: there
`L 0 = 5/2` and `L 1 = 7/2`, the first Lyman line is `96 R_ν / 1225` and the Balmer series limit
is `4 R_ν / 49 = 100 R_ν / 1225`, so the Balmer line `n₂ = 15` (with `L 15 = 35/2`) has
frequency exactly `96 R_ν / 1225` and the lines above it are faster still. In Lean the
hypotheses are `(hd₂ : 2 ≤ H.d) (hd₅ : H.d ≤ 5)` and the arithmetic core is one `nlinarith` fed
with `0 ≤ (d - 2) (5 - d)`.

One side effect: after the generalisation the hypothesis `1 < n₂` is no longer used (the
`unusedVariables` linter flags it), and it is genuinely unnecessary, since for `n₂ ≤ 1` the
left-hand side is non-positive while the right-hand side is positive. I dropped it and recorded
the intended range (`1 < n₂` are the Balmer lines proper) in the docstring. Happy to put it back
as `_h₂` if you prefer the physical reading in the signature.

The module docstring (`## ii. Key results`) is updated accordingly.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
KevorkianPhilippe: Renamed to `energyLevel_strictMono` in `9ca0eb64`, to match `levelIndex_strictMono`.

The `Monotone` variant is there, but it keeps `2 ≤ H.d`; only the hypothesis on `k` drops (`k = 0` gives `Ry = 0`, so every level is `0`):

```lean
lemma energyLevel_monotone (hd : 2 ≤ H.d) : Monotone H.energyLevel
```

Dropping `2 ≤ H.d` would make it false. In dimension `1` the level index of the ground state vanishes, `L 0 = 0 + (1 - 1) / 2 = 0`, so `E_0 = -Ry / 0 = 0` by the `x / 0 = 0` convention, while `E_1 = -Ry < 0` as soon as `k ≠ 0`: the `n = 0` "level" sits above all the others. (Dimension `0` happens to be harmless, since `L 0 = -1/2` and `L 1 = 1/2` have the same square, but `2 ≤ d` is the clean hypothesis and the one the rest of the file uses.) The docstring records this.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
KevorkianPhilippe: Done in `9ca0eb64`, with

```lean
lemma transitionFrequency_monotone (hd : 2 ≤ H.d) (n₁ : ℕ) : Monotone (H.transitionFrequency n₁)
```

Same caveat as for `energyLevel_monotone`: the hypothesis on `k` drops, `2 ≤ H.d` cannot. For `d = 1` we get `1 / L 0 ^ 2 = 0` by the `x / 0 = 0` convention, below `1 / L 1 ^ 2 = 1`, so `n₂ ↦ -1 / L n₂ ^ 2` decreases from `n₂ = 0` to `n₂ = 1` before increasing.

Physlib/QuantumMechanics/Hydrogen/Rydberg.lean
KevorkianPhilippe: Yes, and it is now stated for all of them in `9ca0eb64`: the hypotheses are `(hd₂ : 2 ≤ H.d) (hd₅ : H.d ≤ 5)`.

The sharp condition comes from the two extremes the statement compares: the Balmer frequencies increase to the series limit `R_ν / L 1 ^ 2`, and the smallest Lyman line is `R_ν (1 / L 0 ^ 2 - 1 / L 1 ^ 2)` at `n₃ = 1`, so the comparison holds exactly when

```
2 / L 1 ^ 2 ≤ 1 / L 0 ^ 2   i.e.   2 L 0 ^ 2 ≤ L 1 ^ 2   i.e.   2 (d - 1) ^ 2 ≤ (d + 1) ^ 2
                                                          i.e.   d ^ 2 - 6 d + 1 ≤ 0
```

that is `d ≤ 3 + 2 √2 ≈ 5.83`, hence exactly `d ∈ {2, 3, 4, 5}` among the dimensions `2 ≤ d`. The proof is the one you had in mind: the arithmetic core is a single `nlinarith` fed with `0 ≤ (d - 2) (5 - d)`.

Dimension `6` is the first failure and it fails by an exact tie, which is worth recording in the docstring: there `L 0 = 5/2` and `L 1 = 7/2`, the first Lyman line is `96 R_ν / 1225` and the Balmer series limit is `4 R_ν / 49 = 100 R_ν / 1225`, so the Balmer line `n₂ = 15` (with `L 15 = 35/2`) has frequency exactly `96 R_ν / 1225`, and the lines above it are faster still.

One side effect of the generalisation: the hypothesis `1 < n₂` is no longer used (the `unusedVariables` linter flags it) and is genuinely unnecessary, since for `n₂ ≤ 1` the left-hand side is non-positive while the right-hand side is positive. I dropped it and recorded the intended range (`1 < n₂` are the Balmer lines proper) in the docstring; happy to put it back as `_h₂` if you prefer the physical reading in the signature.

KevorkianPhilippe:
Follow-up in `1bb09161`: CI's `simpNF` linter flagged `rydbergFrequency_pos`. Since `rydbergFrequency` is an abbreviation for `rydbergEnergy / h`, `simp` already closes `0 < R_ν` from `rydbergEnergy_pos`, `div_pos_iff_of_pos_left` and `h_pos`, so the mark was redundant. I dropped it and recorded the reason in the docstring; the `@[simp]` list of point 2 above is the same minus `rydbergFrequency_pos`. Everything else is unchanged.

jstoobysmith:
Sorry requested copilot review by accident please ignore that

PR #1124, opened 2026-05-26
refactor: Reorganise the QuantumInfo directory 

jstoobysmith:
No changes to content, just a reorganisation. New directory structure is the following: 

```
QuantumInfo
 ┣ Capacity
 ┃ ┣ Capacity.lean
 ┃ ┗ Capacity_doc.lean
 ┣ Channels
 ┃ ┣ Bundled.lean
 ┃ ┣ CPTP.lean
 ┃ ┣ DegradableOrder.lean
 ┃ ┣ Dual.lean
 ┃ ┣ MatrixMap.lean
 ┃ ┣ Pinching.lean
 ┃ ┗ Unbundled.lean
 ┣ ClassicalInfo
 ┃ ┣ ForMathlib
 ┃ ┃ ┗ Analysis
 ┃ ┃ ┃ ┗ SpecialFunctions
 ┃ ┃ ┃ ┃ ┗ Log
 ┃ ┃ ┃ ┃ ┃ ┗ NegMulLog.lean
 ┃ ┣ Capacity.lean
 ┃ ┣ Channel.lean
 ┃ ┣ Distribution.lean
 ┃ ┣ Entropy.lean
 ┃ ┗ Prob.lean
 ┣ Entropy
 ┃ ┣ Axiomatized
 ┃ ┃ ┣ Defs.lean
 ┃ ┃ ┗ Renyi.lean
 ┃ ┣ DPI.lean
 ┃ ┣ Relative.lean
 ┃ ┣ SSA.lean
 ┃ ┗ VonNeumann.lean
 ┣ ForMathlib
 ┃ ┣  ...
 ┣ Measurements
 ┃ ┗ POVM.lean
 ┣ Operators
 ┃ ┗ Unitary.lean
 ┣ ResourceTheory
 ┃ ┣ FreeState.lean
 ┃ ┣ HypothesisTesting.lean
 ┃ ┣ ResourceTheory.lean
 ┃ ┗ SteinsLemma.lean
 ┣ States
 ┃ ┣ Mixed
 ┃ ┃ ┣ Fidelity.lean
 ┃ ┃ ┣ MState.lean
 ┃ ┃ ┗ TraceDistance.lean
 ┃ ┣ Pure
 ┃ ┃ ┣ BargmannInvariant.lean
 ┃ ┃ ┣ Braket.lean
 ┃ ┃ ┗ Qubit.lean
 ┃ ┣ Ensemble.lean
 ┃ ┗ Entanglement.lean
 ┗ Regularized.lean
```

PR #1655, opened 2026-09-16
chore: extend check_file_imports to QuantumInfo.lean

zblore:
Each library in this repository has a root file that imports every file in the library, so that building the root builds everything and the linters see everything. A check in CI keeps that root file complete and sorted. Until now the check only knew about `Physlib.lean`. `PhyslibAlpha` has a separate check of its own, and `QuantumInfo` had none. This change gives the check a list of library roots and runs it over `Physlib` and `QuantumInfo` in turn. `Physlib` still passes, and nothing about its check has changed.

I have opened this as a draft because the new check will fail in CI on this branch, and should, until the files below are dealt with. The `QuantumInfo` root file is not sorted, and 45 files under `QuantumInfo/` are not imported from it directly. Most of those, 34, are still built, because another file imports them along the way. I will add them to the root in the same style as `Physlib.lean` once we agree the approach.

The remaining 11 files are imported by nothing, so they are never built and never linted:

`ClassicalInfo.Capacity`, `ClassicalInfo.Channel`, `Entropy.Axiomatized.Defs`, `Entropy.Axiomatized.Renyi`, `ForMathlib.HayataGroup.TraceInequality.JensenOperatorInequalityIVtoV`, `ForMathlib.Superadditive`, `ForMathlib.ULift`, `Regularized`, `ResourceTheory.ResourceTheory`, `States.Pure.BargmannInvariant`, `States.Pure.BlochSphere`.

Three of them were left out on purpose in Lean-QuantumInfo, with a note in the root saying why ("pretty broken", "Experimental"). Those notes did not survive the move in #1124. Two more are commented out in the current root without a reason, and six were never imported anywhere. Deciding which of the 11 to bring in and which to remove needs someone who knows the files. I am happy to do the follow-up either way.

I tested this with `lake env lean --run scripts/check_file_imports.lean`. The compiled version needs a C compiler to link, which CI has and my machine does not.


jstoobysmith:
I think those that do not build should just be deleted, and those that do build and can be added should just be added. We can decide then what to do with them later. Many thanks for this pull request. I think this is a great help. 

Maybe @Timeroot can say something more specific

zblore:
Ok so, QuantumInfo now import every file that builds. I've dropped the four that do not.

The QuantumInfo root imported 39 of the 84 files under it, so the checker reports 45 gaps once it covers QuantumInfo. This commit closes them.

I've sorted the root and added the 41 files that build. Removed four files that do not build against the current Mathlib pin as per your suggestion:

```
QuantumInfo/Entropy/Axiomatized/Defs.lean
QuantumInfo/Entropy/Axiomatized/Renyi.lean
QuantumInfo/Regularized.lean
QuantumInfo/ResourceTheory/ResourceTheory.lean
```

They remain in Timeroot/Lean-QuantumInfo and in this repo's history so nothing has been lost.

One of the added files, ClassicalInfo/Capacity.lean, carries four sorries in the default field values of BlockCode. sorry_lint passes anyway. The defaults compile to helpers named _default, which the linter skips as internal names, so the sorries sit outside the sorryful bookkeeping until someone relies on a default. I have left Capacity.lean as it is. Happy to open a separate PR that either removes those defaults or makes the linter report them, whichever you prefer.

QuantumInfo tree now builds, check_file_imports passes on both roots, and sorry_lint passes. Let me know if anything else is needed on this PR.

jstoobysmith:
Great! Many thanks @zblore . If there is a nice way to fix this linter error with the sorries, I think that would make a great follow-up pull request. If you want me to review this properly and get it and then merge it if it looks good, could you mark this ready for review?

jstoobysmith (approved):
Approved. This looks good to me. Many thanks for making this change. It's very helpful.

zblore:
No problem, happy to help. I'll raise the related PR for the sorries in Capacity next.

PR #1659, opened 2026-09-18
sorry_lint: count a structure's field defaults as its own sorries

zblore:
A default value for a structure field compiles to an internal definition S.field._default. The linter skips internal names so a sorry parked in a default never reached the sorryful bookkeeping.

BlockCode in QuantumInfo/ClassicalInfo/Capacity.lean carries four of them and sorry_lint passed regardless as noted on #1655. It is currently the only occurrence in Physlib and QuantumInfo.

The linter now treats a sorry in a field default as belonging to the structure. The existing two-way check then applies: the structure must carry sorryful and a sorryful tag on a structure with clean defaults fails as before. BlockCode is tagged as part of this PR and its docstring says what is unproven so the four sorries appear in the TODO list.

Without the tag the linter now fails naming BlockCode. With it sorry_lint, check_file_imports and a full build all pass.


jstoobysmith:
Looks good to me here. Though you have a line over 100 characters which needs shortening, or splitting before merging. 

awaiting-author

jstoobysmith (approved):
Approved.

PR #1660, opened 2026-09-19
feat: express variance and Gram defect of vector states as Hilbert space

naype888-cloud:
what this PR do exctaly. okey in my own words with out LLM asistens this work 's like this. first one Take an state from an unitary vector in hilbert (phi val), then  when you can fit the observable at the central point exactly then itś the same thing as substract 1/2 to the original vector this is (phi val) the greek symbol...so whith that result, the variance it's exactly the norm² from that vector, an that gram defect it's Cauchy Swartz of two vectors ‖x‖²‖y‖² − |⟪x,y⟫|². that is all the comit prooves here. tks im waithing for the rev. tks so mucho guys. ATTE LALO NAVA


naype888-cloud:
to clarify my first coment... the "fluctuation vector" is : a phy - ⟨a⟩ phy (the observable minus its expectation value, applied to ψ)

jstoobysmith (approved):

jstoobysmith:
This is good to me. I've approved and will merge shortly.

PR #1662, opened 2026-09-19
perf(WickContraction): avoid casts in the exists_join_singleton lemmas

zhikaip:
Three `exists_join_singleton_*` lemmas in `Join.lean` and `TimeCond.lean` built their witness by casting `quotContraction {a}` along `subContraction {a} _ = singleton _`. They then simplified the cast, which made `isDefEq` try to decide `[subContraction {a} _]ᵘᶜ.length =?= [singleton _]ᵘᶜ.length` by unfolding. The proofs now state the result for a general `c = subContraction {a} _`, substitute, and instantiate at `singleton _`, so no cast is needed.

| | Before | After |
|---|---|---|
| `Join.lean` (thousand heartbeats) | 1,240,610 | 138,977 (−89%) |
| `TimeCond.lean` (thousand heartbeats) | 1,756,497 | 366,153 (−79%) |

This also removes two `respectTransparency false` and three `maxHeartbeats` options. No lemmas or definitions are added or removed, and no statements change; only the three proof bodies differ.

**Disclaimer:** I instructed Claude to investigate and find opportunities to speed up build times; the changes look reasonable to me.


jstoobysmith (approved):
Looks good, approved.

PR #1568, opened 2026-08-23
feat(pauli): prove the epsilon triple-product identities

Robby955:
## Summary

Formalizes the epsilon-sensitive three-Pauli identities corresponding to equations (2.26) and
(2.27) tracked in #341, using Physlib's tensor/index notation.

The linked draft in #341 swaps the barred and unbarred Pauli factors on the right-hand sides;
the statements here use the corrected identities from the maintained reference.

This PR is stacked on #1567, which corrects the direction of non-involutive index permutations.
After #1567 merges, this branch can be rebased onto master so its diff contains only the Pauli
contribution.

## Changes

- In `Physlib/Relativity/Tensors/LeviCivita/Complex.lean`:
  - add `complexLorentzTensor.leviCivita_isReindexing`;
  - define `complexLorentzTensor.leviCivita`, with notation `ε4ℂ`, by complexifying the real
    Lorentz Levi-Civita tensor;
  - prove `complexLorentzTensor.leviCivita_eq_ofRat`, giving its exact standard-basis components
    with `ε⁰¹²³ = 1`;
  - add the private component-cast helper `basisIdxCongr_val` used by that proof.
- In `Physlib/Relativity/PauliMatrices/Relations.lean`:
  - prove `PauliMatrix.pauliContr_mul_pauliContrDown_mul_pauliContr`:
    `σ^μ barσ^ν σ^ρ = g^{μν}σ^ρ - g^{μρ}σ^ν + g^{νρ}σ^μ + i ε^{μνρκ}σ_κ`;
  - prove `PauliMatrix.pauliContrDown_mul_pauliContr_mul_pauliContrDown`, with conjugate Pauli
    matrices and the opposite epsilon sign;
  - add private component tables `sigmaComponent`, `barSigmaComponent`, `metricComponent`,
    `sigmaTensorComponent`, `barContrDownTensorComponent`, `barCoTensorComponent`,
    `sigmaCoDownTensorComponent`, and `metricTensorComponent`;
  - add private product encodings `sigmaBarProduct`, `sigmaBarProductComponent`,
    `barSigmaSigmaProduct`, `barSigmaSigmaProductComponent`, `metricSigmaProductComponent`, and
    `metricBarSigmaProductComponent`;
  - add private component bridge lemmas `toTensor_eq_ofRat_component`,
    `pauliContrDown_eq_ofRat_component`, `pauliCo_eq_ofRat_component`,
    `pauliCoDown_eq_ofRat_component`, `contrMetric_eq_ofRat_component`,
[...]

Issue #341, opened 2025-02-14
Properties of the Pauli matrices


jstoobysmith:
This is an issue to track the formalization of properties of the Pauli matrices. 

The goal is to prove the properties of the Pauli matrices using index notation with HepLean. The properties of the Pauli matrices to be proved are (2.23)-(2.30) of [this note](https://particle.physics.ucdavis.edu/modernsusy/slides/slideimages/spinorfeynrules.pdf). 

- [ ] 2.23
- [ ] 2.24 
- [ ] 2.25
- [ ] 2.26
- [ ] 2.27
- [ ] 2.28
- [ ] 2.29
- [ ] 2.30

It may be best to complete a number of steps before this:

- [x]  Improve the API around the basis of tensors and tensor trees, to allow calculations to be done by looking at each component (Done in #343)
- [x]  Improve the @[simp]'s lemma around tensor trees and tensors, to make the calculations of the above as simple as possible. (Done in #343)
- [x] Define an integer version of (coordinates of) tensors, and show that it injects into all tensors. On these `decide` can be used. (Done in #343)
- [x] Using the above three, simplify the proof and speed of: `pauliCo_contr_pauliContr`. (Done in #343)
- [ ] Define the epsilon Lorentz tensor with four four-vector indices.
- [ ] Write the above lemmas out using index notation in Lean.

After these have all been formalized: 
- [ ] Make a curated noted containing for Pauli matrices and their properties. 

This issue is related to: #286. 

Contributions to this issue are very much welcome in any direction.

PR #1663, opened 2026-09-19
docs(PauliMatrices): cite the published source for the three-Pauli identities

zhikaip:
The docstrings of the two three-Pauli identities in `Relations.lean` referred to "Equation (2.26)" and "Equation (2.27)" without naming a source. Those numbers come from the 2004 draft linked in #341, where the two equations differ from the statements proved here (as noted in #1568).

This removes the bare equation numbers from the docstrings and adds a References section to the module docstring citing equations (2.54) and (2.55) of the published paper, arXiv:0812.1594, which agree with the Lean statements and conventions. It also puts each formula in those docstrings on a single line, so the inline code is not broken across lines, and removes a stale "not currently cited" note from the bibliography entry.

No lemmas or definitions are added, removed or changed.

**Disclaimer:** I instructed Claude to find the proper reference; I have checked the cited equations myself.


jstoobysmith (approved):
Approved

PR #1661, opened 2026-09-19
Bump to version: v4.34.0

NicolasRouquette:
Bumps [mathlib][f-lakefile-mathlib] and [doc-gen4][f-lakefile-docgen] to `v4.34.0` and [the toolchain][f-toolchain] to `leanprover/lean4:v4.34.0`.

Closes #1656.

Both libraries build clean and every linter CI gates on passes:

| | result |
|---|---|
| `lake build` | 9590 jobs, 0 errors, 9 warnings |
| `lake build PhyslibAlpha` | 9278 jobs, 0 errors, 11 warnings |
| [`check_file_imports`][x-cfi], [`check_dup_tags`][x-cdt], [`sorry_lint`][x-sl], [`runPhyslibLinters`][x-rpl] | pass |
| [`runPhyslibAlphaLinters`][x-rpal], [`noAlphaImports`][x-nai], [`alphaFileImports`][x-afi] | pass |
| [`scripts/lint-style.sh`][x-lintstyle], [`scripts/PhyslibAlpha/alphaPythonLinters.sh`][x-alphapy] | pass |

Deprecation warnings in `Physlib`/`QuantumInfo` are down from 147 to 9.
All 20 remaining warnings are deliberate and listed at the end.

Four commits, in the order they are worth reading:

1. [`chore: Bump to Lean v4.34.0`][cm-1] — the bump and the mechanical port.
2. [`fix(PauliMatrices): prove leviCivita_mul_pauliDual via rational components`][cm-2]
   — the only proof whose strategy changed; see below.
3. [`chore(PhyslibAlpha): port to Lean v4.34.0`][cm-3] — `PhyslibAlpha` is not in
   [`defaultTargets`][f-lakefile-targets], so `lake build` never reaches it. It
   has its own CI workflow and needed its own port.
4. [`chore(PhyslibAlpha): clear the linter warnings v4.34.0 surfaced`][cm-4].

This bump touches 176 files, so every section below links its declarations, call
sites and warnings straight to source — the diff is wide but shallow, and the links
are there to spare reviewers the hunt. They resolve against this PR's head commit
([`1bc0543`][cm-4]) in this repository rather than the branch on my fork, so they
survive the merge and the branch's deletion and still work for anyone reading this
PR later. They pin the code as submitted; I will regenerate them if I force-push.

## Checklist (#1656)

**Basics**

- [x] [mathlib rev → `v4.34.0`][f-lakefile-mathlib] in `lakefile.toml`
[...]

PR #1664, opened 2026-09-19
perf(PauliMatrices): evaluate Levi-Civita components as a product of signs

zhikaip:
In a heartbeat profile of all 585 files on master, the three most expensive files are `WickContraction/TimeCond.lean` (1,756k), `WickContraction/Join.lean` (1,241k) and `PauliMatrices/Relations.lean` (1,185k). The next file is at 705k, and a typical file is at 3k–40k. They also sit at the ends of the two longest import chains, so they bound the parallel build time. #1662 deals with the two Wick files; this PR deals with `Relations.lean`.

The three-Pauli identities are proved by `decide +kernel` over all components. `leviCivita_eq_ofRat` gives the components of `ε4ℂ` as a `generalizedKroneckerDelta`, i.e. a 4×4 determinant, which the kernel evaluates through permutation sums. This PR adds the closed form `∏_{i<j} sign (g j - g i)` for the Levi-Civita symbol on `Fin n` and uses it for the components instead. Evaluating all 256 components of `ε4ℂ` costs the kernel 40.3M heartbeats in the determinant form and 4.3M in the product form. The PR also restricts two `rw`s that traversed the whole goal to the summand they apply to.

Added:
- `leviCivitaSymbolProd` (def) and `leviCivitaSymbol_eq_leviCivitaSymbolProd`, in `Mathematics/LeviCivita/Basic.lean`
- `leviCivita_eq_ofRat_prod`, in `Tensors/LeviCivita/Complex.lean`

Nothing is removed and no statement changes. Three proofs change: `pauliContr_mul_pauliContrDown_mul_pauliContr`, `pauliContrDown_mul_pauliContr_mul_pauliContrDown` and `leviCivita_mul_pauliDual`.

| `Relations.lean`, thousand heartbeats | Before | After |
|---|---|---|
| whole file | 1,185,316 | 743,601 (−37%) |
| `pauliContr_mul_pauliContrDown_mul_pauliContr` | 408,138 | 171,955 |
| `pauliContrDown_mul_pauliContr_mul_pauliContrDown` | 517,382 | 236,754 |
| `leviCivita_mul_pauliDual` | ~29,000 | 100,824 |

**Relation to #1661.** The current proof of `leviCivita_mul_pauliDual` hits a kernel timeout on v4.34, and #1661 (commit 191842c) replaces it with a proof by `decide +kernel` over components, using `leviCivita_eq_ofRat`. That proof evaluates `ε4ℂ` 2048 times (256 components × the contracted index × both sides), and every evaluation is a determinant: by the figures above, about 320M heartbeats of kernel work for `ε4ℂ` alone, which would make this small lemma as expensive as one of the three-Pauli identities before this PR. This PR takes the same proof and states it with `leviCivita_eq_ofRat_prod`, which brings that part down to about 34M; the whole lemma then measures 100,824k. That is the last row of the table. It is a regression against the present proof, accepted because the present proof does not survive v4.34. I have checked that this branch builds on v4.34 when merged onto #1661. The two PRs will conflict in this one proof; the resolution is to keep the `leviCivita_eq_ofRat_prod` version, after which 191842c is no longer needed.

**Disclaimer:** I instructed Claude to investigate and find opportunities to speed up build times; the changes look reasonable to me.


zhikaip:
Highlighting the following: this PR will conflict will the version bump in #1661, and the current changes are based on the proposed change there. Either order of merging works, but I think merging this first helps with performance (I forked the #1661 branch and it takes painfully long to build the file)

Physlib/Mathematics/LeviCivita/Basic.lean
```diff
@@ -150,3 +155,38 @@ lemma leviCivitaSymbol_eq_zero_iff {g : ι → ι} :
...
+/-!
+
+## E. Closed form on `Fin n`
+
+-/
+
+/-- The product over pairs `i < j` of the sign of `g j - g i`, for `g : Fin n → Fin n`.
+This equals `leviCivitaSymbol g` by `leviCivitaSymbol_eq_leviCivitaSymbolProd` and, unlike
+the determinant defining `leviCivitaSymbol`, is cheap to evaluate, e.g. by `decide`. -/
+def leviCivitaSymbolProd {n : ℕ} (g : Fin n → Fin n) : ℤ :=
```
jstoobysmith: Do we need this definition, or can we just use the lemma below to rewrite it?

zhikaip: removed the def and inlined it in the lemma

zhikaip:
After the version bump my 16GB RAM PC simply cannot build Relativity/PauliMatrices/Relations.lean on master (stuck at the decide steps), so I think this change is a massive performance improvement.

A related question: is there any reason why you introduced `RatComplexNum`? It seems like Mathlib's `GaussianInt` suffices for this purpose (all the component tables are 0, ±1, ±i), and that would be a further speed-up if we completely replace the `RatComplexNum` API.

jstoobysmith:
I think if Mathlib GaussianInt works, then we should use that instead of RatComplexNum. Is this up to date with the bump? If so, I'll approve it, and I'll leave you to merge it.

jstoobysmith (approved):

